PR #59, opened 2019-07-22
New template title

timarney:
Closes https://github.com/cds-snc/notification-api/issues/113

@willeybryan was it just the one change for this "make it larger and bold"

Screen Shot 2019-07-22 at 10 23 53 AM

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,17 @@ Set a variable
  </div>
 ```
 
+For forms
+
+```
+from flask_babel import _
+```
+
+Wrap your text
+```
+_('Your text here')
+```
+
 - Extract
 
 ```bash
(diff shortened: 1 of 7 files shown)


willeybryan:
Yeah pretty much. The user's eye should already be there as they clicked on the button to reveal that information.

Would it be possible to increase the spacing between New Template and the templates above a little bit? Something Like 12-24 Pixels or 1-2em. Just to add a line of spacing?


I recognize this may also apply to the new template button the user clicks before that. but I think it looks fine:
image

timarney:
Will bump down ...

In the version I'm looking at the UI splits into 2 sections.  The part that says New Template becomes "sticky" i.e. stays in place vs the top half scrolls and actually fades over top of the sticky part.

You might be able to see that here https://notification.cdssandbox.xyz/services/cb9c86a3-d62f-4e87-bcb7-89f6e660365d/users 

<hr>

Screen Shot 2019-07-22 at 1 03 18 PM

willeybryan:
Yeah I am able to recreate that if I adjust my browser window size. That is good to know. I will ensure that @JuliannaR is aware of that for the accessibility audit.

PR #61, opened 2019-07-22
Change "see" to "check"

brdunfield:
Resolves https://github.com/cds-snc/notification-api/issues/111


app/templates/views/dashboard/dashboard.html
```diff
@@ -30,14 +30,14 @@ <h2 class="heading-medium">
     {{ ajax_block(partials, updates_url, 'totals', interval=5) }}
     {{ show_more(
       url_for('.monthly', service_id=current_service.id),
-      'See messages sent per month'
+      'Check messages sent per month'
```
timarney: These should be wrapped for translation see the Readme section for that.

Once wrapped you can run the extract and compile commands to generate a .po file

PR #66, opened 2019-07-23
Update team members content

timarney:
Closes https://github.com/cds-snc/notification-api/issues/115


brdunfield:
I see that the "Current team members" translation was added, but it doesn't look like it's actually on the page?

timarney:
Updated :)

PR #68, opened 2019-07-23
Template - content

timarney:
Closes https://github.com/cds-snc/notification-api/issues/114


brdunfield:
I think some of the text needs to come out:

From the issue:
Under "Optional content" replace text before the example with the following :
"You can enter optional content in your template by uploading a spreadsheet with information about your users. "
"For example, your spreadsheet may include the age of each user. You want users who are under 18 to receive optional content. Use double brackets and '??' to include the optional content."

Looks like the original text was left in instead of replaced.

PR #69, opened 2019-07-23
Updating to support novascotia.ca trialing it

willeybryan:


timarney:
Testing outside the Federal level still?

willeybryan:
Yes, NovaScotia is looking to build one too and I offered to let the use a trial service in exchange for their business case and research. They will get no support and I am setting expectations that they can't just use it instead.

PR #71, opened 2019-07-23
Update terms of service content (#110)

brdunfield:
Closes https://github.com/cds-snc/notification-api/issues/101


brdunfield:
Links still work, but the agreement page seems to be missing (was broken with the long form url as well).

timarney:
@YedidaZalik I don't see a link to the agreement here anymore https://www.notifications.service.gov.uk/features/terms

There's still an agreement page but it uses a service_id.

```
@main.route('/services/<uuid:service_id>/agreement')
```

YedidaZalik:
I guess they have removed it since I worked on this. @willeybryan should I rewrite this section for our purposes or just delete for now?
"Your organization must also accept our data sharing and financial agreement. Download the agreement or contact us to ask if we already have an agreement with your organization."

willeybryan:
Delete it, we don't have anything with respect to that yet.

brdunfield:
That link still works. It's more general vs dealing with the specific agreement that no longer exists. Ok to leave in or should I still delete?

YedidaZalik:
Which link do you mean?  
Or if this question is not for me, no worries!

brdunfield:
The contact us link in the Leaving Notification section:

"You can leave Notification at any time. Just _contact us_ and we’ll close your account."

PR #126, opened 2019-08-07
change financial to fiscal

brdunfield:
Closes https://github.com/cds-snc/notification-api/issues/145


app/templates/views/dashboard/monthly.html
```diff
@@ -6,8 +6,8 @@
 {% extends "withnav_template.html" %}
 
 {% block service_page_title %}
-  {{ _('Messages sent') }},
-  {{ selected_year }} {{ _('to') }} {{ selected_year + 1 }} {{ _('financial year') }}
+  Messages sent,
```
timarney: Looks like this needs to be wrapped.

brdunfield: Wow I merged that totally incorrectly. Thanks for the catch

PR #131, opened 2019-08-07
Update .po + mo

timarney:


timarney:
https://github.com/cds-snc/notification-api/issues/156

PR #145, opened 2019-08-12
JS I18N

timarney:
This covers the initial setup to handle i18n for JS files.

Notes:
- Added a simple WebPack config file  - this allows us to start using other scripts from NPM that use require / imports
- Strings get extracted the the app .po file
- Get current lang now sets the lang for the current locale + html file which should also fix https://github.com/cds-snc/notification-api/issues/205

Update radioSelect.js for testing.

Screen Shot 2019-08-12 at 11 41 31 AM


app/__init__.py
```diff
@@ -232,7 +235,8 @@ def inject_global_template_variables():
         return {
             'asset_path': application.config['ASSET_PATH'],
             'header_colour': application.config['HEADER_COLOUR'],
-            'asset_url': asset_fingerprinter.get_url
+            'asset_url': asset_fingerprinter.get_url,
+            'current_lang': get_current_locale(application)
```
timarney: @brdunfield using current_lang here now vs have duplicate code in the signed out template.

PR #169, opened 2019-08-19
Mask password with a digest

maxneuvians:
Closes https://github.com/cds-snc/notification-api/issues/240

The idea here is that clear text passwords should not show up in API logs. The simple solution was to mask them as a message digest with a secret salt before they get passed to the API. This does not stop the clear text password from showing up in the Admin logs or somebody using the salt to build a rainbow table for the logs if they had access to both.

Using a more complex hashing solution like bcrypt does not work because each time the password is hashed before it goes to the API it receives a new salt which will then be a different string. The API does not know if this is a hashed string or a clear string password.

### DIFF
--- a/app/notify_client/user_api_client.py
+++ b/app/notify_client/user_api_client.py
@@ -1,3 +1,5 @@
+import os
+import hashlib
 from notifications_python_client.errors import HTTPError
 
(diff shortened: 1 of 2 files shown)


brdunfield:
Probably need to merge in the a11y fixes?

PR #175, opened 2019-08-20
More a11y fixes

maxneuvians:
Closes https://github.com/cds-snc/notification-api/issues/271

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -109,10 +109,8 @@ navigation = {
 
 def get_current_locale(application):
     requestLang = request.accept_languages.best_match("sd")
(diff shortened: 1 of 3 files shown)


app/__init__.py
```diff
@@ -108,9 +108,11 @@ def _get_current_service():
 
 
 def get_current_locale(application):
-    requestLang = request.accept_languages.best_match(application.config['LANGUAGES'])
+    requestLang = request.accept_languages.best_match("sd")
+    print(requestLang)
```
brdunfield: Do we still need the print statements? And should the config var be restored in the line above?

PR #209, opened 2019-10-08
WIP - Homepage

timarney:


timarney:
Okay so, outside of the actual sending of the Contact form this is done.

cc: @willeybryan , @maxneuvians 

Now the 
**Blocker**

The current API call does require a user id (in all the calls I see)

```
def send_support_email(self, user_id, message):
        data = {'email': self.contact_email, 'message': message}
        endpoint = '/user/{0}/support-email'.format(user_id)
        self.post(endpoint, data=data)
```

That said this page should allow as you can hit it from a logged out state:
https://notification.alpha.canada.ca/support/ask-question-give-feedback

So we need to dig into that.

willeybryan:
When I hit https://notification.alpha.canada.ca/support/ask-question-give-feedback while logged out and submit a form, I get an error

image

timarney:
Thinking we might have removed something

As it also exists here

https://www.notifications.service.gov.uk/support/ask-question-give-feedback

timarney:
@caycefischer any further comments on the design end of things?

caycefischer:
oh one more thing I just noticed in the English site - "notify" here should be capitalized:

image

timarney:
@caycefischer that text is fixed now.

PR #255, opened 2019-11-07
Pricing page updates - translations

timarney:


timarney:
@smcmurtry not adding the compiled .mo here as the English text needs to be updated.

PR #259, opened 2019-11-07
DIY translations for strings coming from other repo

smcmurtry:
This will have no effect until we add square brackets to the strings we want translated from here: https://github.com/cds-snc/notification-utils/blob/master/notifications_utils/jinja_templates/email_preview_template.jinja2


timarney:
After the utils repo is updated need to update the requirements file for this as well.

PR #264, opened 2019-11-08
Bumped notification-utils version, compiled translations

smcmurtry:
The translations in the email preview template are now working for me locally 🎉


requirements.txt
```diff
@@ -31,54 +31,5 @@ awscli-cwlogs>=1.4,<1.5
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous==0.24  # pyup: <1.0.0
 
-git+https://github.com/cds-snc/notifier-utils.git@36.0.2#egg=notifications-utils==36.0.2
+git+https://github.com/cds-snc/notifier-utils.git@36.0.3#egg=notifications-utils==36.0.3
 
-## The following requirements were added by pip freeze:
```
smcmurtry: Not sure why this and following lines got removed

timarney: Check the Make file

```
SHELL := /bin/sh #bash | sh
```

Swap to bash and switch back without committing that file.

PR #267, opened 2019-11-08
Branding messages

timarney:
Closes:
https://github.com/cds-snc/notification-api/issues/417


timarney:
Merged here - will need translations

PR #303, opened 2019-12-03
Meta / Open Graph

timarney:
Screen Shot 2019-12-03 at 2 26 42 PM


timarney:
Updated to use the image from the blog post (which works well for en + fr).

@willeybryan , @caycefischer happy with this?  We can iterate as we go but would be nice to have something setup sooner than later.

Screen Shot 2019-12-04 at 8 43 56 AM

timarney:
Closes 
https://github.com/cds-snc/notification-api/issues/480
https://github.com/cds-snc/notification-api/issues/481

Will get translations after next round from Elise.

willeybryan:
Yeah looks good to me

PR #304, opened 2019-12-04
Update live notes

timarney:


timarney:
Closes https://github.com/cds-snc/notification-api/issues/483

PR #308, opened 2019-12-06
Redirect usage page

timarney:


tests/app/main/test_permissions.py
```diff
@@ -341,11 +340,11 @@ def test_services_pages_that_org_users_are_allowed_to_see(
     assert mock_get_service.called is organisation_checked
 
 
+@pytest.mark.skip(reason="?")
```
timarney: @maxneuvians can you take a look at this one?

PR #321, opened 2020-01-06
Email scheduler

timarney:


timarney:
> Just add the tests to Grunt and we are GTG. Awesome work!

Going to look into the best way to handle this....

Need to work on the overall flow.  Current setup is copying in the files from the other repo.

The webpack setup in the repo doesn't handle hot reloading etc... mostly because we're in a "mixed" spot for compiling the JS between Webpack and Gulp.

https://github.com/cds-snc/notification-scheduler#notification-scheduler

PR #345, opened 2020-01-24
POC - Alert message threshold

timarney:
@maxneuvians 

This proof of concept is an effort to get the ball rolling on this.

This new route could be hit via a lambda / cron function and send an alert accordingly.

1)  This is slow
2) This could tax the DB
3) Could be attacked (to slow down the service)

==========

A better approach might be to run a check after a message gets updated pulled off the queue - i.e. a running total for the day... something more realtime as messages get sent.  Maybe save / reset that value to avoid constant expensive lookups.

### DIFF
--- a/app/main/views/platform_admin.py
+++ b/app/main/views/platform_admin.py
@@ -3,7 +3,7 @@ import re
 from collections import OrderedDict
 from datetime import datetime
 
-from flask import abort, flash, redirect, render_template, request, url_for
+from flask import jsonify, abort, flash, redirect, render_template, request, url_for
 from notifications_python_client.errors import HTTPError
 from requests import RequestException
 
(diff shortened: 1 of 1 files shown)


timarney:
Thoughts on what to call out to to fire the alert to AWS?  Will also want to only fire once.  

Also we'll need to figure how we want to handle if we receive that alert... 

Are we treating this as more of a flag something has happened that we're expecting the end user to handle or are we being proactive.  In which case what are the steps...

maxneuvians:
We should be notifying the user once they approach their limit by sending a message. We can either look into putting a flag into the service table or using redis as a counter (although that can fail). We can create a log watcher in AWS to alert us on those errors.

timarney:
Moving over here https://github.com/cds-snc/notification-api/pull/587

PR #373, opened 2020-02-29
Content - Add how to do embedded link in email template + typo fix

duboisp:


timarney:
@ioanacontu @caycefischer 

Do we have documentation around the research we did on "Always use full URLs"?

> Always use full URLs, starting with https://

ioanacontu:
> @ioanacontu @caycefischer
> 
> Do we have documentation around the research we did on "Always use full URLs"?
> 
> > Always use full URLs, starting with https://

I don't think I was around when that happened. Cayce or Antoine might know about it (but they are both gone at the time - we'll ask Antoine next week).

timarney:
@AntoineGarciaSuarez can you provide some guidance around this based on your research?

This Pull Request from @duboisp (thanks Pierre)

Would update the docs to show how to make links in "apply now" vs showing the actual link http://example.com/apply 

The current docs state 

> Always use full URLs, starting with https://

But we haven't outlined why we make that statement

Do we have documentation around the research we did on "Always use full URLs"?

AntoineGarciaSuarez:
When we did our reasearch activities with the public, the participants paid attention to the HTTPS. It was clear that this way of writing links increased their confidence.
So I definitly recommand to leave these format in the email template.

However, for the senders, we don't have any relevant insights for that. 
The branding was the element that increased their trust the most.

let me know if you need anything else

timarney:
Okay thanks @AntoineGarciaSuarez 

@anikbrazeau @ioanacontu 

Based on the above I think it's fine to show both ways of adding links.

In which case we shouldn't say "Always use full URLs, starting with https://"

At this point this PR is quite out of sync with master so it will need to be rebased or started fresh.

ioanacontu:
I just remembered a conversation we had a while back about this. We were talking about how it helps preventing against phishing if we show full URLs. That because often with phishing emails, there is an "apply now" call to action that has behind it a link to a false website. Does anyone else remember that or am I mixing things up? @timarney @AntoineGarciaSuarez @anikbrazeau

timarney:
Yes we had that conversation - 

As it stands I can make it look like a "full link" to a valid url - but the link "behind" could be directed else where.

Screen Shot 2020-03-18 at 1 10 00 PM

<hr>

Screen Shot 2020-03-18 at 1 10 11 PM

ioanacontu:
So, in other words. It doesn't make any difference :) So then, we should just move ahead with the above and **show both ways of adding links**.

duboisp:
I understand the best practice and those should be documented. I suggest to add a section in the formatting guide.

But in regard of the content of the email, if it is badly coded is it your responsibility or the responsibility of the user that do the configuration of the email template? 
Like in AEM, we do some recommendation (as best practice) and we prioritize request that is conforming to those best practice. But if an institution decide to publish a web content that is not WCAG compliant, then that non-compliance fall on their shoulder, not on the shoulder of the service provide. And if they ask for help and we see there is some accessibility issue, our first recommendation is get those accessibility issue fixed and then re-evaluate.

timarney:
@ioanacontu 

Closing this 

We should update the formatting guide.

> I understand the best practice and those should be documented. I suggest to add a section in the formatting guide.

PR #377, opened 2020-03-03
Update copy to remove unlimited texts

anikbrazeau:


timarney:
Can you update the FR string in here as well?

PR #378, opened 2020-03-03
Remove unlimited text messages - for accuracy

anikbrazeau:
Changed copy and images in EN and FR to refer only to unlimited emails (not texts)


timarney:
Looks like the pricing image needs to be flipped showing wrong image in "EN" / "FR"

timarney:
As per above

Screen Shot 2020-03-04 at 9 34 38 AM

PR #385, opened 2020-03-05
A few content fixes

anikbrazeau:
- Missing translation wrapping on line
- Add version to integrate image
- Fix broken link in French
- .po and .mo files


timarney:
Did you want to get the "Notify is built for you" image fix for mobile in on this as well.

PR #396, opened 2020-03-11
Hiding letter statistics from the platform admin home page with css

### LINKED ISSUE #393: Remove "Letters Sent" from platform admin summary page
Sending letters with Notify is not possible currently. So the stats that say how many letters have sent are currently just visual clutter:
Screen Shot 2020-03-10 at 1 09 45 PM

smcmurtry:
Closes #393


timarney:
@smcmurtry if either in this PR or another if we can also move the email + sms heading up to line up with the #'s that would be fantastic.

Screen Shot 2020-03-12 at 7 39 43 AM

Approved as is.

timarney:
Also related: https://github.com/cds-snc/notification-admin/issues/399

PR #409, opened 2020-03-16
Update config.py

sboisvert:
Send emails from the contact us form directly into Freshdesk


sboisvert:
@damcom333 when we confirm the email is working let's merge this

PR #410, opened 2020-03-17
Quick fix on Technical documentation page + more

anikbrazeau:
- Clarifying the technical documentation page clients and endpoint setup bilingually
- Making sure consistent in FR term Modèle --> Gabarit 
- Lingering Notification --> Notify


timarney:
@anikbrazeau thoughts on tweaking this a bit further.  Wondering if people will understand it's for the Notify clients vs direct integration.

```
Before integrating the API using the Notify clients below 
you need to change your endpoint to https://api.notification.alpha.canada.ca.

[How to change your endpoint](link)
```

PR #411, opened 2020-03-18
Displaying send stats for each api key to admin users

smcmurtry:
This won't work until my other PR is merged: https://github.com/cds-snc/notification-api/pull/698

698 is now merged, review app is working, here is a page where you can see the change (if your account is a platform admin): https://notification-api-key-st-tjgp75.herokuapp.com/services/10a12f7f-43dd-4828-b690-af5ac4f0b598/api/keys

### DIFF
--- a/app/templates/views/api/keys.html
+++ b/app/templates/views/api/keys.html
@@ -55,7 +55,7 @@
               Expires: <span class="local-datetime-full">{{ item.expiry_date }}</span>
             </div>
             <div class="hint">
-              Last used: <span class="local-datetime-full">{{ item.last_send }}</span>
+              Last used: <span >{{ item.last_send }}</span>
             </div>
           {% endif %}
         </div>
--- a/app/templates/views/api/keys.html
+++ b/app/templates/views/api/keys.html
@@ -46,16 +46,16 @@
           </div>
           {% if current_user.platform_admin %}
             <div class="hint">
-              {{ item.total_sends }} total sends ({{ item.email_sends }} email, {{ item.sms_sends }} sms)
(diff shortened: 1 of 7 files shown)


timarney:
Nice.

A couple of things.

Can we break total sends into  email + sms?

Can we also include: Key created, Expires, Last activity

==================================================

This page puts us into a good spot to add daily activity at some point.

tests/conftest.py
```diff
@@ -1823,6 +1823,15 @@ def _get_keys(service_id):
     return mocker.patch('app.api_key_api_client.get_api_keys', side_effect=_get_keys)
 
 
+@pytest.fixture(scope='function')
```
timarney: Does the api return 0 or False there haven't been any sends?  

Will need to handle accordingly.

smcmurtry: 0

app/templates/views/api/keys.html
```diff
@@ -44,6 +44,20 @@
               {{ _('Test – pretends to send messages') }}
             {% endif %}
           </div>
+          {% if current_user.platform_admin %}
+            <div class="hint">
+              {{ item.total_sends }} total sends ({{ item.email_sends }} email, {{ item.sms_sends }} sms)
```
timarney: Should wrap this text while we're in here.

app/templates/views/api/keys.html
```diff
@@ -44,6 +44,20 @@
...
+            </div>
+            <div class="hint">
+              Created: <span class="local-datetime-full">{{ item.created_at }}</span>
```
timarney: Let's wrap this.

timarney:
Nice work.

One more thing.

Let's add some commas to the numbers to make them easier to read

11010 = 11,010

```
f'{totals:,}'
```

PR #417, opened 2020-03-20
Calendar component updates

brdunfield:
- Content and design updates as per: https://www.figma.com/file/IXg01iWgJiXYKAGJHjAPPu/Calendar-updates?node-id=0%3A1
- logic to not show times that shouldn't be shown
- 24 hr / AM PM support
- general bugfixes and getting the code synced up with notification-scheduler
- french translations
- limit send to 96 hours in advance


app/assets/javascripts/scheduler/Confirmation/Confirmation.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const Confirmation = () => {
+  const { selected, time } = useContext(store);
+  const { translate } = useContext(I18nContext);
+
+  // reconstruct date for display:
+  const date = selected + "T" + time;
+
+  return (selected && time) ? (
+    <div className="confirmation set">
```
timarney: Not looking at the css do you have a class called set?
.confirmation 
.set 

or .confirmation-set ?

timarney: I see you have unset below

app/assets/javascripts/scheduler/store.js
```diff
@@ -89,28 +98,52 @@ export const StateProvider = ({ value, children }) => {
...
+          newState.errors = "";
+
+          if (
+            JSON.stringify(newState.selected) === JSON.stringify(state.selected)
+          ) {
+            // show deselect error
+            newState.errors = [
+              {
+                id: "1",
+                text: "Date must be selected",
```
timarney: I think we handle this translation elsewhere correct?

app/assets/javascripts/scheduler/Confirmation/Confirmation.js
brdunfield: .confirmation
.set

it's .unset for the red box (which is broken in this port but I'll be putting up a PR in notification-scheduler to fix)

app/assets/javascripts/scheduler/store.js
brdunfield: This is probably replaced / can maybe use the same translation as the new red box. I'll check and update it.

timarney:
Screen Shot 2020-03-24 at 5 16 47 AM

<hr>
Screen Shot 2020-03-24 at 5 19 11 AM

<hr>
Screen Shot 2020-03-24 at 5 20 45 AM

<hr>

Screen Shot 2020-03-24 at 5 29 37 AM

timarney:
Lets keep going in this branch to get the Send Now , Send Later buttons in which will gets us to a merge to master state (will be fully functional at that point).

timarney:
Almost there :) really great work on this.

**First load issue:**
Screen Shot 2020-03-30 at 12 40 05 PM
<hr>

**Disabled button**
Need to prevent clicks (event.stopPropagation() )
Screen Shot 2020-03-30 at 12 45 18 PM

<hr>
Might as well fix this one now as well.
https://github.com/cds-snc/notification-api/issues/716

**Scheduling too far in advance **
Screen Shot 2020-03-30 at 12 44 48 PM

PR #419, opened 2020-03-25
Set preview back button more dynamically

sastels:
fixes https://github.com/cds-snc/notification-api/issues/682

The issue was that the way `step_index` is determined the the screens depends on whether or not the user clicked "Use my email address". The back button on the last page was being set based on one way of setting it (if the user typed in the email address manually).

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -787,20 +787,28 @@ def is_current_user_the_recipient():
 
 def get_back_link(service_id, template, step_index):
     if get_help_argument():
-        if step_index == 0:
+        if request.endpoint == 'main.check_notification':
             return url_for(
-                'main.start_tour',
+                'main.send_test',
                 service_id=service_id,
                 template_id=template.id,
+                help=get_help_argument()
             )
-        elif step_index > 0:
-            return url_for(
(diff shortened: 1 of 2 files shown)


timarney:
@sastels - I'll aim to look at this tomorrow.

brdunfield:
From what I can see, I still need to click the back button twice before the page changes. The first click tries to navigate to "step-1" which resolves to the same page. The second click navigates to "step-0", which loads the email entry page.

sastels:
> From what I can see, I still need to click the back button twice before the page changes. The first click tries to navigate to "step-1" which resolves to the same page. The second click navigates to "step-0", which loads the email entry page.

ah crud, I tested it with 3 user variables and it looked like it was working there - no idea why that would make a difference. I shall investigate...

brdunfield:
If it helps, my test only used email and first name.

brdunfield:
Made some changes to the get_back_link logic, and removed some cases that were causing the issue.

brdunfield:
Restored the functionality I took out before (i.e. if it's on the check page it goes back to the start instead of the most recent page)

PR #429, opened 2020-04-08
Added tests for query param change

smcmurtry:
Follow up to #426. This test checks the following:
- english homepage is displayed for `/?lang=en` 
- french homepage is displayed for `/?lang=fr` 
- english homepage is displayed for `/?lang=sa?SDFa?DFa,/` 
- english homepage is displayed for `/?xyz=xyz`
- english homepage is displayed for `/?sa?SDFa?DFa,/=sa?SDFa?DFa,/`

 And that the server responds with 200 for all of these weird urls.

I am arbitrarily just testing this for the homepage and no other pages, but the behaviour should be the same for other pages as well.

### DIFF
--- a/tests/app/main/views/test_index.py
+++ b/tests/app/main/views/test_index.py
@@ -234,3 +234,31 @@ def test_letter_template_preview_headers(
     )
 
     assert response.headers.get('X-Frame-Options') == 'SAMEORIGIN'
+
+
+@pytest.mark.parametrize('query_key, query_value, heading', [
+    ('lang', 'en', 'Notify'),  # 'Notify' = english heading
+    ('lang', 'fr', 'Notification'),  # 'Notification' = french heading
+    ('lang', 'sa?SDFa?DFa,/', 'Notify'),
+    ('xyz', 'xyz', 'Notify'),
+    ('sa?SDFa?DFa,/', 'sa?SDFa?DFa,/', 'Notify')
+])
+def test_query_params(
+    mocker,
+    client,
+    mock_get_service_and_organisation_counts,
+    query_key,
+    query_value,
+    heading
+):
+    mocker.patch('app.service_api_client.get_live_services_data', return_value={'data': service})
+
+    response = client.get(url_for('main.index', **{query_key: query_value}))
+
+    assert response.status_code == 200
+
+    page = BeautifulSoup(response.data.decode('utf-8'), 'html.parser')
+
+    assert page.h1.text.strip() == (
+        heading
+    )


tests/app/main/views/test_index.py
```diff
@@ -234,3 +234,31 @@ def test_letter_template_preview_headers(
...
+])
+def test_query_params(
+    mocker,
+    client,
+    mock_get_service_and_organisation_counts,
+    query_key,
+    query_value,
+    heading
+):
+    mocker.patch('app.service_api_client.get_live_services_data', return_value={'data': service})
```
obrien-j, line 254: What's the purpose of this piece of code?

tests/app/main/views/test_index.py
```diff
@@ -234,3 +234,31 @@ def test_letter_template_preview_headers(
...
+
+    response = client.get(url_for('main.index', **{query_key: query_value}))
+
+    assert response.status_code == 200
```
obrien-j, line 258: Given we're testing invalid inputs, should we be returning a valid response or a 4xx?

tests/app/main/views/test_index.py
```diff
@@ -234,3 +234,31 @@ def test_letter_template_preview_headers(
     )
 
     assert response.headers.get('X-Frame-Options') == 'SAMEORIGIN'
+
+
+@pytest.mark.parametrize('query_key, query_value, heading', [
+    ('lang', 'en', 'Notify'),  # 'Notify' = english heading
+    ('lang', 'fr', 'Notification'),  # 'Notification' = french heading
+    ('lang', 'sa?SDFa?DFa,/', 'Notify'),
```
obrien-j, line 242: I'm curious whether it's better to do validation in our code-base, or to rely on babel itself, and simply handle the errors it throws?

Both work, so I'm open to alternative views on the matter.

tests/app/main/views/test_index.py
smcmurtry: It is mocking the api client that returns stats that gets displayed on the homepage - # of services, # of notifications. I copied this from the existing tests for the homepage.

tests/app/main/views/test_index.py
smcmurtry: Good question but I think it is out of scope for this PR + #426. Notify's existing behaviour was to return 200 and display the page, when you add a made up query param to the URL. So I didn't change that behaviour.

tests/app/main/views/test_index.py
smcmurtry: Prior to #426, we were taking the value from the `lang` query param and [storing it as the user's language in their session](https://github.com/cds-snc/notification-admin/pull/426/files#diff-1f9f743421417fb2794c88447083a031R113-R124). So my change was to check that the language is `'en'` or `'fr'` before putting it in the session. This value gets read directly from the session, not just by Babel, but [to determine which GC logo to display](https://github.com/cds-snc/notification-admin/blob/master/app/templates/main_template.html#L149). So I think it is preferable to not write a bad value to the session, vs write a bad value and then catch the babel error.

tests/app/main/views/test_index.py
obrien-j: +1 john.dumbquestions. :)

tests/app/main/views/test_index.py
obrien-j: Fair enough, let's investigate in another issue.

PR #448, opened 2020-04-16
Security update

smcmurtry:
I ran `npm audit fix`


timarney:
Open to discussion but I think we should let snyk manage vs snyk + npm.  I know npm is saying X vulnerable packages which snyk isn't picking up.

cc: @damcom333

PR #451, opened 2020-04-16
Security update

smcmurtry:
Snyk says the following for a few of our vulnerabilities:
> Remediation: Your dependencies are out of date, otherwise you would be using a newer version of package x. Try relocking your lockfile or deleting node_modules, reinstalling and running snyk wizard. If the problem persists, one of your dependencies may be bundling outdated modules.

So I took this advice and deleted the package-lock file, node_modules, and ran npm install. This resulted in some old dependencies being updated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,11 +5,11 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.5.5",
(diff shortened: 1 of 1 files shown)


smcmurtry:
I'll bump this tomorrow to see if that fixes the tests.

PR #454, opened 2020-04-18
Fix FR template label

smcmurtry:
This is for both the service dashboard page, and the details page you see after following the "Check templates used by month" link.

Before:
Screen Shot 2020-04-17 at 9 18 20 PM

After:
Screen Shot 2020-04-17 at 9 17 35 PM

I wrote a new macro to switch the order of the words for french, and wrapped the string properly.

No change for english text.


anikbrazeau:
LGTM, thanks!

PR #462, opened 2020-04-22
Tie StatsD enabled to HOST

maxneuvians:
This brings it in line with how the API works https://github.com/cds-snc/notification-api/blob/master/app/config.py#L328-L330


app/config.py
```diff
@@ -138,7 +136,6 @@ class Live(Config):
     SHOW_STYLEGUIDE = False
     HEADER_COLOUR = '#005EA5'  # $govuk-blue
     HTTP_PROTOCOL = 'https'
-    STATSD_ENABLED = False
     MOU_BUCKET_NAME = 'notifications.service.gov.uk-mou'
```
timarney, line 139: Thinking this should be updated per above or at least to remove the uk-mou

PR #476, opened 2020-04-27
Initial flask_caching implementation and wrapping of get_latest_stats call

obrien-j:
<grinding rusty noise>

Just popped in a few changes to optimize this main page query as it's a potential path for self-denial of service. I also reorded the calling of the stats endpoint to not call it unnecessarily, as well as removing the live-stats endpoint as it was basically a unmitigated hammer against the DB.

<applies oil, only slightly rusty now.>

### DIFF
--- a/app/main/views/index.py
+++ b/app/main/views/index.py
@@ -35,7 +35,6 @@ QUESTION_TICKET_TYPE = 'ask-question-give-feedback'
 def index():
 
     lang = get_current_locale(current_app)
-    stats = get_latest_stats(lang)
 
     if current_user and current_user.is_authenticated:
         return redirect(url_for('main.choose_account'))
@@ -53,6 +52,8 @@ def index():
         return redirect(url_for(
             '.thanks',
         ))
+
+    stats = get_latest_stats(lang)
 
     if request.method == 'POST':
         return render_template(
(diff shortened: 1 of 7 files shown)


app/utils.py
```diff
@@ -54,16 +55,15 @@
 user_is_logged_in = login_required
 
 
+@cache.cached(timeout=300, key_prefix='latest_stats')
```
maxneuvians, line 58: The API query pulls from an endpoint that updates once a day at 4 AM EST, you could put the timeout to something much higher like an hour. The thing you want to avoid is if the sever reboots at 23:59 and then has the old stats for X time period into the next day.

obrien-j: Yeah, I wasn't sure exactly when https://github.com/cds-snc/notification-api/blob/9a39fcc994d057bf154c07e130ca1c6b3a5789c8/app/dao/services_dao.py#L88 refreshes it's daily stats, so I figured I'd cheat. I'll create an issue to track future improvements to this, as the whole thing's a bit shaky end to end. 

This whole call path should have start/end date configurable time periods via the api.

app/main/views/index.py
```diff
@@ -54,6 +53,8 @@ def index():
             '.thanks',
         ))
 
+    stats = get_latest_stats(lang)
```
smcmurtry, line 56: good call, no need to call when the client is posting form data

mcman12:
waiting for review.

obrien-j:
> This all looks good but how do I know it's working? Is it possible to add a test for this?

Potentially, but not sure how to handle the time.travel(). Glad to add this in at a later date though.

PR #477, opened 2020-04-28
Add ability to write new longcodes into the db

smcmurtry:
Working on https://github.com/cds-snc/notification-api/issues/732.

~~Will keep this a draft until it has tests.~~ It now has tests. This is for https://github.com/cds-snc/notification-api/pull/772 but could be merged independently and nothing would break.

I'm not totally sure why this feature did not already exist - I guess the GDS devs preferred to add new longcodes to the db with a csv file and CLI?


timarney:
API - changes are set to be reviewed on staging

PR #484, opened 2020-04-30
Fix fr display of html elements classed relative-time-past

smcmurtry:
Closes https://github.com/cds-snc/notification-api/issues/804

Screen Shot 2020-04-30 at 2 48 08 PM

Sidenote: I don't like moment. It probably wouldn't be too hard to replace it with the built in formatting methods and get rid of a fair bit of page weight.


smcmurtry:
Okay I fixed all the times that are formatted with client side js in that file. That fixed the example you found @brdunfield:

Screen Shot 2020-05-01 at 3 35 29 PM

PR #490, opened 2020-05-01
Make branding variables easier to read / understand

brdunfield:
This PR is in tandem with other PR's in notification-utils and notification-api.

This addresses the first 2 bullet points of cds-snc/notification-api#761

Cleans up variable names, and makes it more clear how the radio buttons in the admin interface correspond to what the code does.

The tests will fail here until notification-utils is merged.


app/main/forms.py
```diff
@@ -1371,10 +1371,10 @@ def __init__(self, *args, **kwargs):
 
 
 branding_options = (
-    ('govuk', 'GOV.UK only'),
+    ('fip_english', 'GOV.UK only'),
```
timarney, line 1374: Can we update this to remove GOV.UK?

PR #497, opened 2020-05-04
Add FR translations for some (most?) of the "x days ago" type strings

smcmurtry:
Working on https://github.com/cds-snc/notification-api/issues/774

I didn't think this was going to be very easy/possible, but it turns out we were already using a jquery plugin called [timeago](https://timeago.yarp.com/) that supports french! It was being used for formatting some of the "x time ago" type strings. So I just replaced the python ago library with that one. Will keep this a draft until I fix the tests. 

Here are the new screens:

User settings:
Screen Shot 2020-05-04 at 3 40 09 PM

Recent sends:
Screen Shot 2020-05-04 at 3 39 30 PM

Template preview page for a template that has been edited:
Screen Shot 2020-05-04 at 3 42 36 PM

@anikbrazeau can you take a look at these and let me know if you see any problems?

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -1,7 +1,7 @@
 import itertools
 import os
 import urllib
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 from functools import partial
 from numbers import Number
 from time import monotonic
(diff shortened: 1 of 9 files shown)


anikbrazeau:
For the Template preview page for a template that has been edited, can we make a small copy change to the label before (since it's not a date per se):

Date de la dernière modification -----> **Dernière modification**

PR #499, opened 2020-05-05
make flash messages bilingual

sastels:
fixes https://github.com/cds-snc/notification-api/issues/754

make the `flash_messages.html` template bilingual.

Example
Old:
image

New:
image


app/templates/flash_messages.html
```diff
@@ -3,11 +3,12 @@
   {% if messages %}
     {% for category, message in messages %}
       {% if category in ['cancel', 'delete', 'suspend', 'resume', 'remove', 'revoke this API key', 'block', 'archive'] %}
-        {% set delete_button_text = "{}, {}".format(_('Yes'), _(category)) %}
```
sastels, line 6: AFAIK we can't look up translations dynamically like this.

app/translations/en/LC_MESSAGES/messages.po
```diff
@@ -259,183 +259,195 @@ msgid "A key with this name already exists"
 msgstr ""
 
 #: app/main/forms.py:825
+msgid "Inbound SMS number"
+msgstr ""
```
sastels, line 263: This just got magically added by the extract

app/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -823,6 +835,46 @@ msgstr "Sécurité"
 msgid "research mode"
 msgstr "mode de recherche"
 
+#: app/templates/flash_messages.html:6
```
sastels, line 838: This section added intentionally

app/translations/en/LC_MESSAGES/messages.po
smcmurtry: Yeah that's from a PR I merged yesterday. We never update the translations file unless we are specifically adding translated text. Mostly to avoid merge conflicts if someone else is working on translations.

app/templates/flash_messages.html
smcmurtry: I believe you're right

app/templates/flash_messages.html
```diff
@@ -3,11 +3,12 @@
...
+        {% set category_translations = {'cancel': _('cancel'), 'delete': _('delete'), 'suspend': _('suspend'), 'resume': _('resume'), 'remove': _('remove'), 'revoke this API key': _('revoke this API key'), 'block': _('block'), 'archive': _('archive')} %}
+        {% set delete_button_text = "{}, {}".format(_('Yes'), category_translations[category]) %}
```
smcmurtry, line 7: Could you add a check here that category is a key in category_translations? If it's not the server will return an error code.

app/templates/flash_messages.html
smcmurtry: I believe you're right

app/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -823,6 +835,46 @@ msgstr "Sécurité"
...
+msgid "archive"
+msgstr "archives"
+
+#: app/templates/flash_messages.html:9
+msgid "Try again"
+msgstr "Réessayer"
+
+#: app/templates/flash_messages.html:11
+msgid "Unblock"
+msgstr "Débloquer"
```
smcmurtry, line 876: Where did you get these translations from?

sastels: other places in the translation file + Anik for "revoke this API key"

app/templates/flash_messages.html
sastels: ah yeah, forgot that we always want to use `get` to access a dict. changed to a `get` with a default of category coming in
```
{% set delete_button_text = "{}, {}".format(_('Yes'), category_translations.get(category, category)) %}
```

smcmurtry: I had no idea dictionaries had a `get` method that lets you return a default. Very cool!

app/translations/fr/LC_MESSAGES/messages.po
anikbrazeau: Is archive a verb in this context? (do we know where that shows up?

smcmurtry: Yes I think it's safe to say "archive" is a verb since it is an action the button does.

sastels: verb, as in
Text: "There's no way to reverse this! Are you sure you want to archive this user?"
Button: "Yes, archive"

anikbrazeau: Can we switch it to **"archiver"** in French then :D

PR #504, opened 2020-05-06
Apply translations to all email preview templates

smcmurtry:
This fixes part of https://github.com/cds-snc/notification-api/issues/802

The code to do this was already written, the problem was it only got called for the main preview page and nowhere else. So I moved that code to the `get_template` function in utils.py so that it will be called whenever a preview template is used. 

This will also work for the SMS preview template, but I will have to do another PR to notification-utils to wrap to To and From fields there in square brackets, since that is the hack we are using to find strings that need to be translated.

Test send:
Screen Shot 2020-05-06 at 3 04 42 PM

CSV send:
Screen Shot 2020-05-06 at 3 05 03 PM

Tutorial:
Screen Shot 2020-05-06 at 3 10 42 PM


app/translations/fr/LC_MESSAGES/messages.po
```diff
@@ -3272,7 +3276,8 @@ msgid ""
 "Administrators who have to meet departmental service\n"
 "              standards for communication"
 msgstr ""
-"Pour les administrateurs devant respecter les normes de service ministérielles de leurs communications"
+"Pour les administrateurs devant respecter les normes de service "
```
brdunfield, line 3279: Is this related to this PR?

app/__init__.py
```diff
@@ -365,6 +367,26 @@ def format_delta(_date):
     return timeago.format(date, now, lang)
 
 
+def translate_preview_template(_template_str):
+    translate = {
+        "From": _("From"),
+        "To": _("To"),
+        "Subject": _("Subject"),
+        "Reply to": _("Reply to")
```
brdunfield, line 375: It doesn't seem like this is actually called anywhere in the code. Is this for a new feature down the line? Does it need to be in this PR?

brdunfield: the "Reply to" specifically - didn't realize it would highlight the whole block

app/translations/fr/LC_MESSAGES/messages.po
smcmurtry: No it just happened when I re-generated the file 🤷 
I will revert it.

app/__init__.py
smcmurtry: Yes - it does not appear in this codebase because it is a translation for the utils repo 😄 See [this line](https://github.com/cds-snc/notification-utils/blob/0cd4a33acf26ccfd7590f14e943e46637676a060/notifications_utils/jinja_templates/email_preview_template.jinja2#L15).

PR #508, opened 2020-05-08
Translated the send button

smcmurtry:
Closes https://github.com/cds-snc/notification-api/issues/843

Screen Shot 2020-05-08 at 2 27 47 PM


app/translations/en/LC_MESSAGES/messages.po
```diff
@@ -4907,23 +4916,23 @@ msgstr ""
 msgid "Last edited"
 msgstr ""
 
-#: app/templates/views/templates/template.html:59
+#: app/templates/views/templates/template.html:57
```
brdunfield, line 4919: Any idea why all these line numbers changed? It doesn't look like any changes were made to template.html in this PR

smcmurtry: It's because I didn't run `pybabel extract` and `pybabel update` on an older PR that touched that file. Even if it's not a translation-related change, just deleting a line in a file with translations in it will cause this. 
This is one of the reasons I would like to move to just looking up the translations in a normal dict/json file.

brdunfield: That makes sense. I guess in that sense we need to be careful when approving PRs until if / when we come up with a better solution.

PR #515, opened 2020-05-11
Translations for add reply-to email flow

smcmurtry:
Closes https://github.com/cds-snc/notification-api/issues/799

Screen Shot 2020-05-11 at 11 06 14 AM

Screen Shot 2020-05-11 at 10 56 12 AM

This error message was also untranslated before:
Screen Shot 2020-05-11 at 11 04 24 AM


app/main/validators.py
```diff
@@ -79,7 +80,7 @@ def __call__(self, form, field):
 class ValidEmail(Email):
 
     def __init__(self):
-        super().__init__(_('Enter a valid email address'))
+        super().__init__(_l('Enter a valid email address'))
```
sastels, line 83: just curious, why the switch?

smcmurtry: Before making this change I was actually seeing the un-translated error text on the screen in my last screenshot above. The form field is part of a "partial" template that is passed to the main template that gets rendered. So I think the partial template is rendered outside of the request and doesn't know the user's language. But lazy_gettext solves that. That's my best guess at why this works.

anikbrazeau:
@smcmurtry 

Tiny change: remove double period

"Cela peut prendre une minute."

smcmurtry:
> @smcmurtry
> 
> Tiny change: remove double period
> 
> "Cela peut prendre une minute."

That page is actually a loading screen with an animation: it changes between having 1, 2, and 3 periods to show the loading. I just happened to take a screenshot when there were 2 on the screen. So not a typo.  😄

PR #517, opened 2020-05-11
Add assistance link to menu header

sboisvert:
We're often answering the same questions in the support channel which takes a lot of time away from being able to build the application. We want to let users have the best experience possible and to do so we should allow them to find answers to their questions and get the support they need.

This is a first pass at adding a link to Freshdesk's Knowledge Base.

Another thing that we should probably do in a future PR is change the Technical Documentation link to point directly to https://notification.assistance.cds-snc.ca/en/support/solutions/folders/61000161438 since this answers another of the most frequently asked questions which is regarding the base API url.

French content:
https://notification.assistance.cds-snc.ca/en/support/home => 
https://notification.assistance.cds-snc.ca/fr/support/home

FAQs & Assistance = > Assistance et FAQ

### DIFF
--- a/app/templates/views/signedout.html
+++ b/app/templates/views/signedout.html
@@ -122,7 +122,7 @@ Canada Notification service') %} {% set meta_title = _('GC Notify | GC notificat
             class="text-small text-blue mr-10 font-bold no-underline hover:underline"
             >{{ _("Technical documentation") }}</a>
           <a
-            href="{{ _('url.freshdesk.home') }}"
+            href="{{ _('https://notification.assistance.cds-snc.ca/en/support/home') }}"
             class="text-small text-blue mr-10 font-bold no-underline hover:underline"
             >{{ _("FAQs & Assistance") }}</a>
           <a
--- a/app/templates/views/signedout.html
+++ b/app/templates/views/signedout.html
@@ -122,6 +122,10 @@ Canada Notification service') %} {% set meta_title = _('GC Notify | GC notificat
             class="text-small text-blue mr-10 font-bold no-underline hover:underline"
             >{{ _("Technical documentation") }}</a>
           <a
+            href="{{ _('url.freshdesk.home') }}"
+            class="text-small text-blue mr-10 font-bold no-underline hover:underline"
+            >{{ _("FAQs & Assistance") }}</a>
+          <a
             href="{{ url_for('.feedback', ticket_type='ask-question-give-feedback') }}"
             class="text-small text-blue mr-10 font-bold no-underline hover:underline"
             >{{ _("Contact us") }}
--- a/app/templates/views/signedout.html
+++ b/app/templates/views/signedout.html
@@ -153,6 +153,12 @@ Canada Notification service') %} {% set meta_title = _('GC Notify | GC notificat
           <li>
             <a
               class="block text-small mb-5 no-underline hover:underline"
+              href="{{ _('https://notification.assistance.cds-snc.ca/en/support/home') }}"
+              >{{ _("FAQs & Assistance") }}</a>
+          </li>
+          <li>
+            <a
+              class="block text-small mb-5 no-underline hover:underline"
               href="{{ url_for('.feedback', ticket_type='ask-question-give-feedback') }}"
               >{{ _("Contact us") }}</a>
           </li>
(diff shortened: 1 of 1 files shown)


anikbrazeau:
Can we hold on this, we're figuring out the navigation with authenticated / unauthenticated / different states and where things fit top nav, footer, etc.

sboisvert:
@anikbrazeau Can you help me understand why we wouldn't want to iterate on this now? It seems that this small change would make a really big quality of life improvement for people :)

anikbrazeau:
@sboisvert Definitely! You're ahead of us on this one!  

We are iterating on this and want to improve it asap. BUT, there are considerations to take into account here when changing the nav. We're working on making it as relevant and useful as possible so that it matches the different needs of users with different permissions authenticated in Notify. 

This is happening in parallel to a dashboard revamp (to help people achieve their main goals) based on what we learned from usability testing as well. We will obviously also be connecting Notify to the Support platform to help people get help.

Sharing some of our early thinking on this here's with a _draft_ mock-up of just some of the pages/places that might be included:

Screen Shot 2020-05-14 at 11 33 19 AM

It would be great to review something like this in future to see how it fits in with our other ongoing work and to check for consistency of terms we use, sentence case, and accessibility, before building in code.

sboisvert:
@anikbrazeau that looks like great work and I think it'll be an awesome end outcome. I'm not sure I was clear in my question, I'm just wondering what sub-optimal outcomes would come out from having a slightly better menu now and making it awesome later vs keeping the menu with it's fault now and then making it awesome later. Maybe I'm missing something :)

ioanacontu:
@sboisvert The top menu navigation at the moment has a Contact us item. Assistance, FAQ and Contact Us are part of the same category of functionality, therefore eventually we will merge them into a "Support" section. I would also challenge the assumption that using multiple main navigation items to send people to the help section it the best user experience. We can talk more about the urgency of this with Bryan too. If it is a priority, we can focus on creating that 'Support' page and accommodate your team's need for a 'door' from Notify towards the knowledge base in Freshdesk.

anikbrazeau:
+1 

Let's discuss how much of a priority this is for Notify/users. And just make sure we're dotting our i's and crossing our t's before going ahead with this (let's check the authenticated experience too (where people are more likely looking for support), and decide what's the best word to use where, and if this is the best place for this item, or if contact us duplicates the same function etc.)

sboisvert:
I just want to make sure I'm clear :), I think that if we add a link now, folks will be better better served immediately. That won't stop any of the other awesome (and likely far better) work that you all will do with the menu in the future when you take a holistic view of it.

I'm just very concerned that we're letting perfect be the enemy of better. We're having many support requests (including one today) that are answered by our support documentation that is impossible for our users to access.

Again, I agree that a full holistic approach into this later is best, but given that it's been about a month since we first started talking about this and have spent many hours discussing it, I think it's imperative on us to focus on the the end user and their needs. Serving them a bit better now doesn't stop us from having the best fixes in the future :).

anikbrazeau:
Perhaps there's another way we could make a simple quick fix happen so that users can access the knowledge base in the interim? 

Here's a quick mock up:
https://www.figma.com/file/fRVRKm7cM8MUuBWUJbrfgU/Senders-content-creators-service-operators?node-id=1370%3A3326

sboisvert:
I sadly don't have access to the file, but I'm pretty sure I'll be game for it! Whatever quick fix we can do to help folks have access to it now works for me :)

mcman12:
@sboisvert we can't link to documentation in freshdesk straight from the web as freshdesk and the documentation pages have not been assessed for accessibility. So that option would not be on the table for a quick fix.

sboisvert:
@mcman12 Can you send me the link to the Notify accessibility evaluation just so I can see what it looks like? Or just send me a link to what `assessed for accessibility` mean? I'm not sure I know what it is :(

JuliannaR:
> @JuliannaR For some reason I thought you had gone over Freshdesk A11y with @caycefischer, am I just imagining things?

No, this hasn't been done, but I can make space for it next week. Can you send me what you need reviewed so I can look at how long it will take?

sboisvert:
https://notification.assistance.cds-snc.ca/en/support/home is what we'd link to :)

PR #538, opened 2020-05-19
Fix translations workflow

smcmurtry:
What I did:
- installed a new python dependency called translate-toolkit, which contains [csv2po and po2csv](http://docs.translatehouse.org/projects/translate-toolkit/en/latest/commands/csv2po.html)
- ran `make freeze-requirements`, which updated some packages
- generated some csv files (new_translations/en.csv, fr.csv) from our existing po files using po2csv
- deleted the file location column in the csv files - since our translation strings are unique, we don't need it
- changed the existing `make babel` recipe to generate the old .po files from the new .csv files
- gitignored the old .po and .mo files, but kept the folders in place so that we don't run into any errors when we generate these on the fly
- added the `make babel` step to our dockerfile. @maxneuvians do I need to add this to any other dockerfiles, etc to make sure the translations are generated in CI, and when the app is deployed to staging and prod?

Note: I tried using po2json/json2po to get this working w/ a json file, but to no avail.

Takeaways:
- I think this will solve our merge conflict problems when multiple people are working on translations
- we won't be able to run a command to extract newly wrapped strings from the app into the csv file, we will have to add rows manually for now. We could implement that, it would just take some more work
- all rows of the csv have to be unique


smcmurtry:
Snyk is failing me because of license issues - @damcom333 are these dealbreakers?

damcom333:
Hey. Try again on that pull. We changed this issue to flag a warning. But not fail the build

brdunfield:
Is there a reason the csv files can't be in the existing translation folder? (add a csv/ subdir if you want them separated). Not a fan of a new directory and it being named 'new_translations'. 😞

smcmurtry:
@brdunfield No problem - I moved the csv files to the existing directory.

app/translations/csv/en.csv
```diff
@@ -0,0 +1,1122 @@
+"source","target"
```
brdunfield, line 1: I feel like eventually we can simplify this further and only have a single csv (especially since the keys are all just the english string anyways). But that's more of a comment / thought for the future vs anything to do with this PR!

PR #539, opened 2020-05-19
Organisation default branding fix

brdunfield:
Part of the fix for: https://github.com/cds-snc/notification-api/issues/875

This PR, in tandem with the API PR, allows organisation branding to be set to one of the default brandings without throwing an error.

I fixed up the tests so that this shouldn't slip through again in the future.


tests/app/main/views/test_service_settings.py
```diff
@@ -3262,10 +3263,17 @@ def test_should_set_branding_and_organisations(
             )
         assert mock_update_organisation.called is False
     elif endpoint == 'main.organisation_preview_email_branding':
-        mock_update_organisation.assert_called_once_with(
-            ORGANISATION_ID,
-            email_branding_id=submitted_value
-        )
+        if submitted_value == '__FIP-EN__' or submitted_value == "__FIP-FR__":
```
brdunfield, line 3266: @sastels this test (using the parameters above) tests changing organisation branding.

PR #546, opened 2020-05-21
Translated the jobs page

smcmurtry:
Closes https://github.com/cds-snc/notification-api/issues/785
Closes https://github.com/cds-snc/notification-api/issues/883

Screen Shot 2020-05-21 at 9 17 44 AM

~~I may add a translation for "test of"~~ Nevermind that was the subject of my email.

Screen Shot 2020-05-21 at 3 40 53 PM

### DIFF
--- a/app/assets/javascripts/main.js
+++ b/app/assets/javascripts/main.js
@@ -15,6 +15,21 @@ window.formatAllDates = function () {
     }
   });
 
+  $(".local-datetime-short-year").each(function(index) {
+    let datetime = new Date($(this).text().trim());
+    let locale = window.APP_LANG == "fr" ? "fr-CA" : "en-US";
+  
+    if (datetime instanceof Date && !isNaN(datetime)) {
+      $(this).text(
+        datetime.toLocaleString(locale, {
+          month: "short",
+          day: "numeric",
+          year: "numeric"
+        })
+      );
+    }
+  });
+
   $(".local-datetime-full").each(function(index) {
     let datetime = new Date($(this).text().trim());
     let locale = window.APP_LANG == "fr" ? "fr-CA" : "en-US";
--- a/app/templates/partials/jobs/notifications.html
+++ b/app/templates/partials/jobs/notifications.html
@@ -41,7 +41,7 @@
           <p class="{% if template.template_type != 'letter' %}bottom-gutter{% endif %}">
             <a href="{{ download_link }}" download class="heading-small">{{ _('Download this report') }}</a>
             &emsp;
-            <span id="time-left">{{ time_left }}</span>
+            {{ _("Data available until") }} <span id="time-left" class="local-datetime-short-year">{{ available_until_date }}</span>
           </p>
         {% endif %}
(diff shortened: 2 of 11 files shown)


app/assets/javascripts/main.js
```diff
@@ -15,6 +15,21 @@ window.formatAllDates = function () {
     }
   });
 
+  $(".local-datetime-short-year").each(function(index) {
```
smcmurtry, line 18: I added a new datetime format

app/templates/partials/jobs/notifications.html
```diff
@@ -41,7 +41,7 @@
           <p class="{% if template.template_type != 'letter' %}bottom-gutter{% endif %}">
             <a href="{{ download_link }}" download class="heading-small">{{ _('Download this report') }}</a>
             &emsp;
-            <span id="time-left">{{ time_left }}</span>
+            {{ _("Data available until") }} <span id="time-left" class="local-datetime-short-year">{{ available_until_date }}</span>
```
smcmurtry, line 44: This is a content change to make the translation easier. See: https://github.com/cds-snc/notification-api/issues/883

tests/app/main/views/test_jobs.py
```diff
@@ -675,16 +675,18 @@ def test_should_show_updates_for_one_job_as_json(
...
-    "job_created_at, expected_message", [
-        ("2016-01-10 11:09:00.000000+00:00", "Data available for 7 days"),
-        ("2016-01-04 11:09:00.000000+00:00", "Data available for 1 day"),
-        ("2016-01-03 11:09:00.000000+00:00", "Data available for 12 hours"),
-        ("2016-01-02 23:59:59.000000+00:00", "Data no longer available")
+    "job_created_at, expected_date", [
+        ("2016-01-10 11:09:00.000000+00:00", "2016-01-18"),
+        ("2016-01-04 11:09:00.000000+00:00", "2016-01-12"),
+        ("2016-01-03 11:09:00.000000+00:00", "2016-01-11"),
+        ("2016-01-02 23:59:59.000000+00:00", "2016-01-10")
```
smcmurtry, line 682: These tests all changed because now we are just putting a raw datetime string in the span, which later gets formatted by js on the client. That formatting doesn't exist in the python tests so I am just returning the date part of the datetime string and checking it is correct.

brdunfield: It might be worth putting some version of this explanation in the comments near one of these blocks.

brdunfield:
While the page is loading, the unformatted date string is shown (for 1-2 seconds). I'm guessing there isn't any way to make that better? Not sure if it's important enough to put work into or not, might be worth an ask around the team?

smcmurtry:
@brdunfield Yes I also find this annoying. The same problem is present everywhere we are displaying dates. For example on the "last logged in page" that appears at login has a noticable flash of the unformatted date. 

We could solve this either by getting the js to load faster or by doing the date formatting in python. I filed an issue here: https://github.com/cds-snc/notification-api/issues/886

PR #549, opened 2020-05-22
Add platform admin setting for service message limit

sastels:
closes https://github.com/cds-snc/notification-api/issues/764

Add a platform admin service setting to change the daily message limit.

I think it's also clearing the appropriate caches? `service_api_client.update_service()` seems to take care of that.


app/navigation.py
```diff
@@ -294,6 +294,7 @@ class HeaderNavigation(Navigation):
         'service_verify_reply_to_address',
         'service_verify_reply_to_address_updates',
         'services_or_dashboard',
+        'set_message_limit',
         'set_free_sms_allowance',
```
sastels, line 298: I added `set_message_limit` in this file everywhere `set_free_sms_allowance` was... 🤷

app/main/views/service_settings.py
```diff
@@ -916,6 +917,22 @@ def service_set_letter_contact_block(service_id):
     )
 
 
+@main.route("/services/<service_id>/service-settings/set_message_limit", methods=['GET', 'POST'])
```
brdunfield, line 920: Nitpicky, but by convention, routes in notify use `-` instead of `_`. Would you mind changing this to `set-message-limit`?

app/main/views/service_settings.py
```diff
@@ -916,6 +917,22 @@ def service_set_letter_contact_block(service_id):
...
+@user_is_platform_admin
+def set_message_limit(service_id):
```
brdunfield, line 922: This one should stay as it is! `set_message_limit` 👍

PR #554, opened 2020-05-25
remove offending line

sastels:
fixes https://github.com/cds-snc/notification-api/issues/791

Remove a confusing line from the "Team members" page


app/templates/views/manage-users.html
```diff
@@ -14,7 +14,6 @@
   <h1 class="heading-medium">
     {{ _('Team members') }}
   </h1>
-  <p>{{ _('Share templates with your team') }}</p>
```
brdunfield, line 17: Should we also remove it from the translation csvs?

PR #562, opened 2020-05-27
add translations to "Confirm changes to variables" page

sastels:
fixes https://github.com/cds-snc/notification-api/issues/855

image

Add a bunch of translations. Couldn't get the fancy plural `ngettext()` to work in the template for some reason. The function was recognized but it wouldn't put the French in. Possibly some sort of run-time thing?


sastels:
👍  Gonna merge for now, we can revisit if someone feels moved to investigate babel plural + csv files...

PR #567, opened 2020-05-28
Translate template folder display

brdunfield:
Closes https://github.com/cds-snc/notification-api/issues/857

I noticed that in french folders that aren't empty were also not being translated (i.e. they would show as "1 template", "x templates" in french as well). So I added translations for that too.


app/models/template_list.py
```diff
@@ -164,7 +165,7 @@ def _hint_parts(self):
         if self.number_of_templates == 1:
             yield '1 template'
         elif self.number_of_templates > 1:
-            yield '{} templates'.format(self.number_of_templates)
+            yield '{} '.format(self.number_of_templates) + _('templates')
```
sastels, line 168: We can add variables into babel with something like
```
 _('%(number)s templates', number=self.number_of_templates)
```
where the csv lines are
```
"%(number)s templates", ""
```
and
```
"%(number)s templates", "%(number)s gabarits"
```
Not that it's necessary here, but it's cleaner if we find we're breaking up lines into parts and translating the parts seperately.

app/models/template_list.py
```diff
@@ -164,7 +165,7 @@ def _hint_parts(self):
...
 
         if self.number_of_folders == 1:
             yield '1 folder'
```
sastels, line 171: We could do this and the following plural one while we're here 🤷‍♂️

brdunfield:
@sastels So what I've gathered is that these strings are wrapped and translated further downstream, which is why neither 'Empty' nor '1 template' requires the wrapping to work. The plural one I needed to wrap "templates" in order for it to work, since downstream there's no key for "x templates".

However, your suggestion might work for the plural one, which would keep things consistent. I'll give it a try.

sastels:
> @sastels So what I've gathered is that these strings are wrapped and translated further downstream, which is why neither 'Empty' nor '1 template' requires the wrapping to work. The plural one I needed to wrap "templates" in order for it to work, since downstream there's no key for "x templates".

ahhh, I understand now. 👍

PR #570, opened 2020-05-29
Delete Select template content dropdown and related code

sastels:
fixes https://github.com/cds-snc/notification-api/issues/841
fixes https://github.com/cds-snc/notification-api/issues/852

delete the "Select template content" dropdown and related code / translation

Now:

image

Note that I have deleted both horizontal rules. I noticed that the "New text message template" has neither the dropdown nor the rules.

image

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -564,21 +564,12 @@ def add_service_template(service_id, template_type, template_folder_id=None):
             template_id='0'
         ))
     else:
-        template_select = []
-
-        if(template_type == "email"):
-            templates = template_api_prefill_client.get_template_list()
-            if(templates):
-                for template in templates:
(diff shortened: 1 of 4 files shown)


app/main/views/templates.py
```diff
@@ -564,21 +564,12 @@ def add_service_template(service_id, template_type, template_folder_id=None):
...
-                for template in templates:
-                    template_select.append({'id': template["id"], 'name': template["name"]})
-
         return render_template(
             'views/edit-{}-template.html'.format(template_type),
             form=form,
             template_type=template_type,
             template_folder_id=template_folder_id,
             service_id=service_id,
-            template_select=template_select,
```
sastels, line 581: This is only used in the "Select template content" dropdown

PR #571, opened 2020-05-29
Fix copy template button inside a folder

sastels:
fixes https://github.com/cds-snc/notification-api/issues/909

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -338,12 +338,16 @@ def _add_template_by_type(template_type, template_folder_id):
 @main.route("/services/<service_id>/templates/copy/from-folder/<uuid:from_folder>")
 @main.route("/services/<service_id>/templates/copy/from-service/<uuid:from_service>")
 @main.route("/services/<service_id>/templates/copy/from-service/<uuid:from_service>/from-folder/<uuid:from_folder>")
+@main.route("/services/<service_id>/templates/all/folders/<uuid:from_folder>/copy")
 @user_has_permissions('manage_templates')
 def choose_template_to_copy(
     service_id,
     from_service=None,
     from_folder=None,
 ):
+
+    if from_folder and from_service is None:
+        from_service = service_id
 
     if from_service:
 


app/main/views/templates.py
```diff
@@ -335,16 +335,23 @@ def _add_template_by_type(template_type, template_folder_id):
 
 
 @main.route("/services/<service_id>/templates/copy")
+@main.route("/services/<service_id>/templates/all/copy")
+@main.route("/services/<service_id>/templates/email/copy")
+@main.route("/services/<service_id>/templates/sms/copy")
```
sastels, line 340: These are routes that users can get to by clicking "copy" but that didn't exist in the app.

app/main/views/templates.py
```diff
@@ -335,16 +335,23 @@ def _add_template_by_type(template_type, template_folder_id):
...
 @main.route("/services/<service_id>/templates/copy/from-folder/<uuid:from_folder>")
 @main.route("/services/<service_id>/templates/copy/from-service/<uuid:from_service>")
 @main.route("/services/<service_id>/templates/copy/from-service/<uuid:from_service>/from-folder/<uuid:from_folder>")
+@main.route("/services/<service_id>/templates/all/folders/<uuid:from_folder>/copy")
```
sastels, line 344: This is the route that was causing the original problem: clicking on "copy" when in a folder.

brdunfield:
Interesting behaviour in this feature.. (not related to this PR). Looks like if you click the copy button from within a folder, navigate to either root or a higher up folder, the new copy is created in the same folder as its parent, instead of the folder where the copy button was clicked.

I'll make an issue for it to triage.

sastels:
> Interesting behaviour in this feature.. (not related to this PR). Looks like if you click the copy button from within a folder, navigate to either root or a higher up folder, the new copy is created in the same folder as its parent, instead of the folder where the copy button was clicked.

Didn't think of that - the old behaviour was to always create the copy in the root folder, this PR changes that to create the copy where the original was, which, as you point out, could be confusing. Agree to sort it out in a new issue 👍

PR #589, opened 2020-06-04
Email 2fa using a code input instead of a link

brdunfield:
Closes: https://github.com/cds-snc/notification-api/issues/895

I essentially took the existing SMS 2fa pages / code and replicated that for emails.

As part of this, I changed existing variables / references to "sms_code" to "two_factor_code". I also renamed the SMS route from "two_factor" to "two_factor_sms" to keep things clearer. That's why there's so many file changes.

With regards to testing, I got rid of most of the existing email 2fa tests in test_two_factor.py, and rewrote them to match the existing SMS tests, and ported over any other relevant ones that weren't being tested for SMS.

This will probably need to be pulled down and tested with the api PR in tandem.


app/main/views/code_not_received.py
```diff
@@ -41,18 +41,19 @@ def check_and_resend_verification_code():
...
+
     return render_template('views/email-not-received.html')
 
 
 @main.route('/send-new-email-token', methods=['GET'])
 @redirect_to_sign_in
 def resend_email_link():
     user_api_client.send_verify_code(session['user_details']['id'], 'email', None)
-    session.pop('user_details')
+    # session.pop('user_details')
```
sastels, line 58: should this be commented out? If so we probably just want to 🔥 it

app/main/views/code_not_received.py
```diff
@@ -41,18 +41,19 @@ def check_and_resend_verification_code():
     if user.state == 'pending':
         return redirect(url_for('main.verify'))
     else:
-        return redirect(url_for('main.two_factor'))
+        return redirect(url_for('main.two_factor_sms_sent'))
 
 
-@main.route('/email-not-received', methods=['GET'])
+@main.route('/email-not-received', methods=['GET', 'POST'])
```
sastels, line 47: Have we added a POST as well as a GET?

app/main/views/new_password.py
```diff
@@ -25,14 +24,14 @@ def new_password(token):
         token_data = check_token(token, current_app.config['SECRET_KEY'], current_app.config['DANGEROUS_SALT'],
                                  current_app.config['EMAIL_EXPIRY_SECONDS'])
     except SignatureExpired:
-        flash(_('The link in the email we sent you has expired. Enter your email address to resend.'))
+        flash('The security code in the email we sent you has expired. Enter your email address to resend.')
```
sastels, line 27: Why get rid of the `_()`?

app/assets/javascripts/main.min.js
```diff
@@ -1,30 +1,36 @@
...
...n o(e,t,n){var a=this._calendar[e]||this._calendar.sameElse;return Object(r.a)(a)?a.call(t,n):a}},function(e,t,n){"use strict";n.d(t,"a",(function(){return a})),n.d(t,"b",(function(){return o}));var r=n(4),a={LTS:"h:mm:ss A",LT:"h:mm A",L:"MM/DD/YYYY",LL:"MMMM D, YYYY",LLL:"MMMM D, YYYY h:mm A",LLLL:"dddd, MMMM D, YYYY h:mm A"};function o(e){var t=this._longDateFormat[e],n=this._longDateFormat[e.toUpperCase()];return t||!n?t:(this._longDateFormat[e]=n.match(r.e).map((function(e){return"MMMM"===e||"MM"===e||"DD"===e||"dddd"===e?e.slice(1):e})).join(""),this._longDateFormat[e])}},function(e,t,n){"use strict";n.d(t,"a",(function(){return r})),n.d(t,"b",(function(){return a}));var r="Invalid date";function a(){return this._invalidDate}},function(e,t,n){"use strict";var r=n(67);e.exports=Function.prototype.bind||r},function(e,t,n){"use strict";n.d(t,"a",(function(){return f}));var r=n(44),a=n(45),o=n(46),i=n(41),s=n(42),l=n(15),u=n(32),c=n(19),d=n(36),f={calendar:r.b,longDateFormat:a.a,invalidDate:o.a,ordinal:i.b,dayOfMonthOrdinalParse:i.a,relativeTime:s.a,months:l.b,monthsShort:l.c,week:u.a,weekdays:c.a,weekdaysMin:c.b,weekdaysShort:c.c,meridiemParse:d.a}},function(e,t,n){"use strict";
```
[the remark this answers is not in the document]
brdunfield: This happened when I ran npm run webpack / build to get Steve M.'s skip link working after merging master. I can try to figure it out / get rid of it

app/main/views/new_password.py
brdunfield: Hmm. I'm not sure. I'm not even sure what page this is on or how to view it.

app/main/views/code_not_received.py
brdunfield: Good catch, I think this was leftover from an early attempt.

app/main/views/code_not_received.py
brdunfield: I left this in to remind me about some behaviour that has slightly changed. I'll comment it out.

Essentially, it used to be that you would enter credentials, and then if you clicked the "Not received an email" link, it would resend. If you clicked it again, it would direct you to re-enter password because of this line.

However, this seemed to break the code functionality, since we need to persist the user in this session, vs it being repopulated with the link? (I assume).

But the long and the short of it is you can keep clicking "not received an email" and sending new codes without having to re-enter your credentials again. I have no idea how much of an issue this is or if it's even an issue.

app/main/forms.py
```diff
@@ -511,17 +511,16 @@ def __init__(self, validate_code_func, *args, **kwargs):
         self.validate_code_func = validate_code_func
         super(TwoFactorForm, self).__init__(*args, **kwargs)
 
-    sms_code = SMSCode(_l('Please enter security code'))
+    two_factor_code = SMSCode(_l('Please enter the security code.'))
```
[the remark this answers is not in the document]
brdunfield: Probably yes

app/main/views/new_password.py
brdunfield: Found it, and I indeed broke it. I can't recall a reason why I took this out so I'll restore it.

sastels: There's a timeout (I think here https://github.com/cds-snc/notification-api/blob/master/app/config.py#L329) that expired the login link (and probably the code as well?)

PR #585, opened 2020-06-03
Review Translation CSV files line-by-line

anikbrazeau:
Comparative edit of English and French
Review for clarity and consistency of terms
Clean up files and add missing strings

PR #593, opened 2020-06-05
Fix error msgs fr

smcmurtry:
Closes https://github.com/cds-snc/notification-api/issues/891

I did not implement the translation: "This organisation name is already in use" because I realized this was platform admin only and I wasn't able to test the change.

I was not able to trigger all of these errors so most of these changes are untested. But here are screenshots of what I was able to reproduce:

Screen Shot 2020-06-05 at 10 56 04 AM

Screen Shot 2020-06-05 at 2 17 09 PM

Also - I removed the `<i>` tags because they were not being rendered in the error messages.


app/translations/csv/fr.csv
```diff
@@ -101,7 +101,7 @@
 "What’s their email address?","Quelle est son adresse courriel?"
 "Password is blacklisted.","Le mot de passe est sur une liste noire."
 "Not a csv file","Il ne s’agit pas d’un fichier CSV"
-"isn’t a spreadsheet that Notification can read","n’est pas une feuille de calcul que <i>Notification</i> peut lire"
+"isn’t a spreadsheet that Notification can read","n’est pas une feuille de calcul que Notification peut lire"
```
brdunfield, line 104: Does the English string for this still use 'Notification'? (Since it's in the key)

smcmurtry: Yes - I think @anikbrazeau's PR (#585) will fix that. This change is unrelated to the other ones - I just happened to see the french error message here and the `<i>` tags were visible along with the other text, so I removed them. But if you want I can do this as a separate PR.

app/main/views/organisations.py
```diff
@@ -105,8 +106,7 @@ def invite_org_user(org_id):
             org_id,
             email_address
         )
-
-        flash('Invite sent to {}'.format(invited_org_user.email_address), 'default_with_tick')
+        flash('{} {}'.format(_('Invite sent to'), invited_org_user.email_address), 'default_with_tick')
```
brdunfield, line 109: This worked, so the others with this formatting should as well 👍

brdunfield:
If it's being addressed in Anik's PR then ignore my comment about the key!

PR #598, opened 2020-06-08
Fix french grammar on the CSV upload text

smcmurtry:
Closes https://github.com/cds-snc/notification-api/issues/880
Closes https://github.com/cds-snc/notification-api/issues/931

Fixing the grammar wasn't too difficult so I just kept the existing text.  

Screen Shot 2020-06-08 at 2 12 24 PM

Screen Shot 2020-06-08 at 2 12 39 PM


app/templates/views/s3-send.html
```diff
@@ -6,15 +6,15 @@
 {% from "components/select-input.html" import select_wrapper %}
 {% from "components/form.html" import form_wrapper %}
 {% from "components/message-count-label.html" import recipient_count_label %}
-{% set txt = _('Upload a list of') + " " + recipient_count_label(999, template.template_type) + " from Amazon S3" %}
+{% set txt = template.template_type|get_csv_upload_text %}
 
 {% block service_page_title %}
-  {{ txt }}
+  {{ txt + " from Amazon S3" }}
```
brdunfield, line 12: does "from Amazon S3" need translating?

app/translations/csv/en.csv
```diff
@@ -1198,7 +1196,11 @@ Canada Notification service",""
 "{} contains numbers or dates that Notify can’t understand. Try formatting all columns as ‘text’ or export your file as CSV.",""
 "Are you sure you want to remove security key {}?",""
 "You can't accept an invite for another person.",""
+"Upload a list of email addresses",""
+"Upload a list of phone numbers",""
+"Upload a list of recipients",""
+"Not a valid international number","Enter a valid phone number"
```
brdunfield, line 1202: looks like this key got duplicated

smcmurtry: good catch!

app/templates/views/s3-send.html
smcmurtry: Yes, there is an issue for that: https://github.com/cds-snc/notification-api/issues/931

👍  I will add those into this PR

PR #599, opened 2020-06-08
Add a helper function to get the `and n more` text on the service settings page

smcmurtry:
This is part of https://github.com/cds-snc/notification-api/issues/848 but not the whole thing.

This translates the `...and n more` text that shows up if you have multiple reply to email addresses:

Screen Shot 2020-06-08 at 2 55 54 PM

Screen Shot 2020-06-08 at 2 56 43 PM

Also the logic is the same for SMS sending numbers, so I also fixed that.


app/__init__.py
```diff
@@ -509,6 +509,19 @@ def format_notification_status_as_url(status, notification_type):
     }.get(notification_type)
 
 
+def get_and_n_more_text(number_of_addresses):
+    "number_of_addresses could be email addresses or sms sending numbers"
```
brdunfield, line 513: What is this line?

smcmurtry: That's just a comment that describes the function - a convention in python is to write documentation for functions as a string below the function definition. Here's a [link discussing that](https://stackoverflow.com/questions/708649/python-comments-vs-strings) but there are probably better ones.

brdunfield: Oh Interesting! Thanks for the link, didn't know this!

PR #600, opened 2020-06-08
Add translations for the 2 default email brandings

smcmurtry:
This and #599 close https://github.com/cds-snc/notification-api/issues/848.

~~Will keep as a draft until I have the correct translations.~~ ✅ 

I have manually tested that the correct branding is still getting applied to emails.


app/templates/views/service-settings.html
```diff
@@ -331,7 +331,7 @@ <h2 class="heading-medium">{{ _('Platform admin settings') }}</h2>
           {% call row() %}
             {% set txt = _('Email branding') %}
             {{ text_field(txt)}}
-            {{ text_field(current_service.email_branding_name) }}
+            {{ text_field(_(current_service.email_branding_name)) }}
```
brdunfield, line 334: I like this! Only have to wrap once 👍

brdunfield:
I will log this as a separate issue, but just want to flag that it's difficult to tell which branding is being used from the main settings page
Screen Shot 2020-06-11 at 1 56 07 PM

PR #605, opened 2020-06-10
add tests around our translation files

sastels:
add a script that checks if our csv files have extra or missing lines.


sastels:
replaced by #625

PR #614, opened 2020-06-12
Add email domains for pen tester

sastels:
closes https://github.com/cds-snc/notification-api/issues/960

add gmail.com and strivesec.com domains for pen testing on staging.

changes only applied if `NOTIFY_ENVIRONMENT` is `development`, so production is safe.


sastels:
> 👍
> No pushes to prod until this is reverted.

We should be fine to push this code to prod - it'll only allow the extra domains on prod, heroku, and staging (of course if we do push this code to prod we should actually test this)

PR #618, opened 2020-06-15
skip template tutorial

sastels:
resolves https://github.com/cds-snc/notification-api/issues/901

New users are prompted to create a service, then instead of going to the example template they go directly to the service dashboard:

image

### DIFF
--- a/app/main/views/add_service.py
+++ b/app/main/views/add_service.py
@@ -66,16 +66,15 @@ def add_service():
         )
         if error:
             return render_template('views/add-service.html', form=form, heading=heading)
-        if len(service_api_client.get_active_services({'user_id': session['user_id']}).get('data', [])) > 1:
-            return redirect(url_for('main.service_dashboard', service_id=service_id))
 
-        example_email_template = _create_example_template(service_id)
+        return redirect(url_for('main.service_dashboard', service_id=service_id))
 
-        return redirect(url_for(
-            'main.start_tour',
-            service_id=service_id,
-            template_id=example_email_template['data']['id']
-        ))
+        # example_email_template = _create_example_template(service_id)
+        # return redirect(url_for(
+        #     'main.start_tour',
+        #     service_id=service_id,
+        #     template_id=example_email_template['data']['id']
+        # ))
     else:
         return render_template(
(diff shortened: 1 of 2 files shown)


app/main/views/add_service.py
```diff
@@ -66,16 +66,15 @@ def add_service():
...
-            'main.start_tour',
-            service_id=service_id,
-            template_id=example_email_template['data']['id']
-        ))
+        # example_email_template = _create_example_template(service_id)
+        # return redirect(url_for(
+        #     'main.start_tour',
+        #     service_id=service_id,
+        #     template_id=example_email_template['data']['id']
+        # ))
```
sastels, line 77: I'm torn on commenting this out vs deleting it. Are we expecting to add this back in, or are we just going to spiff up the docs so users can figure it out on their own?

tests/app/main/views/test_add_service.py
```diff
@@ -224,7 +223,6 @@ def test_should_add_service_and_redirect_to_dashboard_when_existing_service(
             _external=True,
         )
     )
-    assert mock_get_services.called
```
sastels, line 227: We don't call this now since we're not checking if it's the user's first service anymore.

sastels:
> I'm still seeing the onboarding tutorial when I create a new service in the heroku deployment?

ugh, for some reason the email link you get sent during registration has the link to staging 😞 
No idea why this is happening...

brdunfield:
Confirming it works when making sure you click the right registration link.
What I saw:
- Created a new account
- Received email with link
- Replaced staging part of link with PR base URL and navigated there
- screen requested SMS code verification
- Inputted code and was brought to new Service creation screen
- Created a new service and was immediately brought to the dashboard for that service.

sastels:
> Just would like to make sure it's clear / documented (maybe on the original issue?) that we have decided to 'hide' rather than 'remove' the tutorial.
> 
> (Down the line we may want to take out the code if we decide it won't be used in any way)

I took out the code that generated the tutorial template (as it's now not used) but left in the "start-tour" route, as it's called elsewhere in the code as well depending on and index value (?!)

Will note this in the issue.

PR #621, opened 2020-06-19
Remove pricing link from homepage

anikbrazeau:
Removing the link to the pricing page, which adds no additional information.

Same info (about it being free) is found on Emails/Text message pages in features and on homepage.


brdunfield:
Should we add a bullet indicating there are limits on text messages in lieu of the old link? There's a bullet about unlimited emails but nothing indicating that texts have a limit.

PR #624, opened 2020-06-19
Translate error messages and wrap strings coming from the API

brdunfield:
I wrapped the incoming error messages from the API (for 400 errors that redirect to 400.html). So, if we find any missed strings from the API, we should just need to add the translations to the CSV files and they'll show up.

Also updated some other error strings.

Closes https://github.com/cds-snc/notification-api/issues/953


app/__init__.py
```diff
@@ -642,10 +642,13 @@ def render_http_error(error):
...
         error_code = error.status_code
         if error_code == 400:
+            # all incoming 400 errors from the API are wrapped for translation
+            # Need to make sure all of them have translations in the csv files
             if isinstance(error.message, str):
-                msg = [error.message]
+                msg = [_(error.message)]
             else:
-                msg = list(itertools.chain(*[error.message[x] for x in error.message.keys()]))
+                msg = list(itertools.chain(_(error.message[x]) for x in error.message.keys()))
```
sastels, line 650: not sure why the `itertools.chain` is here - we only have one iterable inside I think? Works though.

PR #629, opened 2020-06-22
delete unused pricing content

sastels:
Fixes https://github.com/cds-snc/notification-api/issues/920

Delete the "How to Pay" page that we are not currently using, as well as commented out code from the pricing page.


app/templates/views/pricing/index.html
```diff
@@ -33,122 +33,5 @@ <h2 class="heading-medium">{{ _('Emails') }}</h2>
...
-  <ul class="list list-bullet">
-    <li>{{ _('250,000 free text messages for central government services') }}</li>
-    <li>{{ _('25,000 free text messages for other public sector services') }}</li>
-  </ul>
-
-  <h3 class="heading-small">{{ _('Long text messages') }}</h3>
-  <p>{{ _('If a text message is longer than 160 characters (including spaces), it’ll be charged as more than one message:') }}</p>
-  <div class="bottom-gutter-3-2">
-    {% call mapping_table(
-      caption=_('Text message pricing'),
```
brdunfield, line 46: Do any of these strings need to be removed from the csv files?

sastels: Luckily (?) we had never actually translated any of these strings.

PR #630, opened 2020-06-22
Fix bad redirect with fido2 key

brdunfield:
Also stops email codes from errantly being sent when a fido2 key is being used

### DIFF
--- a/app/assets/javascripts/fido2.js
+++ b/app/assets/javascripts/fido2.js
@@ -36,7 +36,7 @@
         })
         .then(function(response) {
           if (response.ok) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/fido2.js
```diff
@@ -36,7 +36,7 @@
         })
         .then(function(response) {
           if (response.ok) {
-            window.location = "/two-factor-sms" + window.location.search;
+            window.location = "/two-factor-sms-sent" + window.location.search;
```
brdunfield, line 39: two-factor-sms didn't exist anymore

PR #634, opened 2020-06-24
Fix Email resent content

anikbrazeau:
Minor copy tweaks for clarity on screenreader on Email resent page:
https://notification.alpha.canada.ca/email-not-received

Fixes https://github.com/cds-snc/notification-api/issues/967


anikbrazeau:
Fixes https://github.com/cds-snc/notification-api/issues/967

anikbrazeau:
@brdunfield I found a few more instances with "resend" and fixed those too, tests passing locally 🤞 if you can give it another quick glance before I merge.

app/templates/views/notifications/notification.html
```diff
@@ -49,7 +49,7 @@
         </p>
       {% elif notification_status == 'technical-failure' %}
         <p class="notification-status-cancelled">
-          Technical failure – Notification will resend once the team have
+          Technical failure – GC Notify will re-send once the team have
```
brdunfield, line 52: I don't see this string in the csvs below.

brdunfield: Oh, I guess this isn't even wrapped for translation!

brdunfield: We should probably wrap this and add translations to the csvs :)

anikbrazeau: Good call!

PR #646, opened 2020-06-29
Use new images

sastels:
resolves https://github.com/cds-snc/notification-api/issues/934

~Right now I've uploaded the new images with a "v2" suffix to the names. If we don't care about looking at old versions of Notify with the old images, we don't need this PR and can just rename them over top of the old files and they'll automatically be in prod.~

Old images in S3 have been replaced with the new ones, so no code changes are needed. Here I've just updated the images in the repo.

old:
image
image

new:
image
image


sastels:
New French image uploaded:
image

PR #653, opened 2020-07-03
block user after 10 failed passwords

sastels:
resolves https://github.com/cds-snc/notification-api/issues/986

automatically block a user if they've had too many password fails in a row.


brdunfield:
This looks great code-wise, but I think we need to adjust / remove the email being sent out to the user when their account is locked.

Screen Shot 2020-07-06 at 11 04 18 AM

PR #680, opened 2020-07-21
word wrap name of branding

sastels:
fixes https://github.com/cds-snc/notification-api/issues/956

Now:
image

Default behaviour in table cells is supposed to be to not truncate the text, but it was getting truncated anyway 😞

### DIFF
--- a/app/assets/stylesheets/components/table.scss
+++ b/app/assets/stylesheets/components/table.scss
@@ -9,7 +9,6 @@
 }
 
 .dashboard-table {
-
   .heading-medium {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/components/table.scss
```diff
@@ -9,7 +9,6 @@
 }
 
 .dashboard-table {
-
```
sastels, line 12: All these changes were done by prettier to correct the code style, except for the last 4 lines I added.

PR #684, opened 2020-07-23
Replace most references to "whitelist" and "blacklist"

sastels:
re https://github.com/cds-snc/notification-api/issues/962

replace "whitelist" with "safelist" and "blacklist" with "blocklist"

Still a few remaining that require simultaneous notification_api and notification_utils changes. Will do those in a separate PR.


timarney:
The follow up PR for https://github.com/cds-snc/notification-admin/pull/98 ;)

PR #687, opened 2020-07-24
Translate missing strings

brdunfield:
Closes https://github.com/cds-snc/notification-api/issues/944

Note: I have separated out notification-utils strings into a separate card, as it looks like translating them will require some creative thinking.


tests/app/main/test_errors_for_csv.py
```diff
@@ -60,11 +60,13 @@
     ]
 )
 def test_get_errors_for_csv(
+    app_,
```
sastels, line 63: just wondering why you had to add this?

brdunfield: For whatever reason, translating these strings (babel?) made the tests complain about not having a request context.

PR #695, opened 2020-07-28
Remove /support and associated routes

brdunfield:
Closes https://github.com/cds-snc/notification-api/issues/1021

While removing /support, I also found `/support/bat-phone` and `/support/report-a-problem`, so I removed those as well (With Anik's confirmation).

This ended up being a fairly complex PR.

A couple things I'm not sure about:
- I've removed the test_feedback.py suite. It looks like it was only testing features we don't actually use (like out of hours emergencies, holidays, etc).
- That said, I'm wondering if any tests around the existing feedback form need to be added?
- Also, main.thanks has been totally removed from navigation, which I don't understand. `/support/thanks` is a live link that loads / is redirected to by multiple places, however it was throwing errors if it was in any of the exclude lists in navigation...


app/main/views/feedback.py
```diff
@@ -72,106 +42,3 @@ def feedback(ticket_type):
...
-
-
-def in_business_hours():
-
-    now = datetime.utcnow().replace(tzinfo=pytz.utc)
-
-    if is_weekend(now) or is_bank_holiday(now):
-        return False
-
-    return london_time_today_as_utc(9, 30) <= now < london_time_today_as_utc(17, 30)
```
sastels, line 99: lol @ us. oh dear.

scripts/run_tests.sh
```diff
@@ -35,5 +35,5 @@ npm test
 display_result $? 3 "Front end code style check"
 
 ## Code coverage
-py.test -n4 --maxfail=1 tests/ --strict -p no:warnings
+py.test -n4 --maxfail=10 tests/ --strict -p no:warnings
```
sastels, line 38: Was this for debugging or just something to add because it's a pain when you wait 10 minutes and only learn about one failure?

brdunfield: :( Not intentional. Just accidental commit of my debug setup. Good catch

PR #712, opened 2020-08-06
Users switch branding

sastels:
resolves https://github.com/cds-snc/notification-api/issues/977

Give users the ability to switch to English or French default brandings, as well as upload their own.

image

image

If an image is uploaded the user gets:

image

image


app/notify_client/email_branding_client.py
```diff
@@ -11,9 +11,12 @@ def get_email_branding(self, branding_id):
...
 
-        if sort_key and sort_key in brandings[0]:
-            brandings.sort(key=lambda branding: branding[sort_key].lower())
-        return brandings
+        if len(brandings) > 0:
+            if sort_key and sort_key in brandings[0]:
+                brandings.sort(key=lambda branding: branding[sort_key].lower())
+            return brandings
+        else:
+            return []
```
sastels, line 19: This fixes a crash if there are no custom brandings (ie you couldn't create the first one).

app/main/views/service_settings.py
```diff
@@ -1098,31 +1098,63 @@ def link_service_to_organisation(service_id):
 
 
 @main.route("/services/<service_id>/branding-request/email", methods=['GET', 'POST'])
-@main.route("/services/<service_id>/branding-request/email/<path:logo>", methods=['GET', 'POST'])
```
sastels, line 1101: `logo` was not being used

brdunfield:
The "Save" and "Upload logo" buttons still appear in English when the site is set to French

brdunfield:
Screen Shot 2020-08-11 at 12 32 55 PM
Also we have a strange and also untranslated error message here. CC @anikbrazeau

anikbrazeau:
Good catch @brdunfield! 👍 

For the error messages, could we have them be prevented from uploading a file other than PNG in the first place to avoid the error message altogether? 

If not perhaps something like this similar to the height error message?

EN: Your logo must be an image in PNG format
FR: Votre logo doit être une image au format PNG

EN: Your logo must be exactly 108 pixels tall
FR: Votre logo doit avoir exactement 108 pixels de hauteur

anikbrazeau:
Is the line with "Type of logo" that appears above the radio buttons new/necessary?

Screen Shot 2020-08-11 at 3 20 40 PM

sastels:
> Is the line with "Type of logo" that appears above the radio buttons new/necessary?
hmm I might have added that? I think we need to have some sort of name for the radio button group for screen readers.

anikbrazeau:
Is it possible to remove the heading and sentence above and use this as the name above the radio buttons instead?

EN **Use the federal government signature**
FR **Utilisez la signature du gouvernement fédéral**

sastels:
> The "Save" and "Upload logo" buttons still appear in English when the site is set to French

the Upload logo was missing the `_()`, but the Save wasn't - it is being overwritten by
https://github.com/cds-snc/notification-admin/blob/master/app/assets/javascripts/previewPane.js#L34

fail fail fail. will delete that line and see if everything breaks... 🤞

sastels:
> The spacing on the settings page is inconsistent with the design. There needs to be more between the "sections", because otherwise it looks like a double-border effect

We can probably change the spacing or just have a single line rather than a double - I'd say do it in a new issue / PR though, as otherwise this PR doesn't touch the main settings page.

app/assets/javascripts/previewPane.js
```diff
@@ -31,7 +31,6 @@
   $paneWrapper.append($previewPane);
   $form.find('.grid-row').eq(0).prepend($paneWrapper);
   $form.attr('action', location.pathname.replace(new RegExp(`set-${previewType}-branding$`), `preview-${previewType}-branding`));
-  $form.find('button[type="submit"]').text('Save');
```
brdunfield, line 34: Do we know if this adversely affects any other pages?

brdunfield:
LGTM on the code side. Looks like the 'Type of Logo' discussion is ongoing, so I'll hold off on approving for now.

app/assets/javascripts/previewPane.js
sastels: 🤷 Any buttons that were bilingual before won't be impacted, which hopefully was all of the user facing ones. There could be some platform admin ones that I didn't come across when playing around.

katedee:
Looks good to me, thank you Steve!
There's still the spacing issue on the settings page, but I guess you're handling that in a separate PR?

PR #754, opened 2020-09-14
Remove extra character in template

AntoineAugusti:
There is an extra character (`'`) at the end of the translated string in the template, removing it.

> The text message sender tells your users who the message is from.'

Screen Shot 2020-09-14 at 11 22 12


brdunfield (approved):
Good catch!
👍

PR #759, opened 2020-09-16
Reset failed login attempts when unblocking users

AntoineAugusti:
This PR fixes https://github.com/cds-snc/notification-api/issues/1044

There is no sufficient test coverage for blocking/unblocking users in the existing test suite. Did my best to add a test proving that when unblocking a user, the failed login attempts counter is reset as well.


brdunfield (approved):
Works great! Great job with the test 👍

PR #764, opened 2020-09-17
Validate email addresses with special characters

AntoineAugusti:
Fixes https://github.com/cds-snc/notification-api/issues/760

Need https://github.com/cds-snc/notification-utils/pull/60 to be merged + released first

### DIFF
--- a/app/main/validators.py
+++ b/app/main/validators.py
@@ -92,8 +92,6 @@ class ValidEmail(Email):
         except InvalidEmailError:
             raise ValidationError(_l(self.message))
 
-        return super().__call__(form, field)
-
 
 class NoCommasInPlaceHolders:
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -38,7 +38,7 @@ awscli-cwlogs>=1.4,<1.5
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous==0.24  # pyup: <1.0.0
 
-git+https://github.com/cds-snc/notifier-utils.git@43.0.1#egg=notifications-utils==43.0.1
+git+https://github.com/cds-snc/notifier-utils.git@43.0.2#egg=notifications-utils==43.0.2
 
 rsa>=4.1 # not directly required, pinned by Snyk to avoid a vulnerability
 
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -36,6 +36,6 @@ awscli-cwlogs>=1.4,<1.5
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous==0.24  # pyup: <1.0.0
 
-git+https://github.com/cds-snc/notifier-utils.git@43.0.1#egg=notifications-utils==43.0.1
+git+https://github.com/cds-snc/notifier-utils.git@43.0.2#egg=notifications-utils==43.0.2
 
 rsa>=4.1 # not directly required, pinned by Snyk to avoid a vulnerability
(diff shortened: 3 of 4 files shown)


requirements.txt
```diff
@@ -38,7 +38,7 @@ awscli-cwlogs>=1.4,<1.5
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous==0.24  # pyup: <1.0.0
 
-git+https://github.com/cds-snc/notifier-utils.git@43.0.1#egg=notifications-utils==43.0.1
+git+https://github.com/cds-snc/notifier-utils.git@43.0.2#egg=notifications-utils==43.0.2
```
AntoineAugusti, line 41: Already updating requirements to the next version.

Because this version is not released yet, CI will fail

app/main/validators.py
```diff
@@ -92,8 +92,6 @@ def __call__(self, form, field):
         except InvalidEmailError:
             raise ValidationError(_l(self.message))
 
-        return super().__call__(form, field)
```
AntoineAugusti, line 95: Don't need to call to the WTFForm's `Email` class as the custom validator is inspired by it already + whitelists more stuff. If we were to keep this call, we would not be able to do allow special characters in emails

AntoineAugusti:
Merged the PR on utils this morning and reran the build here

brdunfield:
I guess the one question I have here is whether we do want to except quite this number of special characters. Obviously we've already seen a use-case for ', but do we really want to allow an email like "2+2={5*4/5}@mailinator.com" through?

CC @willeybryan @obrien-j 

The code itself looks great :)

AntoineAugusti:
I took some weird email addresses @willeybryan gave me https://github.com/cds-snc/notification-utils/pull/61

willeybryan:
According RFC 5322 and 5321 the Local part of the email (local-part@domain.topleveldomain) may include any of the following Ascii characters
- uppercase and lowercase Latin letters A to Z and a to z
- digits 0 to 9
- printable characters !#$%&'*+-/=?^_`{|}~
- dot .  (provided that it is not the first or last character and provided also that it does not appear consecutively (e.g., John..Doe@example.com is not allowed).)

Furthermore, and this part I left out @brdunfield because I felt like you feel about the 2+2={5*4/5}, and it was maddening to read. 

If quoted, it may contain Space, Horizontal Tab (HT), any ASCII graphic except Backslash and Quote and a quoted-pair consisting of a Backslash followed by HT, Space or any ASCII graphic; it may also be split between lines anywhere that HT or Space appears. In contrast to unquoted local-parts, the addresses ".John.Doe"@example.com, "John.Doe."@example.com and "John..Doe"@example.com are allowed.

AntoineAugusti:
Wait, only latin letters?

> uppercase and lowercase Latin letters A to Z and a to z

By reading again the issue and some test cases you provided, I thought that the goal was to allow letters with accents like `é`. Am I wrong?

willeybryan:
Oops! I thought é was a latin character and as for practical goals Gmail allows it so we should too. 

Anyway yes, that was the goal, to make the email more inclusive of names we have encountered. I just made this one based on the RFC 2+2={5*4/5}@mailinator.com thinking it should work too, but maybe it doesn't because of the / or {} and that is ok too, But the other stuff for sure is important.

brdunfield (approved):
No, that's fine, I just wanted to confirm that we were okay with such crazy emails being allowed, and it sounds like we are, so I will approve this! :)

AntoineAugusti:
From Wikipedia
> The 2012 standards RFC 6532 and RFC 6531 allow the inclusion of Unicode characters in a header content using UTF-8 encoding, and their transmission via SMTP - but in practice support is only slowly rolling out

So yeah, we know support accentuated letters and a single quote and we have tests to prove it. As a side effect, this super weird email address is also valid but I guess it's not relevant for us.

Good to ship?

PR #765, opened 2020-09-17
Build the correct endpoint for the feedback confirmation page

brdunfield:
Likely broke here: https://github.com/cds-snc/notification-admin/commit/9f74324fc3a9ff490c8d41fd6054d0202d52da2b

> tl:dr is that the thanks route actually doesn’t exist anymore post that /support removal PR i mentioned - instead we just have to point to .feedback and pass in the “thanks” ticket type to get the correct page render


AntoineAugusti (approved):
Tested locally, 💯

PR #771, opened 2020-09-22
Translate placeholders in template

AntoineAugusti:
Fixes https://github.com/cds-snc/notification-api/issues/850

Add more strings to template more fields in templates. Needs https://github.com/cds-snc/notification-utils/pull/62 to be merged and released first. CI will fail before that.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -374,7 +374,10 @@ def translate_preview_template(_template_str):
         "Subject": _("Subject"),
         "Reply to": _("Reply to"),
         "From:": _("From:"),
-        "To:": _("To:")
+        "To:": _("To:"),
+        "phone number": _("phone number"),
+        "email address": _("email address"),
+        "hidden": _("hidden"),
     }
 
     def translate_brackets(x):
--- a/app/translations/csv/en.csv
+++ b/app/translations/csv/en.csv
@@ -1182,3 +1182,4 @@
 "Email authentication",""
 "Send files by email",""
 "Uploading letters",""
+"hidden",""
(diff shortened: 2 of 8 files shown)


app/translations/csv/en.csv
```diff
@@ -1169,3 +1169,4 @@
 "Must be a valid phone number",""
 "Not enough digits",""
 "Not a valid country prefix",""
+"hidden",""
```
brdunfield, line 1172: I am not the biggest fan of having test data inside our app translation csv - is there any way to declare this within the test itself?

app/__init__.py
```diff
@@ -374,7 +374,10 @@ def translate_preview_template(_template_str):
         "Subject": _("Subject"),
         "Reply to": _("Reply to"),
         "From:": _("From:"),
-        "To:": _("To:")
+        "To:": _("To:"),
+        "phone number": _("phone number"),
+        "email address": _("email address"),
+        "hidden": _("hidden"),
```
brdunfield, line 380: Again, if there's any way to avoid the "hidden" test key being in here, that would be preferable..

AntoineAugusti: This is not a test key, previously it was not translated in placeholders (as email address/phone number) and not it's going to be translated https://github.com/cds-snc/notification-utils/pull/62/files#diff-968254ab44ec963a95d7cedaf3f53715R61

brdunfield: Oh, thank you, I missed that. I just saw in in the tests and incorrectly thought you had added it. Whoops!

brdunfield (approved):
LGTM! 👍 🦑

PR #776, opened 2020-09-24
Adding a new docker compose file to help reproduce prod env.

PierrePaul:
https://trello.com/c/ZUQWhMtW/37-new-docker-compose-file

Adds a docker-compose file to help get closer to the production environment.


docker-compose.yml
```diff
@@ -0,0 +1,20 @@
...
+    image: notification-admin
+    build:
+      context: ./
+      dockerfile: ci/Dockerfile
+    command: >
+      bash -c "
+        make babel &&
+        npm install &&
+        npm run build &&
+        make generate-version-file &&
```
AntoineAugusti, line 15: Curious why you need to run these commands again as they appear to be executed already in the original Dockerfile

https://github.com/cds-snc/notification-admin/blob/86e36297a2f5ea6dde821b10bff40514943a3d5f/ci/Dockerfile#L18-L23

AntoineAugusti (changes requested):
Great work! Could you add instructions in the README about using this as an alternative way to run the app?

docker-compose.yml
PierrePaul: The command run in the dockerfile will only be run on `docker-compose build`. For development purpose, you will want to run them more often. By putting them in the overridden command in the docker-compose.yml, those commands will be run everytime you do a `up`. 
The `generate-version-file` command though doesnt need to be run more than once though, so I will take it out.

AntoineAugusti (approved):

PR #780, opened 2020-09-30
Limit local part of email address to 64 bytes

AntoineAugusti:
https://github.com/cds-snc/notification-api/issues/1110
https://trello.com/c/yiFU8McP/44-service-name-can-make-an-email-too-long-restrict-length

If the service name is very long in both languages, separated by a pipe, use only the first part to generate the email address
- _Health Promotion And Chronic Diseases Prevention Branch HPCDP | Direction générale de la promotion de la santé et de la prévention des maladies chroniques PSPMC_ => `health.promotion.and.chronic.diseases.prevention.branch.hpcdp`

If we can't split "nicely" the too long original string, trim to a maximum of 64 bytes

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -243,14 +243,19 @@ def generate_previous_next_dict(view, service_id, page, title, url_args):
 
 
 def email_safe(string, whitespace='.'):
-    # strips accents, diacritics etc
+    original = string
+    # Strips accents, diacritics etc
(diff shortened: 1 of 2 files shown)


AntoineAugusti:
Closing as it doesn't fit requirements.

> We can't trim the generated email addresses because for Government of Canada official language act legal requirements, the email has to be in both English and French. Which is why these emails are so long and hard to read.

PR #783, opened 2020-10-06
Import tailwind

brdunfield:
This PR imports the work done on https://github.com/cds-snc/notification-homepage into this PR.

https://trello.com/c/XF2KTVt0/51-merge-home-page-into-notification-admin-repo
https://github.com/cds-snc/notification-api/issues/1134

I have included the necessary npm dev dependencies from notification-homepage, and included / adjusted the relevant config files. style.css is a direct copy from notification-homepage, as are the postcss and tailwind config files.

The resulting homepage.css is larger than what master outputs, likely because purgecss is scanning *all* html files, instead of just the homepage.

This is probably good enough for now. We can adjust this config if / when we decide to start splitting stylesheets.

NPM is complaining about several outdated dependencies, but in order to keep things clean and not cause issues with Antoine's dependency bump PR I have left them as they are.

Note: Running `npm run webpack` also generates a large amount of changes in main.min.js as a result of a moment.js dependency upgrade. We don't want these changes, as they are currently breaking, so we'll need to make sure they are not committed or included in `npm run build` as style updates are made in the future. I've logged this as an issue to fix here: https://github.com/cds-snc/notification-api/issues/1150


postcss.config.js
```diff
@@ -0,0 +1,27 @@
+const cssnano = require("cssnano");
+
+console.log("=== POST CSS  ===");
```
AntoineAugusti, line 3: I think you can remove this

app/assets/stylesheets/tailwind/style.css
```diff
@@ -0,0 +1,154 @@
+@tailwind base;
```
AntoineAugusti, line 1: I understand this file is a straight copy/paste for now. Do you plan to change it later? I think it's not Tailwind's spirit to add custom classes like `feature-box`, but instead, rely on utilities as much as possible.

It would be a good idea to extract the CSS where we apply classes to a different CSS file as well.

AntoineAugusti (approved):
Good work! 👏 

I think you can update your PR description to reflect that there are no breaking changes for moment as we discovered yesterday.

I'm surprised there are no HTML files in your PR, are they already in this repo?

app/assets/stylesheets/tailwind/style.css
brdunfield: Yes, my next task is to figure out how we want to split up the CSS, do components, etc, and adjust the source / target input / outputs accordingly.

brdunfield:
The HTML is already in this repo, yes. It's just the css. The markup generated by notification-homepage has been copied here previously, and no changes have been made since then. Now that the css build is happening locally, we should be able to modify the markup within this repo and see the css changes reflected in the css here.

AntoineAugusti:
Awesome, thanks for the clarification!

:shipit:

PR #786, opened 2020-10-09
Port button styles to tailwind

brdunfield:
This PR removes `_buttons.scss` and ports its styles into tailwind.

There is also some tailwind file structure renaming / cleanup:
- Renamed homepage.css to index.css. (main.css is currently still taken by the old code)

Note: There are a couple instances where mobile button styles differ from the existing. That said, we don't seem to have a good or ANY rules about mobile buttons. Some are full width, some are not, and I don't think that it's worth putting work at this time into perfectly replicating staging. If / when we get around to auditing mobile, we can homogenize all the styles at that time.


app/assets/stylesheets/tailwind/layout.css
```diff
@@ -1,5 +1,2 @@
 @layer components {
-  .container {
```
brdunfield, line 2: This style was not actually being used because importing other files was broken. .container is a base tailwind util class.

app/assets/stylesheets/tailwind/tailwind.css
```diff
@@ -0,0 +1,3 @@
+@tailwind base;
```
brdunfield, line 1: Moved these imports into its own file to preserve code order

package.json
```diff
@@ -69,6 +69,7 @@
     "jshint": "2.11.0",
     "jshint-stylish": "2.2.1",
     "mini-css-extract-plugin": "^0.9.0",
+    "postcss-import": "^12.0.1",
```
brdunfield, line 72: This was needed to allow other css files to actually be imported in style.css

app/assets/javascripts/apiKey.js
```diff
@@ -8,11 +8,11 @@
     const states = {
       'keyVisible': (key, thing) => `
         <span class="api-key-key">${key}</span>
-        <input type='button' class='api-key-button-copy' value='${window.polyglot.t('copy')} ${thing} ${window.polyglot.t('to_clipboard')}' />
+        <input type='button' class='api-key-button-copy button-secondary' value='${window.polyglot.t('copy')} ${thing} ${window.polyglot.t('to_clipboard')}' />
       `,
       'keyCopied': thing => `
         <span class="api-key-key">${window.polyglot.t('copied_to_clipboard')}</span>
-        <input type='button' class='api-key-button-show' value='${window.polyglot.t('show')} ${thing}' />
+        <input type='button' class='api-key-button-show button-secondary' value='${window.polyglot.t('show')} ${thing}' />
```
AntoineAugusti, line 15: It seems like `api-key-button-copy` and `api-key-button-show` are classes used by JS to perform an action.

In the past I named these classes only used by JS like `js-class-name` to point that deleting them has an impact on features and not on the UI. What do you think of this? Would you like to make these changes as you go?

app/assets/javascripts/fileUpload.js
```diff
@@ -7,8 +7,8 @@
 
     this.cancelText = window.polyglot.t("cancel_upload")
 
-    this.showCancelButton = () => $('.file-upload-button', this.$form).replaceWith(`
-      <a href="" class='file-upload-button-cancel'>${cancelText}</a>
+    this.showCancelButton = () => $('#file-upload-button', this.$form).replaceWith(`
```
AntoineAugusti, line 10: Same remark about `js-` prefix

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
+@layer components {
+
+  /* Note: Buttons are <a> sometimes, which is why :link and : visted styles exist */
+
+  .button, .button:link, .button:visited {
+    padding: .526315em .789473em .263157em;
```
AntoineAugusti, line 6: Can you replace this with Tailwind padding classes?

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
...
+    background-color: theme("colors.green.default");
```
AntoineAugusti, line 7: Any reason to write this instead of `bg-green`?

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
...
+    box-shadow: 0 2px 0 theme("colors.green.border");
+    @apply no-underline inline-block text-white text-smaller leading-tight align-top;
+  }
+  
+  .btn-blue, .btn-blue:link, .btn-blue:visited {
+    background-color: theme("colors.blue.lighter");
```
AntoineAugusti, line 13: `bg-blue-lighter`?

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
...
+    @apply px-5 py-4 border-b-2 border-black font-bold text-black text-base;
+  }
+  
+  .button:hover {
+    background-color: theme("colors.green.hover");
```
AntoineAugusti, line 18: `hover:bg-green-hover`?

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
...
+    outline: 3px solid theme("colors.yellow.default");
+  }
+
+  .button-red {
+    background-color: theme("colors.red.default");
+    box-shadow: 0 2px 0 theme("colors.red.border");
+  }
+  .button-red:hover {
+    background-color: theme("colors.red.hover");
+  }
```
AntoineAugusti, line 31: Delete and use utility classes on these buttons? Unsure if we want to have helper classes/embrace utilities only/have templates we include

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
...
+  .button-secondary, .button-secondary:visited, .button-secondary:link {
+    background-color: theme("colors.gray.button");
+    box-shadow: 0 2px 0 theme("colors.gray.border");
+    font-size: 1em;
+    @apply text-black;
+  }
+
+  .button-secondary:hover {
+    background-color: theme("colors.gray.hover");
+  }
```
AntoineAugusti, line 42: Same question than `.button-red`

tailwind.config.js
```diff
@@ -44,7 +46,10 @@ module.exports = {
       },
       gray: {
         default: "#EEE",
-        selected: "#e1e4e7"
+        selected: "#e1e4e7",
+        button: "#dee0e2",
+        hover: "#d0d3d6",
+        border: "#b5babe"
```
AntoineAugusti, line 52: Do you know if we can reduce the number of gray shades?

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
+@layer components {
+
+  /* Note: Buttons are <a> sometimes, which is why :link and : visted styles exist */
```
AntoineAugusti, line 3:
```suggestion
  /* Note: Buttons are <a> sometimes, which is why :link and : visited styles exist */
```

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
...
+    color: theme("colors.green.darker");
+    border-color: theme("colors.green.darker");
+    background-color: theme("colors.white.default");
+    @apply px-6 py-4 border-b-4 font-bold border-2;
+    border-bottom: 4px solid;
+    box-shadow: none;
+  }
+  #contact-form .button:hover {
+    @apply bg-gray-selected;
+  }
```
AntoineAugusti, line 55: If this button appears only once, can we put all these utilities in the HTML directly?

app/assets/stylesheets/tailwind/buttons.css
```diff
@@ -0,0 +1,61 @@
+@layer components {
```
AntoineAugusti, line 1: I think we've got a mix of `.button` and `.btn`, could we keep only 1 value?

AntoineAugusti (changes requested):
Thanks for you great work tracking down all these buttons in the codebase! A few suggestions and comments.

app/assets/stylesheets/tailwind/buttons.css
brdunfield: This might be a good point to involve design. I know a big concern was having to change a bunch of files to make a small style change. In this case, if we wanted to style a red button differently, we would only need to make changes here.

I know tailwind's core concept is ultility class first, but these helper classes are available for scalability. So.. it's not a clearly defined answer here. :(

tailwind.config.js
brdunfield: It looks like default is not in use. I could make "button" the default and cut this down by one.

app/assets/stylesheets/tailwind/buttons.css
brdunfield: Agreed - this is an artifact of the old and new conventions co-existing before this that I must have missed.

app/assets/stylesheets/tailwind/buttons.css
brdunfield: None of the tailwind padding classes correspond to exactly these values - we could consult with design and see if the closest ones are acceptable.

app/assets/javascripts/apiKey.js
brdunfield: I would need more context around this. I didn't modify the api-key classes here.

app/assets/stylesheets/tailwind/buttons.css
brdunfield: This button comes from the main page_footer template, which is used in various places, not just here.

tailwind.config.js
brdunfield: Removed "button" and made it "default"

app/assets/stylesheets/tailwind/buttons.css
AntoineAugusti: Awesome sounds like a good plan! Migrate CSS files and then update them after consulting with design.

app/assets/javascripts/apiKey.js
AntoineAugusti: It was a proposal to start to prefix classes only used by JavaScript with `js-` to make them pop in the HTML code. It conveys a warning: "attention, I'm not used for UI purposes but for feature purposes"

app/assets/stylesheets/tailwind/buttons.css
AntoineAugusti: My comment was about writing `@apply bg-green` instead of this line of code. They both do the same, agreed.

app/assets/stylesheets/tailwind/buttons.css
AntoineAugusti: Yeah, it's a team preference. If we maintain a low number of these classes, a minimum amount of CSS files, it seems a good idea.

app/templates/components/page-footer.html
```diff
@@ -15,7 +15,7 @@
       <button
         data-button-id={{button_id}}
         type="submit"
-        class="btn btn-green text-base outline-none focus:shadow-outline {% if destructive %}-destructive{% endif %}"
+        class="button text-base outline-none focus:shadow-outline {% if destructive %}-destructive{% endif %}"
```
AntoineAugusti, line 18: I think the CSS class `-destructive` doesn't exist anymore. You may want to remove this from the HTML or convert to utility classes

brdunfield: Good catch - i've updated it to at the "button-red" class when destructive

tailwind.config.js
brdunfield: Turns out the default grey actually was in use on the homepage. I'm going to revert this change.

I think any discussion around actually reducing the colour palette will be a design conversation and will take effect withe the new styles.

AntoineAugusti (approved):
Awesome! :shipit:

PR #787, opened 2020-10-09
Cache npm and pip on GitHub Actions

AntoineAugusti:
Fixes https://github.com/cds-snc/notification-api/issues/1129

The idea is to speed up builds on GitHub Actions by caching npm and pip dependencies. The cache will be invalidated when changing requirements/package-lock.

Speed gain of 1 minute on the last commit 🥳


brdunfield (approved):
👍

PR #789, opened 2020-10-13
Remove blocked attribute on non User models

AntoineAugusti:
Fixes https://github.com/cds-snc/notification-api/issues/1154

The `blocked` attribute was added to various models that don't support it: org, service, InvitedUser, InvitedOrgUser. The only model using this attribute is the User model.

This extra attribute was responsible for a bug I encountered while working on another feature. I fixed it at the same time because I already investigated the root cause.


brdunfield (approved):
🧹

PR #790, opened 2020-10-14
Upgrade Flask-Login

AntoineAugusti:
Reused code from https://github.com/alphagov/notifications-admin/commit/85f159a25fbeb502e1f42fadb2fe52e2734b3502

Here are their notes:
flask_login sets a bunch of variables in the session object. We only use
one of them, `user_id`. We set that to the user id from the database,
and refer to it all over the place.

However, in flask_login 0.5.0 they prefix this with an underscore to
prevent people accidentally overwriting it etc. So when a user logs in
we need to make sure that we set user_id manually so we can still use
it.

flask_login sets a bunch of variables on the `flask.session` object.
However, this session object isn't the one that gets passed in to the
request context by flask - that one can only be modified outside of
requests from within the session_transaction context manager (see [1]).
So, flask_login populates the normal session and then we need to copy
all of those values across.

We didn't need to do this previously because we already set the
`user_id` value on line 20 of tests/__init__.py, but now that
flask_login is looking for `_user_id` instead we need to do this
properly.

[1] https://flask.palletsprojects.com/en/1.1.x/testing/#accessing-and-modifying-sessions

Trello card https://trello.com/c/vacZWEnH/70-upgrade-flask-login-on-admin

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -132,6 +132,7 @@ class User(JSONModel, UserMixin):
 
     def login(self):
         login_user(self)
+        session['user_id'] = self.id
 
     def sign_in(self):
 
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -4,7 +4,7 @@
 timeago==1.0.14
 Flask==1.1.2
 Flask-WTF==0.14.3
-Flask-Login==0.4.1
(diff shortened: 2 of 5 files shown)


brdunfield (approved):
🦑

PR #791, opened 2020-10-14
Upgrade WTForms

AntoineAugusti:
Trello Card https://trello.com/c/uoeGGVUE/71-upgrade-wtforms-on-admin.

The biggest change is regarding error messages, a breaking change WTForms for radio fields. Changed to a single message: "You need to choose an option" instead of required/invalid choice.

See GDS https://github.com/alphagov/notifications-admin/commit/65bb72ef2fe750546e69f1d298569eb4c63f6226

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -22,7 +22,9 @@ from wtforms import (
     HiddenField,
     IntegerField,
     PasswordField,
-    RadioField,
+)
+from wtforms import RadioField as WTFormsRadioField
+from wtforms import (
     SelectField,
     SelectMultipleField,
     StringField,
@@ -104,6 +106,21 @@ def get_next_days_until(until):
         )
         for i in range(0, days + 1)
     ]
+
+
+class RadioField(WTFormsRadioField):
+    def __init__(
+        self,
+        *args,
+        **kwargs
+    ):
+        super().__init__(*args, **kwargs)
+        self.validate_choice = False
+
+    def pre_validate(self, form):
+        super().pre_validate(form)
+        if self.data not in dict(self.choices).keys():
+            raise ValidationError(_l('You need to choose an option'))
 
 
 class MultiCheckboxField(SelectMultipleField):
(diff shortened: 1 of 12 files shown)


brdunfield:
Is the content change (to "you need to select an option") visible on any public facing pages? (I've only been able to find it on platform admin pages). Regardless, we probably want to run it past @anikbrazeau if you haven't already!

AntoineAugusti:
It may be visible on a few pages, if you're a service owner. Not only for platform admins. It's only for radio buttons.

Previously, error messages were "This field is required" or "Not a valid choice" (users shouldn't see this without tampering the form). Know it'll always be "You need to choose an option".

I didn't choose this new wording, I reused this translation string from elsewhere. @anikbrazeau, what's your opinion on this?

app/main/forms.py
```diff
@@ -455,7 +472,6 @@ def __init__(self, all_template_folders=None, *args, **kwargs):
             ('sms_auth', _l('Text message code')),
             ('email_auth', _l('Email code')),
         ],
-        validators=[DataRequired()]
```
brdunfield, line 458: Do you know why these were removed? I couldn't find a reason in the UK PR.

AntoineAugusti: I think that the reason is that our radio buttons are all required. We don't have optional radio buttons. Deleting all these didn't break any tests, proving that it was superfluous

anikbrazeau:
@AntoineAugusti @brdunfield  Looks good, do we need French too? 

What if we went with:
**EN**: "You need to select an option."
**FR**: "Veuillez sélectionner une option."

AntoineAugusti:
It was an existing translation but changed it, it's more polite in French, you're right. Thanks!

brdunfield (approved):
I am not sure the best way to test manually, but I ran through a few of the forms I'm aware of and all seemed to work as expected. And tests are all passing.

LGTM! 🦑

PR #792, opened 2020-10-16
Fix cache for get_latest_stats

AntoineAugusti:
`get_latest_stats` is cached but the `lang` argument is not taken into account. The function is written to format numbers differently according to the lang parameter, but you can see it's not working in production by switching to "Français", these numbers will be formatted in English.

To take arguments into account, we should use the `memoize` function instead. https://flask-caching.readthedocs.io/en/latest/api.html#flask_caching.Cache.memoize

> Use this to cache the result of a function, taking its arguments into account in the cache key.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -55,7 +55,7 @@ with open('{}/email_domains.txt'.format(
 user_is_logged_in = login_required
 
 
-@cache.cached(timeout=300, key_prefix='latest_stats')
+@cache.memoize(timeout=300)
 def get_latest_stats(lang="en"):
     json_data = {}
     email_totals = 0


app/utils.py
```diff
@@ -55,7 +55,7 @@
 user_is_logged_in = login_required
 
 
-@cache.cached(timeout=300, key_prefix='latest_stats')
+@cache.memoize(timeout=300)
```
AntoineAugusti, line 58: This doesn't take a `key_prefix` argument, it uses the method name + arguments

brdunfield:
Can you point me to where I can see this in the app? I haven't had to look at this code before.

AntoineAugusti:
Sure! You can see this bug in prod.

1. Go to https://notification.alpha.canada.ca
1. See the number of messages sent with commas in English
1. Switch to French
1. See that the number doesn't change, because the method is cached

On the deploy app you can see that it's fixed

brdunfield (approved):
🦑 🚢

PR #793, opened 2020-10-16
Retry failed requests to the Notify API

AntoineAugusti:
Turns out that the Notify API client throws a 503 error when it can’t connect to the API. This means that the API never returned an error, the admin didn’t manage to connect to the API. [Code in charge of this](https://github.com/alphagov/notifications-python-client/blob/8ba2c878d50bae4a4e130614e9bcb06a2018c5da/notifications_python_client/errors.py#L62-L84) and test [proving that a connection error results are converted in a 503 error](https://github.com/alphagov/notifications-python-client/blob/master/tests/notifications_python_client/test_base_api_client.py#L44-L52).

They have a note in the client regarding 503 errors, they even raise a specific exception for these. They say
> Specific instance of HTTPError for 503 errors. Used for detecting whether failed requests should be retried.

But turns out the admin doesn’t retry these requests after such a failure. We don’t and the current GDS codebase doesn’t either. I don’t think it’s a good idea. API requests can temporarily fail (hey, distributed systems) so they are worth retrying if it’s a temporary failure (like a 503 error).

Therefore, this PR retries requests to the Notify API up to 3 times when encountering a 503 error.

Slack discussion: https://gcdigital.slack.com/archives/CNWA63606/p1602853930021500

### DIFF
--- a/app/notify_client/__init__.py
+++ b/app/notify_client/__init__.py
@@ -5,6 +5,7 @@ from flask import abort, has_request_context, request
 from flask_login import current_user
 from notifications_python_client import __version__
 from notifications_python_client.base import BaseAPIClient
+from notifications_python_client.errors import HTTP503Error
 
 logger = logging.getLogger(__name__)
 
@@ -87,6 +88,18 @@ class NotifyAdminAPIClient(BaseAPIClient):
             self.log_admin_call(args[0], "DELETE")
         self.check_inactive_service()
         return super().delete(*args, **kwargs)
+
+    def _perform_request(self, method, url, kwargs):
+        # Retry requests to the Notify API up to 3 times
+        # if they fail with a 503 status, thrown when
+        # the admin can't connect to the API.
+        for i in [1, 2, 3]:
+            try:
+                return super()._perform_request(method, url, kwargs)
+            except HTTP503Error as e:
+                logger.warn("Retrying API request after failure {} {}".format(method, url))
+                if i == 3:
+                    raise e
 
 
 class InviteTokenError(Exception):
--- a/tests/app/notify_client/test_base.py
+++ b/tests/app/notify_client/test_base.py
@@ -1,0 +1,45 @@
+import pytest
+import requests
+from notifications_python_client.errors import HTTP503Error
+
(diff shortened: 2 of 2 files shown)


app/notify_client/__init__.py
```diff
@@ -88,6 +89,18 @@ def delete(self, *args, **kwargs):
         self.check_inactive_service()
         return super().delete(*args, **kwargs)
 
+    def _perform_request(self, method, url, kwargs):
+        # Retry requests to the Notify API up to 3 times
+        # if they fail with a 503 status, thrown when
+        # the admin can't connect to the API.
+        for i in [1, 2, 3]:
```
brdunfield, line 96: As Pierre-Paul mentioned - it would be good to know what the retry frequency / timeout is here - how long are pods down for? Will this method actually work most of the time? Do we want to consider spacing out the requests in a specified way?

AntoineAugusti: The Notification client doesn't have a timeout set for requests, for now, it may wait forever before getting a response. https://github.com/alphagov/notifications-python-client/blob/8ba2c878d50bae4a4e130614e9bcb06a2018c5da/notifications_python_client/base.py#L92-L101

By retrying requests after a connection error, the idea is to get assigned another server by the load balancer, hopefully a healthy one.

If the pod is unhealthy/going to be restarted, it'll be removed from the load balancer at the same time we're retrying or just after. With the current number of instances running at the same time, I don't think we need to space out requests - getting the same instance 3 times in a row seems unlikely.

brdunfield (approved):
I think we might be able to come up with something better long term, but for now this is a big improvement.

Is there a good way to test this in staging before we push to prod? I think staging might only have 1 API pod.

AntoineAugusti:
The first test in staging is that it should not impact the admin: API requests/responses should work normally as before and we should not see logs saying that requests were retried. Ideally, we should provoke a retry but making connections to a pod impossible but it may be difficult to do.

In prod we should see a huge decrease in 500 errors on the admin and a few log messages sometimes when requests are retried when pods are restarted.

PR #794, opened 2020-10-16
Low hanging fruit CSS cleanup

brdunfield:
- Removed unused CSS files
- Fixed a bug that was introduced when header.scss was removed, where the header breadcrumbs on smaller screens were not displaying properly (the top menu with Product Features, tech documentation etc)

Part of https://trello.com/c/davC2pOu/66-translating-existing-styles-to-tailwind


app/templates/main_template.html
```diff
@@ -240,7 +240,7 @@
               />
             </a>
           </div>
-          <div id="proposition-links" class="hide lg:block">
+          <div id="proposition-links" class="lg:block xs:hidden sm:hidden md:hidden">
```
AntoineAugusti, line 243: Could it be just `hidden lg:block`?

https://tailwindcss.com/docs/responsive-design#targeting-mobile-screens

tailwind.config.js
```diff
@@ -26,6 +26,7 @@ module.exports = {
       body: ["Noto Sans"]
     },
     screens: {
+      xs: {'max': "639px"},
```
AntoineAugusti, line 29: Why is this needed?

app/templates/main_template.html
brdunfield: I tried that and it didn't work - because I haven't ported over the "hidden" style yet and it interferes. I can add a note / comment to refactor when that gets done?

tailwind.config.js
brdunfield: Because our other styles are min-width, so on a mobile screen the nav bar does not disappear.

AntoineAugusti (approved):
Thanks for cleaning all this code! 🗑️

app/templates/main_template.html
AntoineAugusti: Awesome 👍

tailwind.config.js
AntoineAugusti: Right, hopefully we'll be able to delete the `xs` setting later on

PR #796, opened 2020-10-21
Adding a service to an org cleans orgs cache

AntoineAugusti:
Adding a service to an organisation doesn't clean the organisations cache currently. This has the effect of having a count of live services wrong.

Noticed that yesterday when creating a new org, attaching a service and turning it live.
Screen Shot 2020-10-21 at 10 38 06

[Link to PHAC org with a live service](https://notification.alpha.canada.ca/organisations/6d5f3682-54ff-4aba-8405-96e71de45bb1)

This PR cleans the `organisations` cache when the count of the live services for an org can be affected:
- turning it live
- changing if the service should be counted as live or not
- attaching a service to an org


brdunfield (approved):
🦑

PR #798, opened 2020-10-21
Remove banner.scss and port to tailwind. Also upgrade tailwind to current version

brdunfield:
https://trello.com/c/davC2pOu/66-translating-existing-styles-to-tailwind

This one knocks out banner.scss 🔨 

I didn't bother porting over banner-tour, as it's connected to the onboarding tour that was removed months ago. I created a second ticket to clean out the rest of that code, as it's technically still there.


app/assets/javascripts/apiKey.js
```diff
@@ -8,11 +8,11 @@
     const states = {
       'keyVisible': (key, thing) => `
         <span class="api-key-key">${key}</span>
-        <input type='button' class='api-key-button-copy button-secondary' value='${window.polyglot.t('copy')} ${thing} ${window.polyglot.t('to_clipboard')}' />
+        <input type='button' class='api-key-button-copy button button-secondary' value='${window.polyglot.t('copy')} ${thing} ${window.polyglot.t('to_clipboard')}' />
```
brdunfield, line 11: These got missed in the button PR, they needed "button" added

tailwind.config.js
```diff
@@ -64,12 +68,26 @@ module.exports = {
...
-      'extra-tight': '0.9',
-    }
+    },
+    extend: {
+      backgroundSize: {
+        '19': '19px',
+      },
+      backgroundPosition: {
+        tickPos: "15px 15px",
+      },
```
AntoineAugusti, line 78: Only used for the tick I think? What do you think of doing it in the CSS file directly? Not sure these will be reused elsewhere.

tailwind.config.js
```diff
@@ -64,12 +68,26 @@ module.exports = {
...
+        'extra-tight': '0.9',
+      },
+      outline: {
+        yellow: '3px solid #ffbf47'
+      },
+      spacing: {
+        gutter: '30px',
+        gutterHalf: '15px',
+        gutterAndAHalf: '45px',
+      }
```
AntoineAugusti, line 89: Added to not change the existing design I assume? Can we delete these values later?

app/assets/javascripts/apiKey.js
AntoineAugusti: You could also sneak in this button in the PR or open another PR if you feel like it! https://staging.notification.cdssandbox.xyz/documentation

AntoineAugusti (approved):
Good work 👏  A suggestion to clean the config and if you want to fix another button in the same PR or open another PR

tailwind.config.js
brdunfield: Yes these are existing styles. If the new values design selects are ones that tailwind has by default, then yes these can be removed.

tailwind.config.js
brdunfield: I originally had it just in the CSS directly but was trying to get in the habit of doing things the tailwind way. But you're right, single use values don't belong here.

It looks like the background size is used in another instance where the tick is used, but background position is not. So I'll clean up background position and just do it in the CSS.

PR #799, opened 2020-10-22
Fix method to get the real IP address

AntoineAugusti:
`request.remote_addr` is not accurate when requests go through proxies as they do in our production environment. This commit introduces a util method to get the real IP address, no matter if it went through a proxy or not. It uses the [`access_route` method](https://werkzeug.palletsprojects.com/en/1.0.x/wrappers/#werkzeug.wrappers.BaseRequest.access_route) to do the heavy work. The original IP address is at the beginning of the list of the `X-Forwarded-For` header if it is set ([see on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-For)).

Trello card https://trello.com/c/sg03A4Mt/150-repair-ip-geolocation-service

### DIFF
--- a/app/event_handlers.py
+++ b/app/event_handlers.py
@@ -1,6 +1,7 @@
 from flask import request
 
 from app import events_api_client
+from app.utils import get_remote_addr
 
 
 def on_user_logged_in(_sender, user):
@@ -40,16 +41,8 @@ def _send_event(event_type, **kwargs):
 
 
 def _construct_event_data(request):
-    return {'ip_address': _get_remote_addr(request),
+    return {'ip_address': get_remote_addr(request),
             'browser_fingerprint': _get_browser_fingerprint(request)}
-
-
-# This might not be totally correct depending on proxy setup
-def _get_remote_addr(request):
-    if request.headers.getlist("X-Forwarded-For"):
-        return request.headers.getlist("X-Forwarded-For")[0]
-    else:
-        return request.remote_addr
 
 
 def _get_browser_fingerprint(request):
--- a/app/utils.py
+++ b/app/utils.py
@@ -260,6 +260,8 @@ def id_safe(string):
 def get_remote_addr(request):
     try:
         return request.access_route[0]
+    # This except block is here to prevent to fail when the env `REMOTE_ADDR`
+    # is not set. This only happens when running tests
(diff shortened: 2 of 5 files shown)


app/event_handlers.py
```diff
@@ -40,18 +41,10 @@ def _send_event(event_type, **kwargs):
 
 
 def _construct_event_data(request):
-    return {'ip_address': _get_remote_addr(request),
+    return {'ip_address': get_remote_addr(request),
             'browser_fingerprint': _get_browser_fingerprint(request)}
```
AntoineAugusti, line 45: The previous method was partially correct. It uses the right HTTP header but doesn't parse properly the list. You can look at recent rows of the `events` table in prod to see JSON payload like this.

```json
{
    "ip_address": "205.193.22.34, 192.168.16.121",
    "browser_fingerprint": {
        "browser": "msie",
        "platform": "windows",
        "version": "11.0",
        "user_agent_string": "Mozilla/5.0 (Windows NT 10.0; WOW64; Trident/7.0; Touch; rv:11.0) like Gecko"
    },
    "user_id": "7daba5db-57de-4fc2-a39a-3cec0a1006d1"
}
```

app/utils.py
```diff
@@ -257,6 +257,13 @@ def id_safe(string):
     return email_safe(string, whitespace='-')
 
 
+def get_remote_addr(request):
+    try:
+        return request.access_route[0]
+    except IndexError:
+        return None
```
AntoineAugusti, line 264: Used to catch the `REMOTE_ADDR` header not set on the request when tests are executed, will not happen in production

brdunfield:
Is this something we will want on direct API requests as well?

AntoineAugusti:
Good question! After a quick search, I didn't find any occurrences of `request.remote_addr` or `request.access_route` in the API repo. It seems like the admin is responsible for passing it to the API to perform actions

app/utils.py
brdunfield: this would be nice to have as an in-code comment so this can be more legible down the line :)

brdunfield (approved):
🦑 LGTM! Consider adding in-code comments as I mentioned above, but actual code itself looks good!

PR #804, opened 2020-10-26
Css components

brdunfield:
Just doing 2 files here so this doesn't grow too big
- remove api-key.scss (affects styles related to displaying api-keys and the associated buttons)
- remove big-number.scss (affects.. "big numbers" .. i.e. the larger numbers displayed on dashboards and so on.

I simplified the styles / markup around big-numbers a bit, as styles were previously dependent on html hierarchies. Instead, I just passed a variable into the jinja partial and added different tailwind classes based on the variable and other existing ones. It can probably be simplified further - if you see any obvious ways I'm happy to incorporate them.


app/templates/views/api/index.html
```diff
@@ -1,6 +1,5 @@
 {% extends "withnav_template.html" %}
 {% from "components/table.html" import list_table, field, hidden_field_heading %}
-{% from "components/api-key.html" import api_key %}
```
brdunfield, line 3: Cleaned out old references that were not being used

tailwind.config.js
```diff
@@ -47,7 +44,9 @@ module.exports = {
       blue: {
         lighter: "#B2E3FF",
         default: "#26374A",
-        selected: "#75b9e0"
+        selected: "#75b9e0",
+        govukblue: "#005ea5",
```
brdunfield, line 48: These were copied from the palettes scss file. I think our colour palette needs an audit and will likely be addressed with kate's work, but for now I will just copy the values and their old names.

app/templates/views/dashboard/_inbox.html
```diff
@@ -2,7 +2,7 @@
 
 <div class="ajax-block">
   {% if inbound_sms_summary != None %}
-    <div id="total-received" class="big-number-meta-wrapper">
+    <div id="total-received" class="big-number-meta-wrapper relative bg-blue-govukblue mx-0 mt-gutterHalf mb-gutter">
```
brdunfield, line 5: big-number-meta-wrapper is not doing any styling here - it's just to keep tests passing. We can potentially come up with another prefix for this, similar to the js prefix.

tailwind.config.js
```diff
@@ -63,16 +62,32 @@ module.exports = {
...
+      },
+      fontFamily: {
+        sans: ["lato"],
+        body: ["Noto Sans"]
+      },
+      inset: {
+        '2': '2px',
+        '5': '5px',
+        '7': '7px',
+      },
```
AntoineAugusti, line 90: What do you think of not putting these values in the config and using an inline style?

tailwind.config.js
```diff
@@ -63,16 +62,32 @@ module.exports = {
...
+        default: "transparent"
       }
     },
     extend: {
       backgroundSize: {
         '19': '19px',
       },
+      borderWidth: {
+        '1':'1px',
+      },
```
AntoineAugusti, line 81: Used only once for now I think? Maybe put the style directly in the component and then decide if we can use `border-2` from Tailwind?

tailwind.config.js
```diff
@@ -63,16 +62,32 @@ module.exports = {
...
         hover: "#00692f",
-        border: "#003618"
+        border: "#003618",
+        green: "#006435",
       },
       black: {
         default: "#000"
+      },
+      transparent: {
+        default: "transparent"
```
AntoineAugusti, line 72: Is this needed? `.bg-transparent` exists in Tailwind

postcss.config.js
```diff
@@ -15,6 +15,7 @@ const purgecss = require("@fullhuman/postcss-purgecss")({
     "banner-dangerous",
     "banner-with-tick",
     "banner-list-bullet",
+    "bottom-2"
```
AntoineAugusti, line 18: If it becomes too difficult to maintain this list, I think you add JS files in the PostCSS config as well

app/templates/components/big-number.html
```diff
@@ -1,9 +1,9 @@
-{% macro big_number(number, label, link=None, currency='', smaller=False, smallest=False) %}
+{% macro big_number(number, label, link=None, status=None, currency='', smaller=False, smallest=False) %}
   {% if link %}
     <a class="big-number-link" href="{{ link }}">
   {% endif %}
-  <div class="big-number{% if smaller %}-smaller{% endif %}{% if smallest %}-smallest{% endif %}">
-    <div class="big-number-number">
+  <div class="big-number{% if status %} big-number-dark{% endif %}{% if status and link %} bg-transparent{% endif %}">
+    <div class="big-number-number{% if smaller %} text-lg{% endif %}{% if smallest %} text-title{% endif %}">
```
AntoineAugusti, line 6: Curious about why you use utilities for `smaller` VS class names for `status`. Why not go all utilities as we have an HTML component here?

app/assets/javascripts/apiKey.js
```diff
@@ -39,14 +39,14 @@
...
-          'click', '.api-key-button-copy', () =>
+          'click', '.js-api-key-button-copy', () =>
             this.copyKey(
               $('.api-key-key', component)[0], () =>
                 $component.html(states.keyCopied(thing))
             )
         )
         .on(
-          'click', '.api-key-button-show', () =>
+          'click', '.js-api-key-button-show', () =>
```
AntoineAugusti, line 49: Love the use of the `.js-` prefix 👍

tailwind.config.js
brdunfield: It does, but the colour section is not extending tailwind, it is overwriting it.

postcss.config.js
brdunfield: I'll look into that and keep it in mind

app/templates/components/big-number.html
brdunfield: Mostly because a) there are multiple utilities in big-number-dark, and that class is used elsewhere.

tailwind.config.js
brdunfield: I don't see these values used elsewhere from a quick search, so this would make sense

tailwind.config.js
brdunfield: we do have several other border 1px widths elsewhere in the app that haven't been ported yet, so I might want to leave this one.

AntoineAugusti (approved):
Thanks for your work! 👏 Okay to ship!

Decide if you want to keep some custom config for `inset` or use styles instead 👍

PR #810, opened 2020-11-02
Activity dashboard

AntoineAugusti:
This is the new activity dashboard.

There is an API PR to review/merge first to add the new endpoint to get stats by month. https://github.com/cds-snc/notification-api/pull/1165

Previews
localhost_6012_stats
localhost_6012_stats(iPhone 6_7_8)

Currently chatting with @ioanacontu and @anikbrazeau to put the final touches to the design and content but this will not change much


app/templates/components/table.html
```diff
@@ -8,7 +8,7 @@
     <thead class="table-field-headings{% if field_headings_visible %}-visible{% endif %}">
       <tr>
         {% for field_heading in field_headings %}
-          <th scope="col" class="table-field-heading{% if loop.first %}-first{% endif %}" width="{% if equal_length %}{{ (100 / field_headings|length)|int }}%{% endif %}">
+          <th scope="col" class="table-field-heading{% if loop.first and not equal_length %}-first{% endif %}" width="{% if equal_length %}{{ (100 / field_headings|length)|int }}%{% endif %}">
```
AntoineAugusti, line 11: Another table was called with the `equal_length` set to `True` and it was not displaying table headers so nothing else changed in the UI

app/utils.py
```diff
@@ -632,6 +639,30 @@ def report_security_finding(
...
+        12: _l("December"),
+    }
+
+    return translatedMonth.get(int(monthNumber), _l("Invalid month"))
+
+
+def yyyy_mm_to_datetime(string):
+    return datetime(int(string[0:4]), int(string[5:7]), 1)
+
+
```
AntoineAugusti, line 665: Moved to utils.py from the dashboard.py view

brdunfield:
Are we sure we want to remove the numbers from the home page? Aren't they a good marketing tool? I'm trying to dig through our old portage designs / notes to see if this was something we had planned / agreed on but am not having much luck.

AntoineAugusti:
Sounds like the plan is to remove numbers from the homepage. @ioanacontu wants to add a link on the homepage to this page. Something more visible than the footer link.

brdunfield:
Given the unanswered design questions here, I think we can park this until after Anik returns next week.

ioanacontu:
Sorry misunderstanding, we're not removing the numbers on the homepage. Only need to find a way to link them to the dashboard (for more detailed numbers).

AntoineAugusti:
@ioanacontu @anikbrazeau Do you know how we're going to link the activity dashboard from the numbers on the homepage?

anikbrazeau:
Thanks @AntoineAugusti! Let's keep the change simple by just [adding a link on the homepage](https://www.figma.com/file/z0V0cw5uDbOqKLu1ugGE3p/Activity-dashboard-(Portage)-%7C-Tableau-de-bord-de-l'activit%C3%A9-(Portage)?node-id=478%3A2) to the existing `GC Notify by the numbers` section:  

Screen Shot 2020-11-17 at 9 09 51 AM

Screen Shot 2020-11-17 at 9 09 41 AM

While we're at it, can we also fix the text size since the text in prod is very small and making it a link as well might not pass AAA? The heading/title for "GC Notify by the numbers is also off in size compared to other sections of the homepage. 

Here's what @ioanacontu's proposing for new sizing:
Screen Shot 2020-11-17 at 09 35 50

ioanacontu:
Sorry for late reaction. Do you think we could break the text in French down onto two lines? @anikbrazeau @AntoineAugusti 
Screen Shot 2020-11-17 at 11 14 20

brdunfield:
To have the "Learn more in the Activity Dashboard" link visually match the rest of the links on the page, you'll need to add `class="text-blue link:text-blue visited:text-blue"` to the anchor

AntoineAugusti:
@brdunfield Thanks, just changed it 🙌

brdunfield (approved):
LGTM! 🦑

PR #811, opened 2020-11-02
Remove cds_template_jinja dependency

brdunfield:
Most of this dependency was already being overwritten. None of the HTML or JS was being used. Only a few CSS styles were not being overwritten, and I've ported them over here.


AntoineAugusti (changes requested):
On the deployment preview, all links on the homepage have the blue colour

brdunfield:
Ah this was due to me porting over the basic link styling from the main app - which was not present on the home page before this PR. Should be fixed now!

AntoineAugusti (approved):
🚢 :shipit:

PR #812, opened 2020-11-03
Upgrade Python dependencies

AntoineAugusti:


.github/workflows/test.yaml
```diff
@@ -14,7 +14,7 @@ jobs:
     - uses: actions/cache@v2
       with:
         path: ~/.cache/pip
-        key: ${{ runner.os }}-pip-${{ hashFiles('requirements_for_test.txt') }}
+        key: ${{ runner.os }}-pip-${{ hashFiles('requirements.txt') }}
```
brdunfield, line 17: Why this change?

AntoineAugusti: Right, I should have left a note. The file is used to know when to bust the cache in CI. The `requirements.txt` is the one to use as this is the file that's going to change when updating 3rd party or `requirements-app.txt`. `requirements_for_test.txt` is very stable in comparison

brdunfield (approved):
🦑

PR #814, opened 2020-11-04
Remove inbound message / conversation views

brdunfield:
Given we are not currently supporting inbound messages, I decided to rip this out.

Inbound SMS options / platform admin views / logic are all still preserved, as we need to leverage this feature to set long codes using PinPoint for outgoing messages.

Conversation / inbound message related views are removed in this PR.

🚨 I do however recommend we assign an inbound number to a service in staging (if we don't have any already) to 100% verify this all behaves as expected before we merge to prod.


app/assets/stylesheets/tailwind/sms-message.css
```diff
@@ -0,0 +1,3 @@
+@layer components {
+  
+}
```
brdunfield, line 3: I guess this is here because I thought this was going to be a CSS PR at the start. I can remove if desired.

AntoineAugusti (approved):

PR #819, opened 2020-11-10
Report: send method stats by service

AntoineAugusti:
Add a new report to the admin!

Needs the API PR first https://github.com/cds-snc/notification-api/pull/1174

Trello card: https://trello.com/c/dFATfrV6/114-new-report-api-key-send-vs-ui-send


app/templates/views/platform-admin/reports.html
```diff
@@ -14,6 +14,10 @@ <h1 class="heading-large">
   <a target="_blank" href="{{ url_for('main.live_services_csv') }}">{{ _('Download live services csv report') }}</a>
 </p>
 
+<p>
+  <a href="{{ url_for('main.send_method_stats_by_service') }}">{{ _('Send method stats by service') }}</a>
```
brdunfield, line 18: I think this can be worded / phrased better, this is not very clear to me what this report contains.
Maybe something like:
"Notifications sent by method (api key and UI) by service."
That still sounds clunky. Maybe @anikbrazeau could help here? I know this is not user facing, but I still think we should try to be clear.

app/templates/views/platform-admin/send_method_stats_by_service.html
```diff
@@ -0,0 +1,20 @@
...
+{% from "components/form.html" import form_wrapper %}
+
+{% block per_page_title %}
+  Send methods report for live services
+{% endblock %}
+
+{% block platform_admin_content %}
+
+  <h1 class="heading-large">
+    Send methods report for live services
```
brdunfield, line 11: Again, I think this can be phrased more clearly. See above.

app/templates/views/platform-admin/reports.html
anikbrazeau: Is it focused on the services or the notifications, or both? Are one of the following more precise descriptions?

`List of services by sending method` // `Liste de service par mode d'envoi` ?

or

`Sending method: Services sending via API vs. User Interface` // `Mode d'envoi : Services utilisant l'API ou l'interface utilisateur` ?

or 

`Comparing services by sending method` // `Comparaison des services par mode d'envoi` ?

app/templates/views/platform-admin/send_method_stats_by_service.html
anikbrazeau: Is this essentially?

`How live services are sending notifications: API vs. User interface` // `Comment les services actifs envoient des notifications : API vs. interface utilisateur`

or if we want something shorter,

`List of live services by sending method` // `Liste de services actifs par mode d'envoi`

AntoineAugusti:
Thanks @anikbrazeau for your suggestions, changed to `Sending method: services sending via API vs. User Interface`

brdunfield (approved):
LGTM! Thank you Anik for taking a look :)

🦑

PR #827, opened 2020-11-23
Fix a few CSS bugs

brdunfield:
Thank you @AntoineAugusti for finding these before they hit prod

- personalisation boxes are in the right place
- templates search bar is back


AntoineAugusti (approved):
Perfect, thanks for the quick fix!

AntoineAugusti:
Merging this myself as you're off today

PR #829, opened 2020-11-24
Disable IP geolookup locally

AntoineAugusti:
:warning: **Do not merge before next Monday**

The IP geolookup service is going away from Google Cloud and the URL is not going to be accessible locally https://github.com/cds-snc/notification-manifests/pull/14

Therefore, disable this feature locally for developers.


brdunfield (approved):
LGTM 🦑

PR #830, opened 2020-11-24
Disable honeypot field on contact page

AntoineAugusti:
The autocomplete of browsers may fill the honeypot `phone` field on the contact form, not on purpose. This has the effect of displaying the "thanks" page without actually sending the contact request. This is unfortunate.

First I thought about setting `autocomplete=off` on the honeypot field but [many browsers ignore this](https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off).

Therefore, I've removed it for now and we'll revisit this if we get some spam.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -862,7 +862,6 @@ class ContactNotifyTeam(StripWhitespaceForm):
         ],
     )
     email_address = email_address(label=_l('Your email'), gov_user=False)
-    phone = StringField(_l('Your phone'))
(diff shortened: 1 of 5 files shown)


AntoineAugusti:
Tested on the PR review app successfully

brdunfield:
I'm going to hold off on reviewing this until we decide 100% that this is the route we want to take.

brdunfield (approved):
LGTM! 🦑 

We will be mitigating this with new WAF rules in our production instance that we will be migrating to next week. New rules will be in place next sprint sometime, so this will not be open for very long.

PR #831, opened 2020-11-25
Add security.txt for Vulnerability Disclosure Practices

RenDavidDimen:
#### Overview
Adding security.txt route for Vulnerability Disclosure Practices. Information on page generated by filling out the security form on [securitytxt.org](https://securitytxt.org/)

#### Addresses
https://trello.com/c/F7wzxTz3/110-create-a-securitytxt-file-for-vulnerability-disclosure-practices
https://github.com/cds-snc/notification-api/issues/1148


app/main/views/index.py
```diff
@@ -80,6 +80,18 @@ def robots():
     ), 200, {'Content-Type': 'text/plain'}
 
 
+@main.route('/.well-known/security.txt')
+def security_txt():
+    return (
+        'Contact: mailto:security@cds-snc.ca\n'
+        'Contact: mailto:security@cds-snc.ca\n'
```
AntoineAugusti, line 87:
```suggestion
        'Contact: mailto:securite@cds-snc.ca\n'
```
The email address is different in French

app/main/views/index.py
```diff
@@ -80,6 +80,18 @@ def robots():
...
+        'Preferred-Languages: en, fr\n'
+        'Policy: https://github.com/cds-snc/.github/blob/master/SECURITY.md\n'
```
AntoineAugusti, line 89: Could you link to the security page instead?

app/main/views/index.py
```diff
@@ -80,6 +80,18 @@ def robots():
     ), 200, {'Content-Type': 'text/plain'}
 
 
+@main.route('/.well-known/security.txt')
+def security_txt():
```
AntoineAugusti, line 84: Could you have a list and then `'\n'.join(list)` instead?

tests/app/main/views/test_index.py
```diff
@@ -80,6 +80,21 @@ def test_robots(client):
     )
 
 
+def test_security_txt(client):
+    assert url_for('main.security_txt') == '/.well-known/security.txt'
```
AntoineAugusti, line 84: That's a good thing to verify!

AntoineAugusti (changes requested):
Good work, thanks!

A few suggestions

app/main/views/index.py
RenDavidDimen: Good catch, I'll fix that now 👍

RenDavidDimen: Fix should be in place now

app/main/views/index.py
RenDavidDimen: That's a good idea, I'll switch this out for the security page link

app/main/views/index.py
RenDavidDimen: Yes, I don't see why not. It might be better to do it that way as well should this information need some updating. (It could also help in avoiding some minor syntax errors)

app/main/views/index.py
RenDavidDimen: I've switched the URL. For future reference, once notify is out of alpha, I believe this reference will need to be updated.

tests/app/main/views/test_index.py
```diff
@@ -72,11 +72,32 @@ def test_robots(client):
...
+    security_info = [
+        'Contact: mailto:security@cds-snc.ca',
+        'Contact: mailto:securite@cds-snc.ca',
+        'Preferred-Languages: en, fr',
+        'Policy: https://notification.alpha.canada.ca/security',
+        'Hiring: https://digital.canada.ca/join-our-team/',
+        'Hiring: https://numerique.canada.ca/rejoindre-notre-equipe/'
+    ]
+    assert response.get_data(as_text=True) == (
+        '\n'.join(security_info)
```
RenDavidDimen, line 100: I'm curious if anyone has any qualms with how this test is currently set up. My view is that it is prepping the data the same way it is being done in `index.py`. Should we write tests asserting the result to be the literal expected value or is setting them up in the same way fine?

AntoineAugusti: I'm not sure you need to make sure that the exact content is the same. As you say, you're recreating the logic. You can test maybe a specific important line, maybe `Contact: mailto:security@cds-snc.ca`

RenDavidDimen: That makes sense. I think that since we're following `security.org`'s file standard, I may leave them in for these tests to ensure that the format is the same.

app/main/views/index.py
```diff
@@ -80,6 +80,18 @@ def robots():
...
+    return (
+        'Contact: mailto:security@cds-snc.ca\n'
```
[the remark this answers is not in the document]
brdunfield: Do we actually have a french security email as well? I don't see it anywhere obvious.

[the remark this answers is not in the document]
brdunfield: Oh it's in this PR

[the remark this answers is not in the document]
RenDavidDimen: Created a config variable to store the `security-securite@cds-snc.ca` email

app/main/views/index.py
```diff
@@ -72,11 +72,28 @@ def index():
...
-        'Disallow: /contact\n'
-        'Disallow: /register\n'
+        '\n'.join(robot_rules)
+    ), 200, {'Content-Type': 'text/plain'}
+
+
+@main.route('/.well-known/security.txt')
+def security_txt():
+    security_info = [
+        'Contact: mailto:' + current_app.config["SECURITY_EMAIL"],
```
AntoineAugusti, line 89:
```suggestion
        f'Contact: mailto:{current_app.config["SECURITY_EMAIL"]}',
```

If you like f-strings, you can use this syntax. Really not important, just a suggestion

AntoineAugusti (changes requested):
Nice, thanks for adding a config!

Could you use the new config in the `/security` page as well?

app/main/views/index.py
RenDavidDimen: Read a little more into using f-strings and it seems the consensus is that it tends to be more read-able. As someone who likes to understand what they read, I'll be switching to using f-strings here!

app/main/views/index.py
```diff
@@ -72,11 +72,28 @@ def index():
...
+        '\n'.join(robot_rules)
+    ), 200, {'Content-Type': 'text/plain'}
+
+
+@main.route('/.well-known/security.txt')
+def security_txt():
+    security_info = [
+        f'Contact: mailto:{current_app.config["SECURITY_EMAIL"]}',
+        'Preferred-Languages: en, fr',
+        'Policy: https://notification.alpha.canada.ca/security',
```
AntoineAugusti, line 91: You should `url_for('main.security', _external=True)` here to avoid having the alpha domain hardcoded here

AntoineAugusti (changes requested):

AntoineAugusti (approved):
Looks perfect to me! Good to go 🚢

PR #832, opened 2020-11-25
Select default logo branding upon service creation

jimleroyer:
## Description

Adds a radio button to pick bilingual default branding between English first logo or French first logo. Based on the navigation workflow and [mocks on figma](https://www.figma.com/file/rOWYP6enG7YDY77T6BX6S1/Email-branding-%7C-Image-de-marque-des-courriels?node-id=0%3A1).

## Trello Card

https://trello.com/c/ALRncOSm/120-pair-and-push-on-bilingual-branding-when-setting-up-a-new-service

## Changes

* Added new screen to pick default bilingual logo after the service name has been selected.

## Tests

I made sure of code coverage by the existing and new tests as I traced execution by debugging into the different code paths.

* Repaired failing tests due to the new wizard nature of the *add service* form.
* Added tests around steps selection.
* Added tests around the branding screen.

### Things to watch for..

State management in the form is tricky with the different steps of the wizard. Losing information about the current state should not lead to an internal server error but at worst, lead back to the first step with clean step. This is what tripped me up the most while working on this feature. 

## Screenshots

Name-your-service-in-both-official-languages-–-Notify--00

Choose-a-logo-for-your-service-–-Notify


brdunfield (approved):
LGTM, and seems to work well! 🦑 

The only thing I'm a little unsure about is having all the steps in a single HTML / route, which is not quite the same as other "step" forms we have in the app. However it looks like you've written comprehensive tests around this, and as I said it does work well.

jimleroyer:
I pushed again to the PR to have two partials HTML to the add service screen -- one per step (2 total).

brdunfield:
Much cleaner! Still works great :)

PR #834, opened 2020-11-27
Update Live config to Production

AntoineAugusti:
The sole goal of this change is to be consistent with the API and to clean StatsD prefixes https://github.com/cds-snc/notification-api/pull/1183


brdunfield (approved):
🦑

PR #835, opened 2020-11-30
feat: remove AWS credentials for S3 bulk send

AntoineAugusti:
The S3 bucket used for bulk sending is now on the same AWS account so we don't need to use other credentials.

There is [a Kubernetes PR](https://github.com/cds-snc/notification-manifests/pull/28) to deploy to remove these variables as well


brdunfield (approved):
Tested locally and works. Please make sure to update staging / prod env files with the new bucket name(s). I had to go digging to find the new value for staging so I could properly test this.

Did any of the test env vars get changed? "Bethany Test Service" on staging used to be bulk send enabled, but doesn't appear to be anymore.

In any case:
🧹 🦑

AntoineAugusti:
Updated env variables in LastPass.

In staging the bulk send service is currently set to [this service](https://staging.notification.cdssandbox.xyz/services/ea608120-148a-4eba-a64c-4d9a8010e7b0) through the env `BULK_SEND_TEST_SERVICE_ID`.

https://github.com/cds-snc/notification-admin/blob/d288da27e92ecaf8ea8868b8399c2e1372caf8af/app/main/views/send.py#L68-L74

We could adjust it!

PR #837, opened 2020-12-01
Minor French content fixes

anikbrazeau:
Missing French string in main template sending flow: "How to format your message"

Change "téléphone cellulaire" to "téléphone mobile" in French


AntoineAugusti (approved):
Templates already had this string
https://github.com/cds-snc/notification-admin/blob/d288da27e92ecaf8ea8868b8399c2e1372caf8af/app/templates/views/edit-sms-template.html#L22
https://github.com/cds-snc/notification-admin/blob/d288da27e92ecaf8ea8868b8399c2e1372caf8af/app/templates/views/edit-email-template.html#L16

And I've confirmed it was not in the `en.csv` file. Good to go

PR #839, opened 2020-12-01
Porting CSS '/views' files

brdunfield:
Ported every .scss in the /views folder.
Notes:
- product-page.scss was unused so I just deleted it
- template.scss contained a bunch of letter styles, and I did not port these.
- send.scss was a single style so rather than add an import I just added this to index.css

7 more files down! 🧹


app/assets/stylesheets/tailwind/style.css
```diff
@@ -13,6 +14,15 @@
...
 @import "./textbox.css";
 
+/* views */
+@import "./views/dashboard.css";
+@import "./views/users.css";
+@import "./views/api.css";
+@import "./views/template.css";
+@import "./views/notification.css";
+
+/*! purgecss start ignore */
```
AntoineAugusti, line 24: Curious why you added this here as most of the CSS code was already there without this instruction

app/assets/stylesheets/tailwind/views/dashboard.css
```diff
@@ -0,0 +1,55 @@
...
+  .dashboard table td {
+    @apply text-smaller leading-tight font-normal border-0;
+  }
+  .dashboard > .heading-medium:first-of-type {
+    @apply mt-gutterHalf;
+  }
+
+  .keyline-block {
+    @apply pt-gutterHalf border-solid border-1 border-gray-grey2;
+  }
```
AntoineAugusti, line 15: Screen Shot 2020-12-02 at 09 25 11
I think this should only be a single border top

AntoineAugusti (approved):
Found a minor thing with a border, apart from that it seems fine!

Wondering why you need to add `/*! purgecss start ignore */` everywhere on this PR

app/assets/stylesheets/tailwind/style.css
brdunfield: I'd like to move away from having the long list of exemptions in the postcss config. So many of the styles don't actually exist in the base html and are constructed by jinja / JS. I guess my thought process here is that if we include this in every file, we can be lazier about having to upkeep the safelist. There will always need to be a few entries in the safelist for classes coming from utils, and a few that are dropped in the html files directly but are conditional (jinja).

I haven't actually gone through and cleaned out the safelist yet, which might be why there's a conflict here..

PR #841, opened 2020-12-02
Bump a few dependencies to fix pip issue

brdunfield:
Keeping as a draft until the utils PR (https://github.com/cds-snc/notification-utils/pull/70) goes in, as this tag doesn't exist yet! But this config works locally.


brdunfield:
This PR accidentally downgraded bleach which is throwing the Snyk errors. However, when I re-upgrade bleach on my machine it throws a new set of dependency resolution conflicts..

PR #843, opened 2020-12-04
Css app components

brdunfield:
There's only a few more "components" as listed in the original CSS to move, but I wanted to stop here as there's already 34 files in this PR.
After the "components" there will still be some base gov uk styles and dependency styles to clean up, but I think we're getting to the home stretch here!

This PR ports:
- placeholder.scss (many of the styles here are box shadows and styles that don't easily port to tailwind, so you'll see a lot of regular css here)
- navigation.scss
- file-upload.scss
- pill.scss (difficult to tell if this is currently being used - most references are on pages we may not be using, but there's one reference that looks to still be live so I ported it)
- browse-list.scss
- show-more.scss

I've also moved all the "components" files into a folder to clean things up a little bit, which is the main source of file changes in this PR.


jimleroyer (approved):
This looks good, although I don't have much clue of what's required honestly. A lot of files move around! Do we completely move away from Sass? or we use something else? or just plain CSS?

I will approve the request because @brdunfield is blocked without it and I see nothing wrong with the PR.

app/assets/stylesheets/tailwind/style.css
```diff
@@ -1,18 +1,24 @@
...
+@import "./components/browse-list.css";
+@import "./components/email_sms_message.css";
+@import "./components/buttons.css";
+@import "./components/checkboxes-radios.css";
+@import "./components/pill.css";
+@import "./components/show-more.css";
+@import "./components/api-key.css";
+@import "./components/list-entry.css";
+@import "./components/message.css";
+@import "./components/fullscreen-table.css";
```
AntoineAugusti, line 21: Wondering if you want to put `/*! purgecss start ignore */` before importing every file instead of adding it on every component

AntoineAugusti:
@jimleroyer The goal is to move away from class-based CSS to Tailwind and utilities as much as possible. For now, this is mostly moving code around, later it'll be about having a library of components built using Tailwind.

If you want to see pages built with Tailwind, you can look at the homepage and the activity dashboard

app/assets/stylesheets/tailwind/style.css
brdunfield: I wasn't sure if it worked like that. If it does then yes 100% I want to do that. I'll just double check before I make that change.

brdunfield: It looks like it DOES work! I will add this to my next PR (which is almost ready to go just needs some merge work with this one)

PR #844, opened 2020-12-07
Add Functionality to Send New Registered User Data to Freshdesk

RenDavidDimen:
Upon a successful registration, a new user will have their email, name, and date/time of sign up to Freshdesk

Requires [changes made for the same ticket](https://github.com/cds-snc/notification-api/pull/1188) in the notification-api repo.


tests/app/main/views/organisations/test_organisation_invites.py
```diff
@@ -364,13 +365,15 @@ def test_verified_org_user_redirects_to_dashboard(
...
     mock_login,
+    mocker,
 ):
     invited_org_user = InvitedOrgUser(sample_org_invite).serialize()
     with client.session_transaction() as session:
         session['expiry_date'] = str(datetime.utcnow() + timedelta(hours=1))
         session['user_details'] = {"email": invited_org_user['email_address'], "id": invited_org_user['id']}
         session['organisation_id'] = invited_org_user['organisation']
 
+    mocker.patch('app.user_api_client.send_new_registration_data')
```
RenDavidDimen, line 376: I'm mocking the API call in these tests as they appear to be hitting the `"main.verify"` route. I had originally attempted to put this code in the `conftest.py` file and reference the mock accordingly but that appears to give me 403 response statuses. I've attempted to explicitly import the mock definition to have the same 403 errors occur.

I'm currently looking at the parameters passed to the mocking functions in `conftest.py` as a potential culprit to the 403 errors

RenDavidDimen: I couldn't seem to find the reason why centralizing the test in `conftest.py` caused some tests to throw. If anyone might have a better solution, I'm open to suggestions 🙂

app/notify_client/user_api_client.py
```diff
@@ -120,6 +121,13 @@ def send_already_registered_email(self, user_id, to):
         endpoint = '/user/{0}/email-already-registered'.format(user_id)
         self.post(endpoint, data=data)
 
+    def send_new_registration_data(self, name, email):
+        now = datetime.datetime.now()
+        date = now.strftime('%Y-%m-%d %H:%M:%S')
+        data = {'id': self.notify_user_id, 'name': name, 'email': email, 'date': date}
```
AntoineAugusti, line 127: You need to build a `message` here or on the API side

AntoineAugusti (changes requested):

AntoineAugusti:
Closing this stale PR, we reworked our Freshdesk integration in https://github.com/cds-snc/notification-api/pull/1222

PR #845, opened 2020-12-08
Push Docker images to AWS ECR

AntoineAugusti:
This PR makes sure that we push our Docker containers on AWS Elastic Container Registry (ECR) public registry instead of Google Cloud.

- Do not rely on `gcr.io/cdssnc` Docker images anymore
- Install AWS CLI and kubectl from official sources
- Do not use 3rd party actions
- Use different AWS credentials to push containers and to deploy images in staging

I briefly changed the workflow to run it after every push to make sure that everything was running as expected. 

You can confirm that Docker images are available on https://gallery.ecr.aws/v6b8u5o6/notify-admin. If you look at admin pods in staging, they're using an AWS ECR Docker image at the moment as well.

Note: we requested the `cds-snc` alias for our organisation, our container URLs should be `public.ecr.aws/cds-snc/notify-admin` soon.


brdunfield (approved):
🦑 Less hard coding is 👍

PR #847, opened 2020-12-08
Css grids

brdunfield:
This is unfortunately a massive PR.

Most of the changes are just class name refactors - either renaming or splitting out old mixins (particularly 'contain-floats') into new class names that have now been added to the html.


app/assets/stylesheets/tailwind/components/grid.css
```diff
@@ -0,0 +1,160 @@
...
+    .column-one-eighth {
+      width: 12.5%;
+    }
+    .column-five-eighths {
+      width: 62.5%;
+    }
+    .column-seven-eights {
+      width: 87.5%;
+    }
+  }
```
AntoineAugusti, line 51: Could we use only Tailwind classes instead and get rid of this file? I feel like we're not deleting enough CSS files

brdunfield: Yes probably. I had a note to refactor these to be tailwind utilities that I didn't realise existed the first time around.

The issue is where the tests are looking for specific class names - but in this instance we can refactor those to look for the tailwind utilities.

tests/app/main/views/test_jobs.py
```diff
@@ -322,7 +322,7 @@ def test_should_show_letter_job(
         job_id=fake_uuid,
     )
     assert normalize_spaces(page.h1.text) == 'thisisatest.csv'
-    assert normalize_spaces(page.select('p.bottom-gutter')[0].text) == (
+    assert normalize_spaces(page.select('p.mb-12.clear-both')[0].text) == (
```
RenDavidDimen, line 325: This line appears to have a `.` in `'p.mb-12.clear-both'` where similar lines in the files do not. I'm not sure if that is intentional, just something I noticed 🙂

brdunfield:
Weird bug introduced in this PR which is breaking purgeCSS - in that it is purging things in the component files despite the ignore comment. 😠 I'm hoping I can figure out what is causing this..

tests/app/main/views/test_jobs.py
brdunfield: Yes, this is because the `bottom-gutter` class no longer exists, and has been replaced by the combination of `mb-12` and `clear-both` tailwind utility classes.

brdunfield:
Ok.. I'm seeing the same issue on master. So I'm not sure why staging is not affected.

brdunfield:
I think this has to do with the fact that staging is not pulling the latest instances anymore since the change to pull down AWS ECR images. Admin is still pushing images to GCR at this time.

Ultimately it seems as though the /*! purgecss start ignore */ comment does not work for imported dependencies, so I have restored those comments in every sub file for now. We can investigate this more and find a better solution if it exists, but I would just like to get this work in so that I can continue.

AntoineAugusti:
I don't see any problems, though this table is weird. Maybe rebase with `master`?
Screen Shot 2020-12-11 at 14 02 03

AntoineAugusti (approved):
🚢 🦑

PR #854, opened 2020-12-17
Revamp contact form with multiple steps

AntoineAugusti:
The new multi-steps contact form should be ready to go out as is as a first version. This is already a big improvement over what we have at the moment.

Follow up PRs I'll do:
- saving contact form data to the database to ease analysis (new API endpoint + model)
- maybe looking at improving the 2nd step of the demo funnel to have a textarea appear when selecting a radio button as it was in the Figma design file

You can play with the [review app](https://notification-contact-fo-rhey9i.herokuapp.com). This will trigger things to Freshdesk but I think it's important to do several manual tests. If you submit the form, you can go back to it if you navigate to another page. You can use also the back links at the very top.

@jimleroyer I would love your eyes on the code + tests part
@anikbrazeau I would love your eyes on the design + content part


app/main/views/contact.py
```diff
@@ -1,23 +1,157 @@
...
+    },
+    {
+        "form": SetUpDemoOrgDetails,
+        "current_step": "set_up_demo.org_details",
+        "previous_step": DEFAULT_STEP,
+        "next_step": "set_up_demo.primary_purpose",
+        "support_types": ["demo"],
+        "step": 1,
+        "total_steps": 2,
+    },
```
AntoineAugusti, line 39: After the first step (identity), there are 2 paths:
- the demo path with 2 more steps
- the message path, with a single textarea step

The `support_types` field on these forms determine where you will be routed based on the value of the `support_type` field filled in the first step.

app/main/views/contact.py
```diff
@@ -1,23 +1,157 @@
...
 
 
 @main.route('/contact', methods=['GET', 'POST'])
 def contact():
-    form = ContactNotifyTeam()
+    current_step = request.args.get('current_step', session.get(SESSION_FORM_STEP_KEY, DEFAULT_STEP))
+    try:
+        form_obj = [f for f in steps if f["current_step"] == current_step][0]
+    except IndexError:
+        return redirect(url_for('.contact', current_step=DEFAULT_STEP))
```
AntoineAugusti, line 57: Here only to handle the case when someone provides an invalid step in the URL, to avoid 500 errors

app/main/views/contact.py
```diff
@@ -1,23 +1,157 @@
...
+        possibilities = [f for f in steps if f["previous_step"] == current_step]
+        try:
+            if len(possibilities) == 1:
+                form_obj = possibilities[0]
+            else:
+                form_obj = [
+                    e for e in possibilities if form.support_type.data in e["support_types"]
+                ][0]
+        except IndexError:
+            return redirect(url_for('.contact', current_step=DEFAULT_STEP))
```
AntoineAugusti, line 78: Here only to handle the case when someone provides an invalid step in the URL, to avoid 500 errors

app/main/views/contact.py
```diff
@@ -1,23 +1,157 @@
...
-        # send email here
-        user_api_client.send_contact_email(form.name.data, form.email_address.data, form.feedback.data, form.support_type.data)
+        possibilities = [f for f in steps if f["previous_step"] == current_step]
+        try:
+            if len(possibilities) == 1:
+                form_obj = possibilities[0]
+            else:
+                form_obj = [
+                    e for e in possibilities if form.support_type.data in e["support_types"]
+                ][0]
```
AntoineAugusti, line 76: There are 2 possible forms after the identity step. Inspect the `support_type` value to route to the appropriate funnel

app/main/views/contact.py
```diff
@@ -1,23 +1,157 @@
...
+            f'- intended_recipients: {form.intended_recipients.data}',
+            f'- main use case: {form.main_use_case.data}',
+            f'- main use case details: {form.main_use_case_details.data}',
+        ])
+    else:
+        message = form.message.data
+
+    if current_user and current_user.is_authenticated:
+        user_profile = url_for('.user_information', user_id=current_user.id, _external=True)
+        message += f"<br><br>---<br><br> {user_profile}"
```
AntoineAugusti, line 143: Add a link to the Notify's user profile when logged in

app/main/views/index.py
```diff
@@ -37,31 +35,13 @@
...
-            'views/signedout.html',
-            form=form,
-            scrollTo="true",
-            admin_base_url=current_app.config['ADMIN_BASE_URL'],
-            stats=get_latest_stats(get_current_locale(current_app)),
-        )
-
     return render_template(
         'views/signedout.html',
-        form=form,
```
AntoineAugusti, line 64: Delete the contact form on the homepage. It's been replaced with a button to the contact form.

app/templates/components/page-header.html
```diff
@@ -1,11 +1,17 @@
...
 ) %}
 
   {% if back_link %}
     {{ govuk_back_link(back_link) }}
   {% endif %}
+  {% if sub_text %}
+  <div class="text-small text-gray-grey1 my-5">
+    <p>{{ sub_text }}</p>
+  </div>
+  {% endif %}
```
AntoineAugusti, line 14: Add potential sub text in gray. This is used to display when there are multiple steps.

anikbrazeau:
Looks great!! :D 

Could we adjust the question above the text box in Step 2 to be more conversational, to something like:  
**EN:** What will messages be about? 
**FR:** Sur quoi porteront les messages?

anikbrazeau (approved):
Looks good from a content perspective in heroku! Thanks for the quick implementation! 🙌

app/main/views/contact.py
jimleroyer: We are building small navigation trees out of these wizard. It's something to keep in mind if we extract a generic logic for reuse later on.

app/main/views/contact.py
```diff
@@ -1,23 +1,157 @@
...
+    },
+]
 
 
 @main.route('/contact', methods=['GET', 'POST'])
 def contact():
-    form = ContactNotifyTeam()
+    current_step = request.args.get('current_step', session.get(SESSION_FORM_STEP_KEY, DEFAULT_STEP))
+    try:
+        form_obj = [f for f in steps if f["current_step"] == current_step][0]
```
jimleroyer, line 55: A dictionary of dictionary (and maybe in a future iteration, a dictionary of `Step` objects?) instead of a list of dictionary could offer a more succinct and fitting approach here as we would also get safe retrieval and default fallback (via `dict#get`). I don't mind the current state btw, thinking loud if we want to generalize that kind of logic for wizard.

app/main/views/contact.py
```diff
@@ -1,23 +1,157 @@
...
+        try:
+            if len(possibilities) == 1:
+                form_obj = possibilities[0]
+            else:
+                form_obj = [
+                    e for e in possibilities if form.support_type.data in e["support_types"]
+                ][0]
+        except IndexError:
+            return redirect(url_for('.contact', current_step=DEFAULT_STEP))
+        form = form_obj["form"](data=session.get(SESSION_FORM_KEY, {}))
```
jimleroyer, line 79: I very much like that improvement (over what I did previously with the service creation form).

app/main/views/contact.py
```diff
@@ -1,23 +1,157 @@
...
+        form_obj = [f for f in steps if f["current_step"] == current_step][0]
+    except IndexError:
+        return redirect(url_for('.contact', current_step=DEFAULT_STEP))
+    form = form_obj["form"](data=session.get(SESSION_FORM_KEY, {}))
+
+    # Validating the final form
+    if form_obj["next_step"] is None and form.validate_on_submit():
+        session.pop(SESSION_FORM_KEY, None)
+        session.pop(SESSION_FORM_STEP_KEY, None)
+        send_form_to_freshdesk(form)
```
jimleroyer, line 64: What happens if sending the form to freshdesk fails? It seems the previous form did not handle error management too so this PR changes will not affect current functionality (but improve it in general). It's something though we should address in a future iteration, unless I am told that this is already handled in some logic I haven't seen yet.

jimleroyer (approved):
It looks great overall. This is a pragmatic improvement over the short wizard done with the service creation form and some changes from this PR could be applied back there. I like how the form management / detection is done and how it's kept in the session. 

Changes I would like to see in a future iteration:

* Error management if submitting the ticket to freshdesk fails over the API, including testing around it.
* Tests for the different wizard's paths. On that, if we generify the logic for wizard, we would save on tests overall with these.

Nothing blocking these changes and I know Antoine will continue some work on the form but wants to deliver a first version for now.

AntoineAugusti:
@jimleroyer 

Regarding
> Error management if submitting the ticket to Freshdesk fails over the API, including testing around it.

The admin only sends a request to the API and the API then sends a request to Freshdesk. The admin already has a retry strategy if Notify's API is unavailable. The retry logic for Freshdesk would have to be on Notify's API. It could be a matter of retrying up to 3 times the request or something more robust like queuing a job to do it.

As you pointed out, this logic was not there before so it's not a regression

jimleroyer:
@AntoineAugusti As soon as we touch the network, there is a risk of failure -- even to speak to its nearest neighbour. Is there a common failsafe in _notification-admin_ to avoid network failures for all calls and gracefully tell the user a problem happened?

AntoineAugusti:
@jimleroyer API requests are retried multiple times to avoid a failure after only 1 error https://github.com/cds-snc/notification-admin/pull/793 otherwise they will get the usual 500 error page

PR #856, opened 2020-12-18
Remove Gunicorn kill switch for keepalive

AntoineAugusti:
Follow up of https://github.com/cds-snc/notification-admin/pull/852

It's running fine in production so we don't need to keep this env variable around


dorianjp (approved):
removing code we don't need, thats dedication!

PR #857, opened 2020-12-21
Display API rate limit on platform admin

AntoineAugusti:
Services are rate limited by the number of notifications sent per day but there is also a rate limit per minute at the API level.

Display the rate limit on platform admin to ease debugging. We could have a follow-up task later to be able to change it on the UI rather than through the database only.

Screen Shot 2020-12-21 at 13 53 51


AntoineAugusti:
Did a small refactor to use `| format_number` in the second commit to format numbers differently in French/English.

jimleroyer:
Super useful. :shipit:

jimleroyer (approved):

PR #858, opened 2020-12-22
Sent Messages screen: `None` value is shown as data retention policy

jimleroyer:
# Summary | Résumé

Fixed the reported bug in this [Trello ticket](https://trello.com/c/WzSNNkhO/191-bug-none-jours) where a `None` value is displayed instead of a numeric value for data retention. policy in the *Sent Messages* screen. 

ss-sent-messages-none-jours

Steps to reproduce the bug:

- Having an account that only has permission to `manage_api_keys` -- the views are different depending on team member permissions.
- Then, once logged in to this state of permissions, selecting "Sent messages" in the menu.
- With zero messages sent, you should see _les messages sont conservés pendant None jours_ (French message but the English page also has the same bug).

If you want to strip down all permissions for your user to test via a SQL query, tailor the following one to your own service ID and user ID:

```
delete from permissions p
 where p.user_id = 'd0de5b14-1ba2-4d74-be9d-e7b9be26149f' 
   and p.service_id = '6ac41c2f-6fb6-4942-8eaa-6c4177556b19'
   and permission != 'manage_api_keys';
```

Overall, I went with my guts feeling of what the fix should be but it can be wrong. I had to take some special considerations on the fix because visiting the commit history from past developers, there was a conscious choice to deliver a `None` value if no channel type was selected (basically the screen display all channels). Here is an excerpt from [such commit history](https://github.com/cds-snc/notification-admin/commit/f3d1dc4a9c93c466b70979cd8e1426158154e6a8):

> Don’t ask for data retention unless channel is known

> If the user is looking at the notifications page for all message types
(which is what we show ‘caseworkers’) then it doesn’t make sense to ask
the API for the data retention period for that message type (because
it will be `None`). Doing so causes the API to return a `404`, which
then causes the admin app to return `404`.

> Passing through `None` as the value of limit days will just cause the
API to return everything in the `notifications` table, which is fine
for us.

Looking into calls for the `get_days_of_retention` which is ultimately called if the channel is defined, I've noticed that a default is loaded if not found via the `ACTIVITY_STATS_LIMIT_DAYS` config value. So I decided to reuse this value instead of initializing to `None` as a default. I let the rest of the logic be the same then.

I could have removed the condition that removes check to channel type, pass a `None` notification type which would then return the same default value I've setup but I am not knowledgeable enough of the application and I preferred to err on the side of caution. 

# Screenshot of fixed version

My current setup in the _admin_ project is screwed, it seems I have CSS issues, but the fix still is visible in the following SS:

ss-sent-messages-none-jours-fixed


app/main/views/jobs.py
```diff
@@ -256,7 +257,7 @@ def get_notifications(service_id, message_type, status_override=None):
         abort(404)
     filter_args = parse_filter_args(request.args)
     filter_args['status'] = set_status_filters(filter_args)
-    service_data_retention_days = None
+    service_data_retention_days = current_app.config.get('ACTIVITY_STATS_LIMIT_DAYS', 0)
```
AntoineAugusti, line 260:
```suggestion
    service_data_retention_days = current_app.config['ACTIVITY_STATS_LIMIT_DAYS']
```
I think I would put it like this as `ACTIVITY_STATS_LIMIT_DAYS` is currently hardcoded to `7`. If it was defined by an env in the config, there would be a fallback value, probably `0` or `None`.

jimleroyer: Where is the `ACTIVITY_STATS_LIMIT_DAYS` config variable defined? I am wondering what the fallback would be, because if it's `None`, we could get the same problem.

AntoineAugusti: It's hardcoded to 7 at the moment
https://github.com/cds-snc/notification-admin/blob/0294af08b4c092ddd218383d816ba6186ba88197/app/config.py#L84

If it was defined with an env there would be a fallback value

AntoineAugusti (approved):

app/main/views/jobs.py
jimleroyer: After talking to @AntoineAugusti offline, we have decided to put a programmatic default to `None` to fail fast if the configuration variable was not properly set with a default.

AntoineAugusti (approved):

PR #863, opened 2020-12-30
SMTP typo

RenDavidDimen:
Fixed what I believe to be a small typo and updated it's reference in the product


anikbrazeau (approved):
Nice catch!

PR #865, opened 2020-12-30
Features content update

anikbrazeau:
# Summary

Update content of Features section to focus on core offering and benefits/value

This includes:
- better descriptions of features and how people can take advantage
- cohesion with other parts of site
- alignment to value and benefits for GC audience
- clarifying dashboard use
- efficiency of automating messages

Figma reference: https://www.figma.com/file/bgHBy4o1STV4DlpkpdtIUV/Features-Compliance-Fonctionnalit%C3%A9s-Conformit%C3%A9?node-id=514%3A387

Eventually would be great to add images here to help users visualize each feature before adoption.

PR #866, opened 2020-12-30
New 'Why GC Notify' page content

anikbrazeau:
# Summary

Create new page for Why GC Notify (ahead of Beta) with available English and French content. Having this page will be useful to link to in marketing to decision makers who have questions about whether it meets all the GoC requirements.

[Link to the issue](https://trello.com/c/E3Gto2Pj/207-new-why-gc-notify-content-page)

This page includes information that :

- shows how GC Notify meets GC requirements and standards
- help people advocate for it more confidently
- highlights official languages (bilingual)
- federal identity program compliance
- privacy, public data protection
- security and reliability
- no approvals needed
- no procurement (free)
- cloud and digital standards

Figma reference: https://www.figma.com/file/bgHBy4o1STV4DlpkpdtIUV/Features-Compliance-Fonctionnalit%C3%A9s-Conformit%C3%A9?node-id=514%3A3

## Tests

* The new navigation menu was declaratively added to the existing ones for the test to properly assert expectations.
* The *why notify* page was added to the list of static pages to test for loading.

## Screenshots

The custom icons were added in this PR as well as displayed in [Figma screen design](https://www.figma.com/file/bgHBy4o1STV4DlpkpdtIUV/Features-%2B-Compliance-%7C-Fonctionnalit%C3%A9s-%2B-Conformit%C3%A9?node-id=514%3A3).

Why-use-GC-Notify-–-Notify


app/assets/stylesheets/tailwind/style.css
```diff
@@ -174,6 +174,15 @@ label {
...
 }
 
+.checkmark {
+  display: inline-block;
+  transform: rotate(45deg);
+  height: 28px;
+  width: 17px;
+  border-bottom: 7px solid #00823B;
+  border-right: 7px solid #00823B;
+}
```
jimleroyer, line 184: This is a CSS based checkmark. I didn't add it to the HTML yet but I can try tomorrow; this requires a bit more CSS tailwind machinery to correctly align it as showed in the figma diagram. The following screenshot shows the broken version:
image

app/templates/views/signedout.html
```diff
@@ -183,7 +192,7 @@ <h2 class="pb-0 font-sans text-white leading-tight font-bold text-lg">
         <div class="pt-6 pb-6 ml-10 image-content hidden lg:block lg:visible">
           <img
             loading="lazy"
-            alt="An illustration of a laptop with a message window on the screen and a cellphone with a message and a callout blurb.") }}"
+            alt="An illustration of a laptop with a message window on the screen and a cellphone with a message and a callout blurb."
```
jimleroyer, line 195: I removed the extra curly and round braces. It seems like a typo for a potential translation of the alt image text.

app/assets/stylesheets/tailwind/style.css
jimleroyer: I realized though that if we support IE 11, we should use an image, because it does not support much in terms of CSS rotate transformation. 😢 

https://caniuse.com/?search=rotate

jimleroyer: In the end, I reused the existing `tick.png` image which is what I was searching for in a checkmark.

app/templates/views/why-notify.html
```diff
@@ -0,0 +1,62 @@
...
+</div>
+<div class="pl-24">
+  <img src="/static/images/tick.png" class="float-left" style="width: 4rem; height: 4rem; margin-left:-5rem;"/>
+  <h2 class="heading-medium">{{ _('Build modern, cloud-first services') }}</h2>
+</div>
+<div class="pl-24">
+  <p>{{ _('GC Notify is a scalable cloud platform with a robust API that allows you to connect to other services. For example, connect to electronic mailing list software like Listserv or other applications to send password resets, email confirmations, or responses to forms.') }}</p>
+</div>
+<div class="pl-24">
+  <img src="/static/images/tick.png" class="float-left" style="width: 4rem; height: 4rem; margin-left:-5rem;"/>
```
jimleroyer, line 25: I would prefer to have the style declared as a class in our CSS files. I will look what I can do tomorrow morning but I am opened to suggestion on proper practice in the project.

app/templates/views/why-notify.html
```diff
@@ -0,0 +1,62 @@
...
+  <p>{{ _('GC Notify does not store contact lists and only stores information about recipients for 7 days. You can also hide sensitive information by redacting personal details from messages after they’ve been sent, so that only recipients see their personal information.') }}</p>
+  <p><a href="{{ url_for('main.privacy') }}">{{ _('Learn more about privacy protection') }}</a></p>
+</div>
+<div class="pl-24">
+  <img src="/static/images/tick.png" class="float-left" style="width: 4rem; height: 4rem; margin-left:-5rem;"/>
+  <h2 class="heading-medium">{{ _('Keep data secure') }}</h2>
+</div>
+<div class="pl-24">
+  <p>{{ _('GC Notify keeps your account secure with two-factor authentication. This means you’ll need to enter a code that gets sent to your phone or email when you sign in. GC Notify is authorized for Protected A information. Data is encrypted whenever it passes through GC Notify and when it is stored in the system.') }}</p>
+  <p><a href="{{ url_for('main.security') }}">{{ _('Learn more about security') }}</a>.
```
AntoineAugusti, line 45:
```suggestion
  <p><a href="{{ url_for('main.security') }}">{{ _('Learn more about security') }}</a>
```
No final dot in Figma and to be consistent with other links

app/templates/views/why-notify.html
```diff
@@ -0,0 +1,62 @@
...
+{% block content_column_content %}
+
+<div class="pl-24">
+  <h1 class="heading-large">{{ _('Why use GC Notify') }}</h1>
+</div>
+<div class="pl-24">
+  <p>{{ _('GC Notify is designed in a way that meets the needs of government services.') }}</p>
+</div>
+<div class="pl-24">
+  <img src="/static/images/tick.png" class="float-left" style="width: 4rem; height: 4rem; margin-left:-5rem;"/>
```
AntoineAugusti, line 18: You should use `{{ asset_url('images/tick.png') }}` (top of my head, may want to check elsewhere)

app/templates/views/why-notify.html
```diff
@@ -0,0 +1,62 @@
+{% extends "content_template.html" %}
+{% from "components/table.html" import mapping_table, row, text_field, edit_field, field with context %}
```
AntoineAugusti, line 2: Clean up unused imports

app/templates/views/why-notify.html
```diff
@@ -0,0 +1,62 @@
...
+{% from "components/sub-navigation.html" import sub_navigation %}
+
+{% block per_page_title %}
+  {{ _('Why use GC Notify') }}
+{% endblock %}
+
+{% block content_column_content %}
+
+<div class="pl-24">
```
AntoineAugusti, line 11: The page is currently not responsive, ticks are outside of the viewport on mobiles.

app/templates/views/why-notify.html
```diff
@@ -0,0 +1,62 @@
...
+  <p>{{ _('GC Notify keeps your account secure with two-factor authentication. This means you’ll need to enter a code that gets sent to your phone or email when you sign in. GC Notify is authorized for Protected A information. Data is encrypted whenever it passes through GC Notify and when it is stored in the system.') }}</p>
+  <p><a href="{{ url_for('main.security') }}">{{ _('Learn more about security') }}</a>.
+</div>
+<div class="pl-24">
+  <img src="/static/images/tick.png" class="float-left" style="width: 4rem; height: 4rem; margin-left:-5rem;"/>
+  <h2 class="heading-medium">{{ _('Skip the procurement process') }}</h2>
+</div>
+<div class="pl-24">
+  <p>{{ _('GC Notify is a government-run service, using it does not require any external procurement. You can send up to 10 million emails and 10,000 text messages per year for free for each of your Government of Canada services. No set up fee and no cost to use.') }}</p>
```
AntoineAugusti, line 52: My understanding is that we don't any limits for emails and we have a variable limit for SMS. Could you check this information and use a variable? We have `DEFAULT_FREE_SMS_FRAGMENT_LIMITS` in config.py, would be a good thing to update it there as well 👌

AntoineAugusti: Integers can be formatted in French/English with `| format_number`

app/templates/views/signedout.html
AntoineAugusti: The alt text shouldn't be in English only

app/templates/views/why-notify.html
AntoineAugusti: You should be able to use utility classes for that without declaring a new CSS class.

Take a look at:
https://tailwindcss.com/docs/height
https://tailwindcss.com/docs/width
https://tailwindcss.com/docs/margin

AntoineAugusti (changes requested):
Can you not commit the following files as they are not related to changes?
- app/assets/javascripts/main.min.js
- package-lock.json

If you're changing Tailwind styles, you can run `npm run tailwind` and do not commit an updated version of the main.min.js file. I think @brdunfield said that it currently is broken and we need to take a look.

Made some commits about responsive and content as well

app/templates/views/why-notify.html
AntoineAugusti: For these icons, you should add `aria-hidden="true"` to say that these images do not convey meaningful information and are just decorative

app/templates/views/why-notify.html
jimleroyer: It's true, but I was thinking more of a single class regrouping these properties in the `style` attribute, or a reusable component.

jimleroyer:
> Finally, I've seen that the Figma file has some unique icons per section, could you use these instead of the generic tick every time?

I don't understand what you mean. There is one green checkmark for all of the subtitles. Can you rephrase?

AntoineAugusti:
@jimleroyer The [Figma file](https://www.figma.com/file/bgHBy4o1STV4DlpkpdtIUV/Features-%2B-Compliance-%7C-Fonctionnalit%C3%A9s-%2B-Conformit%C3%A9?node-id=514%3A3) has multiple screens, 2 for English (not sure why). The bottom one has various icons and I like this version 🎨

app/templates/views/why-notify.html
AntoineAugusti: If this is only used in this page, the Tailwind way would be to duplicate a bunch of utility classes in the `class` attribute of these images

jimleroyer:
@AntoineAugusti Can you post a screenshot of what you see on Figma? I see one page for each English and French version and they all have the green checkmark.

app/templates/views/signedout.html
jimleroyer: Absolutely.. there are 23 instances of `alt` text that are not localized. Is that okay if I open a new ticket for this?

AntoineAugusti: Sure, this is not blocking for this PR if this is not a one time thing 👍

app/templates/views/why-notify.html
jimleroyer: Hmm I tested it with iphone 6 resolution and it seems to work:

Can you share with me what is not working? or a screenshot on your end of what you have?

A screenshot of what I have, not with the ticks but the new icons:

Why-use-GC-Notify-–-Notify--mobile

app/templates/views/why-notify.html
jimleroyer: One thing we could do is to put these sections into a macro, like what [tailwind documentation recommends](https://tailwindcss.com/docs/extracting-components) but with using *vue.js*. Maybe it isn't enough for now though.

app/templates/views/why-notify.html
jimleroyer: I see we have `DEFAULT_SERVICE_LIMIT` defined like this on service creation:

```
service_id = service_api_client.create_service(
            service_name=service_name,
            organisation_type=organisation_type,
            message_limit=current_app.config['DEFAULT_SERVICE_LIMIT'],
            restricted=True,
            user_id=session['user_id'],
            email_from=email_from,
            default_branding_is_french=default_branding_is_french,
        )
```

Is this okay to assume that this would be a limit on emails? I will make changes as this is the case for now and remove it if not.

jimleroyer: Actually, forget about it, it does not make sense as the config variable is set to `50` (and it's way too low obviously).

jimleroyer: Let's talk tomorrow on Slack about this.

jimleroyer: Regarding the limit on messages: I spoke to Bryan and he confirms that 10 million emails per service per year would be the limit. After limit would be reached, a MOU would need to be agreed upon with the service that gets out of this limit. It does not seem like there is any hard limit coded in though, but the text description would be valid per our business rules.

jimleroyer: @anikbrazeau @AntoineAugusti @willeybryan @ClementineHahn For reference here ^^^^ we should probably concert on the email limit per year.

willeybryan: We have a draft MOU that Policy is helping with. We don't have anyone on an MOU right now, we just want to have one available. The limits per year are business limits and are not hard limits. We will make exceptions. 

The 10 Million email one was picked because at a cost of $1000 USD we want to negotiate an MOU with that departments. For SMS We picked 10,000 because of the unknown factors, such as international SMS potentially causing large cost surprises. 

These limits are just for when we need to start thinking about collecting $$$ in the form of cost recovery. We can control system limits differently. 

We aren't doing any cost recovery right now, and we may not for a while, but this will help set expectations for department on when that kicks in. 

We may want to put in more info explaining that More is available it's just gonna need cost recovery.

app/templates/views/signedout.html
anikbrazeau: 😅 happy to help fill in those alt text gaps! do you have a link to that new ticket?

jimleroyer: Just created an issue in the parking lot column: https://trello.com/c/ByGjBswg/300-translate-alternative-text-on-images-in-the-signed-out-view-for-accessibility

After more careful look, the issue is that we have 1 instance of a text that is not translated and numerous instances of hardcoded translation values using `if... else... endif` conditions instead of our newer CSV translation mechanism.

AntoineAugusti (approved):
Great work @anikbrazeau @jimleroyer and the design team.

Good to go 🚢 

@jimleroyer, you'll new to merge in the `master` branch. You may need to run `npm run tailwind` one last time, don't include the `main.min.js` file in your changes. This is required because https://github.com/cds-snc/notification-admin/pull/865 just landed on the other branch

PR #869, opened 2021-01-05
Translate Email authentication

AntoineAugusti:
Screen Shot 2021-01-05 at 13 26 58

Translated a few more strings at the same time. These strings are already in CSV files, the issue was only on the code side.

Related Trello card: https://trello.com/c/N8VVJxpt/224-missing-french-string-email-authentication

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -13,6 +13,7 @@ from flask import (
     url_for,
 )
 from flask_babel import _
+from flask_babel import lazy_gettext as _l
 from flask_login import current_user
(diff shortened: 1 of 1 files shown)


anikbrazeau (approved):
Looks good in heroku!
Question: Do we need the upload_letters bit?

AntoineAugusti:
@anikbrazeau It's not in use/visible but it's already translated. We have follow-up tasks to remove unused pages/features so this piece may go later!

As you know, removing unused stuff takes time if we do a deep clean of content/logic/views/API

PR #873, opened 2021-01-06
Remove duplicate translation for contact us

AntoineAugusti:
"contact us" was translated in French in 2 various ways:
- nous joindre
- communiquer avec nous

The current translation system does not support 2 different wording for the same source therefore I'm removing one version (it was not used because of the technical limitation)


anikbrazeau (approved):
🚀

PR #876, opened 2021-01-08
Add missing string values for both english and french

RenDavidDimen:
Strings were missing from the En/Fr CSV files. 

I've added the required strings and translations to address the issue of the default english error message appearing on a the french pages.


AntoineAugusti (approved):
Works fine on the review app, thanks!

Screen Shot 2021-01-08 at 09 59 59

🚢

PR #874, opened 2021-01-07
Decouple email servicename

dorianjp:
https://trello.com/c/yeyTL5es/158-decouple-email-and-service-name

- Added new page to wizard for creating a service.
- Added new item for editing email from address in settings.
- Moved validations to wizard steps in form vs. end submit validation
- Modified validations for service name
- Modified validations for email from
- New shared form validators
- Refactoring on create service wizard
  - remove hardcoding on steps
  - refactored next steps process from client-side
  - removed organisation type form hidden fields as it was not used in system
  - added api_client validations
  - new template page for wizard
  - content changes throughout to align with design
  - test fixes

PR #881, opened 2021-01-13
Remove line in logo page

anikbrazeau:
# Summary | Résumé

Remove redundant content on logo page, as per suggestion on https://github.com/cds-snc/notification-admin/pull/874

Removed top sentence from `/add-service?current_step=choose_logo` page and both csv translation files:
"The default logo use in email messages is the English Government of Canada signature." 
"Le logo par défaut utilisé dans vos messages courriel est la signature du gouvernement du Canada en anglais d'abord."


AntoineAugusti (approved):
LGTM, thanks for cleaning that up and opening a PR

PR #883, opened 2021-01-14
fix reference to months so french strings appear when expected

RenDavidDimen:
# Summary | Résumé

Fixing month string reference so French months appear when expected.

Trello: https://trello.com/c/ms7DKi9n/135-missing-french-strings-months-in-template-usage


dorianjp:
[SNS warning Topic should only be used by Alarms](https://trello.com/c/4e58AlOx/178-sns-warning-topic-should-only-be-used-by-alarms)

PR #893, opened 2021-01-27
feat: move SMS senders to platform admin

AntoineAugusti:
Move settings
- Receive text messages
- Text message senders

to the Platform admin section and make sure that service owners cannot change these settings themselves.

I've kept tests testing the behaviour of text message senders, just updated those to make sure that the logged-in user was a platform admin to perform these actions.

## Do we need to change the API?
My understanding is no.

SMS sender HTTP endpoints are defined here (just 1 HTTP endpoint, others are present if you search for `sms-sender`)
https://github.com/cds-snc/notification-api/blob/757e94bcec48347663d978e9bd10121f1eaf7b21/app/service/rest.py#L783

And the blueprint that makes sure that requests come from the admin
https://github.com/cds-snc/notification-api/blob/757e94bcec48347663d978e9bd10121f1eaf7b21/app/__init__.py#L134-L135

As a user, you cannot handcraft HTTP requests and ask the admin to do those to the API, so we're safe.

Other "platform admin only" things like inbound numbers don't have a different authentication mechanism.
https://github.com/cds-snc/notification-api/blob/757e94bcec48347663d978e9bd10121f1eaf7b21/app/__init__.py#L158-L159

## How to test this
Use the review app and go to "Service settings" with a Platform admin account/with a normal user account.

## Screenshots
New settings in Platform Admin
Screen Shot 2021-01-27 at 09 37 22

Fewer settings in "Text message settings"

Screen Shot 2021-01-27 at 09 36 54

Trello card: https://trello.com/c/ugFh402K/250-make-text-message-senders-platform-admin-only

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -864,7 +864,7 @@ def service_delete_letter_contact(service_id, letter_contact_id):
 
 
 @main.route("/services/<service_id>/service-settings/sms-sender", methods=['GET'])
-@user_has_permissions('manage_service', 'manage_api_keys')
+@user_is_platform_admin
 def service_sms_senders(service_id):
     return render_template(
         'views/service-settings/sms-senders.html',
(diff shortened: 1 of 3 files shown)


dorianjp (approved):
clever how it auths also based on the client in the API, so no API changes required

PR #895, opened 2021-01-27
feat: no identity step for logged in users

AntoineAugusti:
Don't ask for the name and email address on the contact form if the user is logged in. These 2 fields are hidden and values are taken directly from the user's profile.

## Screenshot while logged in
Screen Shot 2021-01-27 at 14 06 09


tests/app/main/test_contact.py
```diff
@@ -88,7 +115,7 @@ def test_invalid_step_name_redirects(client_request):
 @pytest.mark.parametrize('support_type', [
     'ask_question', 'technical_support', 'give_feedback', 'other'
 ])
-def test_message_step_valides(client_request, support_type, mocker):
+def test_message_step_validates(client_request, support_type, mocker):
```
AntoineAugusti, line 135: Fixed a typo

tests/app/main/test_contact.py
```diff
@@ -18,7 +18,32 @@ def test_contact_page_does_not_require_login(client_request):
     client_request.get('.contact')
```
AntoineAugusti, line 18: Added a bunch of `client_request.logout()` because previously I was using `client_request.post` and this uses a logged-in account. Previously I overwrote the name and email address of the default logged-in user but now that these fields are taken from the user profile, tests did not pass.

I logged out before making requests and added a specific test to demonstrate that the email address and name field are not shown but that the form works as intended anyway.

tests/app/main/test_contact.py
```diff
@@ -308,14 +354,6 @@ def submit_form(keys):
...
     # Fields are blank
     assert page.select_one('input[checked]') is None
-    assert [
-        (input['name'], input['value'])
-        for input in page.select('input')
-        if input['name'] in ['name', 'email_address']
-    ] == [
-        ('name', ''),
-        ('email_address', '')
-    ]
```
AntoineAugusti, line 318: Those fields don't exist anymore if you're logged in, as in this test case

jimleroyer (approved):
LGTM

PR #897, opened 2021-02-01
Tweaks to clean up "Create an account" page

anikbrazeau:
# Summary | Résumé

Cleaning up the `/register` page with small tweaks
- Remove long redundant Privacy summary.
- Add short Terms of use link instead.
- Summarized Terms of use to figure in Go live process.


AntoineAugusti (approved):
LGTM

Did a quick commit to have a single translation instead of two as it is really a sentence and not 2 parts.

PR #909, opened 2021-02-05
feat: add create an account on why notify

AntoineAugusti:
Add a nudge to register "Create an account and try it for yourself" at the bottom of the "Why Notify" page if folks are not logged in.

Screen Shot 2021-02-05 at 12 58 49

Something similar is done on the "Features" page as well.


anikbrazeau (approved):
Super rapide! Fantastique! 🚀

PR #913, opened 2021-02-19
feat: update fallback for geolocate service

AntoineAugusti:
Update the fallback mechanism for the env `IP_GEOLOCATE_SERVICE`. Previously, in order to deactivate this feature the env variable would need to be absent. Having it present and set to a `""`/`False` would not disable the feature.

This PR makes sure that the feature is only enabled if we put an HTTP-like value. It also removes the required `/` at the end of the domain (it's unsual to be forced to put a base URL ending with a slash)
https://github.com/cds-snc/notification-manifests/blob/23f3bb87a8e4900a5b000b65f26e9237e90bcd07/base/admin-deployment.yaml#L93-L94 and now accepts a value ending with a `/` or not.

### DIFF
--- a/tests/app/main/views/test_sign_in.py
+++ b/tests/app/main/views/test_sign_in.py
@@ -265,7 +265,7 @@ def test_sign_in_geolookup_disabled_in_dev(
     mock_get_security_keys,
     mocker
 ):
-    assert current_app.config.get('IP_GEOLOCATE_SERVICE') is None
+    assert current_app.config['IP_GEOLOCATE_SERVICE'] == ""
 
     mocker.patch('app.user_api_client.get_user', return_value=api_user_active_email_auth)
     mocker.patch('app.user_api_client.get_user_by_email', return_value=api_user_active_email_auth)
--- a/app/config.py
+++ b/app/config.py
@@ -101,7 +101,7 @@ class Config(object):
 
     HIPB_ENABLED = True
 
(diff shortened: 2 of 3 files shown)


RenDavidDimen (approved):
Changes look good to me 👍

RenDavidDimen:
Just out of curiosity. I had made some small changes that I wanted to put up for review that covered the same goal. Is there a reason that the `IP_GEOLOCATE_SERVICE` env variable isn't removed from the project?

tests/app/main/views/test_sign_in.py
```diff
@@ -265,7 +265,7 @@ def test_sign_in_geolookup_disabled_in_dev(
     mock_get_security_keys,
     mocker
 ):
-    assert current_app.config.get('IP_GEOLOCATE_SERVICE') is None
+    assert current_app.config['IP_GEOLOCATE_SERVICE'] == ""
```
RenDavidDimen, line 268: Just out of curiosity. I had made some small changes that I wanted to put up for review that covered the same goal. Is there a reason that the IP_GEOLOCATE_SERVICE env variable isn't removed from the project?

AntoineAugusti: > Is there a reason that the IP_GEOLOCATE_SERVICE env variable isn't removed from the project?

What do you mean? Not sure I understand but happy to clarify

PR #916, opened 2021-02-26
feat: generate EN translations from FR file

AntoineAugusti:
At the moment we leverage 2 CSV files for translations on the admin:
- a French one
- an English one

The English one can be derived from the French automatically but this is not done. At the moment we spend a lot of time copy/pasting strings from one file to another.

Let's make this a tiny bit better and remove the need for the `en.csv` file. This file is generated based on: a new file, `en_base.csv` with some English overrides and then leverages keys from `fr.csv`.

Trello card: https://trello.com/c/yE8y27Ul/428-improve-translation-process-on-the-admin


scripts/generate_en_translations.py
```diff
@@ -0,0 +1,26 @@
+import csv
+from pathlib import Path
+
+root_folder = Path(__file__).parent.parent.absolute()
+translation_folder = "app/translations/csv"
```
jimleroyer, line 5: The `pathlib` module seems quite handy. This above also works on Windows.

AntoineAugusti: Yeah, love it, the API is so nice to read.

scripts/generate_en_translations.py
```diff
@@ -0,0 +1,26 @@
...
+
+fr_file = root_folder / translation_folder / "fr.csv"
+en_base_file = root_folder / translation_folder / "en_base.csv"
+en_file = root_folder / translation_folder / "en.csv"
+
+headers = ["source", "target"]
+
+with open(fr_file) as fr, open(en_base_file) as en_base, open(en_file, "w") as en:
+    reader = csv.DictReader(fr, fieldnames=headers)
```
jimleroyer, line 14: Can we rename to `fr_reader`? It would be more explicit for code readability to me.

AntoineAugusti: 💯 👍

jimleroyer (approved):
:shipit:

scripts/generate_en_translations.py
```diff
@@ -14,7 +14,14 @@
     fr_reader = csv.DictReader(fr, fieldnames=headers)
     base_en_reader = csv.DictReader(en_base, fieldnames=headers)
 
-    # Write a few English overrides
+    # Write the few English overrides we have.
+    #
+    # Rows in `en_base.csv` are what an ideal translation system
+    # should be: they map keys to values, changing what's displayed
+    # in English when used in HTML templates.
```
jimleroyer, line 21: 👍

PR #918, opened 2021-03-02
Update AWS CLI when building containers

AntoineAugusti:
# Summary | Résumé

GitHub Actions recently started using Ubuntu 20.04 when we specify we want to use the `ubuntu-latest` image. The AWS CLI v2 is already installed in this new version, see https://github.com/actions/virtual-environments/issues/1816

Our [CI job failed](https://github.com/cds-snc/notification-admin/runs/2013660441?check_suite_focus=true) because we tried to install it even if it was already present.

Add the `--update` flag to install the latest available version and override the default one GitHub gives us.


qcorporation-cds (approved):
LGTM

PR #920, opened 2021-03-04
feat: add Google Analytics on homepage

AntoineAugusti:
Moved analytics to a partial and require it on the homepage, where it was missing before. Likely due to the fact that the homepage does not inherit from the Notify base template.

Fixing the analytics stuff now but will write [a refactor card](https://trello.com/c/ehZaCZgO/440-ensure-that-the-homepage-uses-the-notify-template) for the homepage.

## Testing

Open the review app, inspect the source code, see that we have `gtag` at the very end.

Trello card: https://trello.com/c/p4fexMtQ/355-put-google-analytics-on-the-home-page


app/templates/partials/analytics.html
```diff
@@ -0,0 +1,22 @@
...
+></script>
+<script>
+  window.dataLayer = window.dataLayer || []
+
+  function gtag() {
+    dataLayer.push(arguments)
+  }
+
+  gtag("js", new Date())
+  gtag("config", "UA-102484926-14", {anonymize_ip: true})
```
AntoineAugusti, line 14: `anonymize_ip` was already here but it was a requirement of the Trello card

davidbgk:
The snippet from https://minimalanalytics.com/ might be a good option considering the performance gain.

AntoineAugusti:
Hey @davidbgk, thanks for the suggestion! We'll hold off for now, we've got a lot of other things to tackle first regarding frontend performance on Notify, such as caching.

For example:
```
curl -I https://notification.canada.ca/static/stylesheets/index.css | grep -i cache
cache-control: no-store, no-cache, private, must-revalidate
```

caused by https://github.com/cds-snc/notification-admin/blob/e1ef468b762c78c6120e50c9d91b610f26baf609/app/__init__.py#L640-L646 and having private cache headers for all things is definitely not ideal.

davidbgk:
Hey @AntoineAugusti 👋

Indeed, not ideal! I'm surprised you have to set these headers at the application level, is that to avoid putting nginx on top of gunicorn?</offtopic-but-curious>

AntoineAugusti:
@davidbgk That's a very old GDS leftover we need to take care of https://github.com/cds-snc/notification-admin/commit/7d4c2045bcf4077504e7b4d1d301ddd6ac973cdd https://github.com/cds-snc/notification-admin/commit/7fc1db9d19e9c4c95746bfb98eb4c97c8dccc378. 

We run a load balancer in front of admin deployments and we'll add CloudFront in front of the AWS load balancer later, but this code needs to change first

jimleroyer (approved):
LGTM

PR #921, opened 2021-03-04
Fix alignment / padding for the "Why GC Notify" page

anikbrazeau:
# Summary | Résumé

To be consistent with other unauthenticated pages, removed left padding from heading and paragraph text


anikbrazeau:
[NAV Bug: Page content alignment](https://trello.com/c/i6KcX8pk/335-nav-bug-page-content-alignment)

app/templates/views/why-notify.html
```diff
@@ -5,13 +5,13 @@
 {% endblock %}
 
 {% block content_column_content %}
-
-<div class="pl-24">
+<div>
```
AntoineAugusti, line 8: Could remove this div if it's empty

app/templates/views/why-notify.html
```diff
@@ -5,13 +5,13 @@
...
   <h1 class="heading-large">{{ _('Why use GC Notify') }}</h1>
 </div>
-<div class="pl-24">
+<div>
```
AntoineAugusti, line 11: Could remove this div if it's empty

AntoineAugusti (changes requested):

AntoineAugusti (approved):

PR #924, opened 2021-03-05
Update settings page with new EN/FR content including displaying sending limits to users

QuanNguyen-CDS:
# Summary | Résumé
[Trello](https://trello.com/c/Y9vpgrWn)
refactor: settings ui changes for annual limits
refactor: updating units to test for a wide range of strings with restricted=False
refactor: config has annual limits for email and sms (hardcoded for now)
refactor: translations in french per figma mockups
fix: fixme note because `10 million emails` isn't handled if annual email limit changes to 1M
refactor: altered settings_row macro to add alternative condition (to handle .live property)

1. Create a service, keep it in trial mode, validate that "Daily message limit" is set to 50 notifications
2. From 1. Validate in french the *new* wording is "Limite d'envoi quotidienne" before it was just "Limite quotidienne"
3. Turn the service live, and validate:
- Daily message limit is now "10,000 notifications", in french "10 000 notifications"
- [new field] Yearly free maximum email is now "10 million emails", in french "10 millions de courriels"
- [new field] Yearly free maximum sms is now "25,000 text messages, in french "25 000 messages texte"
See attachment for more details
4. Run `test_service_settings.py` and validate that the tests all pass (or skipped intentionally)

> Étapes consécutives (1., 2., 3., …) qui décrivent la façon de tester la modification. Elles aideront les développeurs à faire des tests sans avoir à jouer au détective. Veuillez aussi inclure toutes les étapes de configuration de l’environnement qui ne font pas partie des étapes normales dans le fichier README et tout élément temporel requis.

Due to the nature of the hard code values of 10M annual email messages and 25000 annual sms messages, the UI is also hard coded for the translations. The translations for french do not take into account pluralization if for example the value is reduced to 1 million. When we have a customizable field we will want to handle these edge cases properly

settings-changes


app/config.py
```diff
@@ -109,6 +109,9 @@ class Config(object):
     CSV_MAX_ROWS = os.getenv('CSV_MAX_ROWS', 50000)
     CSV_MAX_ROWS_BULK_SEND = os.getenv('CSV_MAX_ROWS_BULK_SEND', 100000)
 
+    ANNUAL_SMS_LIMIT = os.getenv('ANNUAL_SMS_LIMIT', 25000)
```
AntoineAugusti, line 112: Could you make it clear that this is the "free tier" and not a limit?

We already had a variable for that https://github.com/cds-snc/notification-admin/blob/ce8b8649984c28e7ddc0e2074cf89c368d746d40/app/config.py#L57-L66

Though at the moment we were leveraging only the `central` value, other values are unused at the moment for us and we didn't clean the code yet. Maybe switch to a clear "free yearly SMS limit" env will help. In that case, you can refactor where the old env value is used?

Finally, you'll need to cast the value to an int: if the value is coming from an env variable, it'll be a string and it should be cast to an integer.

app/config.py
```diff
@@ -109,6 +109,9 @@ class Config(object):
...
+    ANNUAL_EMAIL_LIMIT = os.getenv('ANNUAL_EMAIL_LIMIT', 10000000)
```
AntoineAugusti, line 113: Same comment about casting + free limit.

Additionally, you can use underscores to improve readability for the default value https://www.python.org/dev/peps/pep-0515/

app/templates/views/service-settings.html
```diff
@@ -184,7 +185,13 @@ <h2 class="heading-small mt-5">{{ _('Your service is in TRIAL mode') }}</h2>
             attributes=""
           )}}
         {% endcall %}
-
+        {% call settings_row(if_has_permission='email', alt_cond=current_service.live) %}
```
AntoineAugusti, line 188: Do you think `alt_cond` is needed? I think elsewhere it leverages a simple `if` condition to choose to display the value or not

tests/app/main/views/test_service_settings.py
```diff
@@ -128,7 +132,8 @@ def test_should_show_overview(
     service_one = service_json(SERVICE_ONE_ID,
                                users=[api_user_active['id']],
                                permissions=['sms', 'email'],
-                               organisation_id=ORGANISATION_ID)
+                               organisation_id=ORGANISATION_ID,
+                               restricted=False)
```
AntoineAugusti, line 136: Do we still have a test case with `restricted=True` now?

app/main/views/service_settings.py
```diff
@@ -80,10 +80,19 @@
 
 @main.route("/services/<service_id>/service-settings")
 @user_has_permissions('manage_service', 'manage_api_keys')
-def service_settings(service_id):
+def service_settings(service_id: str):
+
+    # FIXME: the formatting for annual email limits and translation within service-settings.html can be improved
```
AntoineAugusti, line 85: WDYT about raising an exception if the free yearly limit for email is < 1M? At the moment it'll break translations.

app/main/views/service_settings.py
```diff
@@ -80,10 +80,19 @@
...
+    limits = {
```
AntoineAugusti, line 86: Feel free to inline in `render_template` if you feel like it

AntoineAugusti (changes requested):

app/config.py
QuanNguyen-CDS: will update - thanks for the clarification

app/config.py
QuanNguyen-CDS: good suggestion, will update the file with this suggestion include the other constants numbers (to align)

app/templates/views/service-settings.html
QuanNguyen-CDS: My interpretation of the FIgma mockups ([link-live](https://www.figma.com/file/A4yAAkCGUaqoaTjpWL411i/Going-live-%7C-Activation-du-service?node-id=498%3A157) vs [link-trial](https://www.figma.com/file/A4yAAkCGUaqoaTjpWL411i/Going-live-%7C-Activation-du-service?node-id=299%3A380)) has a clear differentiation of when a service is in trial and when a service is live. So we will also need this differentiation within the code to display certain settings in the two use cases

tests/app/main/views/test_service_settings.py
QuanNguyen-CDS: we do not - the test cases do not cover all permutations.
I decided to adapt this test case to restricted=True as it tests all the fields.

One suggest is we can adapt this test case to include more permutations - just let me know if you'd like to do that

app/templates/views/service-settings.html
AntoineAugusti: My comment was questioning the need to change the component `settings_row` to add a new parameter, but I think it's a good idea 👍

tests/app/main/views/test_service_settings.py
AntoineAugusti: > One suggest is we can adapt this test case to include more permutations - just let me know if you'd like to do that


I didn't look carefully at all the test cases covering this area, you can decide 👍

app/main/views/service_settings.py
QuanNguyen-CDS: did not make it inline as i need it for the assert now, marking as resolved

tests/app/main/views/test_service_settings.py
QuanNguyen-CDS: I think this is good for now to validate the fields (existing as well as newly added), the additional test cases we could address later

app/config.py
```diff
@@ -106,11 +106,11 @@ class Config(object):
...
     BULK_SEND_TEST_SERVICE_ID = os.getenv('BULK_SEND_TEST_SERVICE_ID')
-    CSV_MAX_ROWS = os.getenv('CSV_MAX_ROWS', 50000)
-    CSV_MAX_ROWS_BULK_SEND = os.getenv('CSV_MAX_ROWS_BULK_SEND', 100000)
+    CSV_MAX_ROWS = os.getenv('CSV_MAX_ROWS', 50_000)
+    CSV_MAX_ROWS_BULK_SEND = os.getenv('CSV_MAX_ROWS_BULK_SEND', 100_000)
 
-    ANNUAL_SMS_LIMIT = os.getenv('ANNUAL_SMS_LIMIT', 25000)
-    ANNUAL_EMAIL_LIMIT = os.getenv('ANNUAL_EMAIL_LIMIT', 10000000)
+    FREE_YEARLY_SMS_LIMIT = os.getenv('FREE_YEARLY_SMS_LIMIT', 25_000)
+    FREE_YEARLY_EMAIL_LIMIT = os.getenv('FREE_YEARLY_EMAIL_LIMIT', 10_000_000)
```
AntoineAugusti, line 113: Could you cast to integers here? Otherwise, we'll need to take care of that in multiple places

app/main/views/service_settings.py
```diff
@@ -82,11 +82,11 @@
...
 def service_settings(service_id: str):
 
-    # FIXME: the formatting for annual email limits and translation within service-settings.html can be improved
     limits = {
-        'annual_email': current_app.config["ANNUAL_EMAIL_LIMIT"],
-        'annual_sms': current_app.config["ANNUAL_SMS_LIMIT"]
+        'free_yearly_email': int(current_app.config["FREE_YEARLY_EMAIL_LIMIT"]),
+        'free_yearly_sms': int(current_app.config["FREE_YEARLY_SMS_LIMIT"])
     }
+    assert limits['free_yearly_email'] >= 2_000_000, 'The user-interface does not support french translations of < 2M'
```
AntoineAugusti, line 89: I think we usually raise exceptions instead of using `assert`

Nitpick `*F*rench translations`

QuanNguyen-CDS: Sorry, I don't quite understand this comment unfortunately. The assert will raise an AssertionError exception. Was there a specific `raise Exception` you wanted to do?

AntoineAugusti: Forgot that `assert` raised this exception, this is fine. Intuitively I would have raised `ValueError` or `NotImplementedError` but the code is fine. Pick what you want! 🙏

app/config.py
QuanNguyen-CDS: sounds good

AntoineAugusti (approved):
Thanks for your great PR @QuanNguyen-CDS!

PR #928, opened 2021-03-08
Missing French translation for alt text on homepage

anikbrazeau:
**Top banner image:**

alt="Illustration d'un ordinateur portable dont l'écran affiche une fenêtre avec un message et d'un téléphone cellulaire affichant un message et une bulle de texte."

> Do we want to fix the non-standard hardcoded localisations in the file using `if.. else.. endif` statements? See Trello for more details: https://trello.com/c/ByGjBswg/346-translate-alternative-text-on-images-in-the-signed-out-view-for-accessibility

### DIFF
--- a/app/templates/views/signedout.html
+++ b/app/templates/views/signedout.html
@@ -76,12 +76,21 @@
               </div>
             </div>
             <div class="pt-6 pb-6 ml-10 image-content hidden lg:block lg:visible">
+              {% if session["userlang"] == "en" %}
               <img
                 loading="lazy"
                 alt="An illustration of a laptop with a message window on the screen and a cellphone with a message and a callout blurb."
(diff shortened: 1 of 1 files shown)


anikbrazeau:
[Translate alternative text on images in the signed out view for accessibility](https://trello.com/c/ByGjBswg/346-translate-alternative-text-on-images-in-the-signed-out-view-for-accessibility)

AntoineAugusti (approved):

AntoineAugusti:
@anikbrazeau

> Do we want to fix the non-standard hardcoded localisations in the file using if.. else.. endif statements?

Would be good at some point. The problem is that sometimes image paths are different in FR/EN and sometimes they are the same (when this is just an illustration without text). Ideally, we would have a refactor card handling that.

PR #929, opened 2021-03-08
Added 'More' menu item when all items do not fit current resolution

jimleroyer:
# Summary | Résumé

[Trello card](https://trello.com/c/773NHG0t/333-nav-bug-more-plus-menu-option-when-there-are-many-items)

These changes add the *More* menu item on medium and larger viewports.

There are a few fixes accompanying this PR though:

* Partial move of the logic into the Notify Javascript extension mechanism. The Notify front-end has support for a `data-module` HTML attribute that specifies which piece of Javascript module should run on the decorated HTML element. This makes the code easier to test, reuse and maintain. I put all navigation logic into one new `NavMenu` module, but that isn't right because there are lots of unrelated logic (although all for navigation) in this new module. I made it this way for now to get to the next stage while improving just a bit. 
* Some fixes involving `undefined` variables when the JS uglifier did not run. This made the debugging of Javascript files in the local environment harder as we could not get it to run properly without the uglification process. Well that works now, we can debug locally and read the original Javascript files now.  

**tl;dr:** Visit the homepage screens.

At the moment, only the non-logged-in homepage screens **in French** triggers the *More* menu item to appear (that I know of). Smaller resolutions will trigger the smaller viewports before the *More* menu item appears; these don't have the horizontal menu items so the *More* menu item is unnecessary. 

# Design // SS

The following SS comes from [the figma design](https://www.figma.com/file/r3GzRBjBDWvL2AwCWq1e1W/Navigating-the-site-%7C-Naviguer-le-site?node-id=5724%3A967):

image

The following screenshot comes from the actual implementation:

image

I was not sure of the *More* menu item behaviour so I made it an overlay menu for now. These are the differences:

* It should push down the content, it should create a white space between the menu and content for every menu items displayed vertically. 
* The alignment with the `More` menu item and its sub-elements is a bit off. This is hard to align in an exact similar manner. I can see a way to do it but with more work. 


app/assets/javascripts/navMenu.js
```diff
@@ -0,0 +1,183 @@
+(function(Modules) {
```
jimleroyer, line 1: This Notify JS module can be used with `data-attribute="nav-menu"` on an HTML element to trigger the Javascript below.

app/assets/javascripts/navMenu.js
```diff
@@ -0,0 +1,183 @@
...
+      moreMenuItems.children().removeClass("text-center");
+      moreMenuItems.children().addClass("text-right");
+    } else {
+      // Hide the More menu when it does not contain item(s).
+      moreMenu.addClass("hidden");
+    }
+  }
+
+  Modules.NavMenu = function() {
+    this.start = function(component) {
```
jimleroyer, line 178: The `component` parameter is the HTML element decorated with the `data-attribute="nav-menu"` attribute.

jimleroyer: At the moment, it is not ideal because this module deals with 4 different menus (i.e. desktop nav menu, mobile nav menu, account desktop menu & account mobile menu). There could be one module for each of these, isolated from each other, along with potential generic modules, for example: scoping tabbing navigation for accessibility purposes (a simple `data-attribute` on an HTML element would add the functionality on any other element).

app/templates/admin_template.html
```diff
@@ -22,8 +22,3 @@
     </div>
   </div>
 {% endblock %}
-
-{% block body_end %}
-  <script src="{{ asset_path }}javascripts/main.min.js"></script>
-  <script type="text/javascript" src="{{ asset_url('javascripts/all.min.js') }}"></script>
-{% endblock %}
```
jimleroyer, line 29: The Javascript files were exclusive to the logged-in admin template before. I moved it to be available on other pages now.

app/templates/main_template.html
```diff
@@ -152,6 +152,9 @@
     )
 </script>
 
-{% block page_script %}{% endblock %}
+{% block page_script %}
+<script src="{{ asset_path }}javascripts/main.min.js"></script>
+<script type="text/javascript" src="{{ asset_url('javascripts/all.min.js') }}"></script>
+{% endblock %}
```
jimleroyer, line 158: Notice the override of the `page_script` template block. We did not use this intended block before.

app/templates/notify_template.html
```diff
@@ -22,5 +22,7 @@
   </head>
   <body>
     {% block main_content %}{% endblock %}
+    <script src="{{ asset_path }}javascripts/main.min.js"></script>
+    <script type="text/javascript" src="{{ asset_url('javascripts/all.min.js') }}"></script>
```
jimleroyer, line 26: This adds the Javascript files to the homepage as well.

app/templates/partials/nav/gc_header_nav.html
```diff
@@ -12,7 +12,7 @@
     <!-- Mobile menu and nav bar -->
     <div class="grid grid-cols-2">
         <div class="py-2">
-          <a href="#" id="menu"
+          <a href="#" id="menu" data-module="nav-menu"
```
jimleroyer, line 15: The Notify js extension mechanism is triggered with this attribute.

app/templates/partials/nav/gc_header_nav.html
```diff
@@ -102,106 +102,27 @@
...
-
-  safeAddEventListener("#account-icon-mobile", "click", accountMobileToggle)
-  safeAddEventListener("#account-close-mobile", "click", accountMobileToggle)
-  registerMenuEscape("#account-menu-mobile", accountMobileToggle)
-
-  safeAddEventListener("#account-menu", "click", accountMenuToggle)
-  registerMenuEscape("#account-menu-options", accountMenuToggle)
-  registerTabNavigationScope('#account-menu-mobile')
-
-</script>
```
jimleroyer, line 207: All this Javascript will now get into the regular JS build pipeline and aligned with Notify Javascript way of doing.

app/translations/csv/fr.csv
```diff
@@ -1300,3 +1300,4 @@
 "Daily sending limit","Limite d’envoi quotidienne"
 "{} notifications","{} notifications"
 "{} calls","{} appels"
+"More","Plus"
```
jimleroyer, line 1303: Do we really need this? I heard before that *Less is more*. 🤔 💭

app/assets/javascripts/index.js
```diff
@@ -8,11 +8,22 @@ import { App } from "./scheduler/App";
...
 
+if (!window.APP_PHRASES || typeof APP_PHRASES === "undefined") {
+  window.APP_PHRASES = {
+    now: "Now"
+  };
+}
+
+if (!window.APP_LANG || typeof APP_LANG === "undefined") {
+  window.APP_LANG = "en";
+}
```
AntoineAugusti, line 19: Why is this needed? Is this code not enough https://github.com/cds-snc/notification-admin/blob/dfe53538dd49f40f009162793578de2bd347450e/app/templates/main_template.html#L40-L41?

app/assets/javascripts/navMenu.js
AntoineAugusti: Can you point out what's new in this file? I understand you moved it here from the HTML template. This would help understand what to pay attention to.

app/assets/javascripts/navMenu.js
```diff
@@ -0,0 +1,183 @@
...
+    safeAddEventListener("#menu", "click", menuMobileToggle)
+    registerMenuEscape(".mobile-nav", menuMobileToggle)
+
+    safeAddEventListener("#account-icon-mobile", "click", accountMobileToggle)
+    safeAddEventListener("#account-close-mobile", "click", accountMobileToggle)
+    registerMenuEscape("#account-menu-mobile", accountMobileToggle)
+
+    safeAddEventListener("#account-menu", "click", accountMenuToggle)
+    registerMenuEscape("#account-menu-options", accountMenuToggle)
+    registerTabNavigationScope('#account-menu-mobile')
```
AntoineAugusti, line 107: I've seen teams prefixing HTML IDs used only for JS purposes with `js-` to make sure it's clear that it's used for this purpose and it could break if deleted. WDYT?

app/templates/main_template.html
```diff
@@ -152,6 +152,9 @@
     )
 </script>
 
-{% block page_script %}{% endblock %}
+{% block page_script %}
+<script src="{{ asset_path }}javascripts/main.min.js"></script>
```
AntoineAugusti, line 156: Could use `asset_url` here to leverage fingerprinting?

app/translations/csv/fr.csv
AntoineAugusti: I think it's helpful to highlight that there is indeed _more_. Like adding "Menu" on top of the "hamburger icon" to be explicit.

app/templates/notify_template.html
```diff
@@ -22,5 +22,7 @@
   </head>
   <body>
     {% block main_content %}{% endblock %}
+    <script src="{{ asset_path }}javascripts/main.min.js"></script>
```
AntoineAugusti, line 25: Could use `asset_url` here to leverage fingerprinting?

app/assets/javascripts/navMenu.js
jimleroyer: I agree, sounds like a good practice. With my upcoming PR, we should not have these explicitly referred IDs though.

app/templates/main_template.html
jimleroyer: Sounds good, let me try if that does not give some trouble before.

app/assets/javascripts/index.js
jimleroyer: It produces an error if ran locally and without the uglification process. I saw that same logic in other files as well and it is my assumption at the moment that there is no priority ordering of js module execution guaranteed. Hence, many similar code excerpts can be found to assume that previous code wasn't ran. I would validate this theory and consider this a potential tech debt.

app/assets/javascripts/navMenu.js
jimleroyer: Yes let me add more comments to say what is new.

app/assets/javascripts/navMenu.js
```diff
@@ -0,0 +1,183 @@
...
+
+  function accountMenuToggle() {
+    const accountMenu = document.querySelector("#account-menu-options")
+    const arrow = document.querySelector(".account-menu-arrow")
+    accountMenu.classList.toggle("hidden")
+    arrow.classList.toggle("flip")
+    accountMenu.querySelector('a').focus()
+  }
+
+  function init($menu) {
```
jimleroyer, line 97: This function is new.

app/assets/javascripts/navMenu.js
```diff
@@ -0,0 +1,183 @@
...
+      ctx.varWidth += $(this).outerWidth();
+    });
+
+    resizeMenu(ctx); 
+    $(window).resize(function() {
+      resizeMenu(ctx);
+    });
+  }
+
+  function resizeMenu(ctx) {
```
jimleroyer, line 137: This function is new.

app/assets/javascripts/navMenu.js
```diff
@@ -0,0 +1,183 @@
...
+      moreMenu.css({"display": moreMenuDisplay, "width": moreMenuWidth});
+      moreMenuItems.children().removeClass("text-center");
+      moreMenuItems.children().addClass("text-right");
+    } else {
+      // Hide the More menu when it does not contain item(s).
+      moreMenu.addClass("hidden");
+    }
+  }
+
+  Modules.NavMenu = function() {
```
jimleroyer, line 177: The module setup is new.

app/assets/javascripts/mobileMenuAccount.js
```diff
@@ -0,0 +1,64 @@
+(function (Modules) {
+  "use strict";
+
+  function registerMenuEscape($selector, toggleFn) {
```
QuanNguyen-CDS, line 4: I see this code also being in menu.js, can we combine them (following the DRY principles)?

app/assets/javascripts/moreMenu.js
```diff
@@ -0,0 +1,94 @@
...
+    });
+  }
+
+  function resizeMenu() {
+    // Current available width of the menu.
+    elemWidth = $menuItems.width();
+
+    // Calculate how many items fit on the total width.
+    // Substract 1 as CSS :gt property is 0-based index.
+    fitCount = Math.floor((elemWidth / varWidth) * ctr) - 1;
```
QuanNguyen-CDS, line 47: could varWidth be zero here (ever?) - could we run into a div-by-zero scenario?

app/assets/javascripts/scheduler/store.js
```diff
@@ -59,6 +59,7 @@ export const defaultState = (
     updateMessage: "",
     _24hr: LOCALE === "en" ? "off" : "on",
     errors: "",
+    // REVIEW: Raises error past 23h00m in a day, not great when debugging late.
```
QuanNguyen-CDS, line 62: 👍

app/assets/javascripts/scopeTabNavigation.js
```diff
@@ -0,0 +1,62 @@
...
+   * other elements of the page as long as the container is not reached. This
+   * works well within a modal popup that took focus away from the main window.
+   */
+  function registerTabNavigationScope(modal) {
+    const firstFocusableElement = modal.querySelectorAll(focusableElements)[0];
+    const focusableContent = modal.querySelectorAll(focusableElements);
+    const lastFocusableElement = focusableContent[focusableContent.length - 1];
+
+    document.addEventListener("keydown", function (e) {
+      let isTabPressed = e.key === "Tab" || e.keyCode === 9;
```
QuanNguyen-CDS, line 27: From my understanding the use of keyCode is no longer recommended (https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode)? From the Mozilla reference, the recommendation is to use `KeyboardEvent.code` however it also states that some browser still do not have support for this either...

Not sure if there's a better alternative.

app/templates/partials/nav/gc_header_nav.html
```diff
@@ -102,106 +106,30 @@
...
+          data-module-menu-items="proposition-links" 
+          data-module-menu-container="more-menu-container"
+          class="flex flex-shrink-0 text-center p-5 bg-gray hidden">
+          <a
+            href="#"
+            aria-hidden="true"
+            class="text-smaller text-blue visited:text-blue link:text-blue font-bold no-underline focus:outline-none">
+            {{ _('More') }}
+            <img
+              alt="Account menu expansion arrow"
```
QuanNguyen-CDS, line 121: question: is this page language aware? do we need to consider `alt` in both languages [French, English]?

anikbrazeau:
This is starting to take shape! Encountered it a few times in signed in (with platform admin) as well 😄 

Wondering how much work it'd be to _not_ push down the content instead, when the `More` menu opens? Worried it might not be the behaviour users expect on desktop.

app/assets/javascripts/menu.js
```diff
@@ -0,0 +1,65 @@
+(function (Modules) {
+  "use strict";
+
+  function registerMenuEscape($selector, toggleFn) {
+    $selector.keydown(function (e) {
```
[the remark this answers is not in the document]
jimleroyer: It should trigger on key down and will only happen once.

app/assets/javascripts/mobileMenuAccount.js
jimleroyer: Let me try to put this in a utils file.

My initial hope was to have the escape key registration as a reusable module that can stand on its own... but current module support can't let me define several of these on the same HTML element.

app/assets/javascripts/moreMenu.js
jimleroyer: If there are no children, i.e. menu items. This should not happen.. but I can add an extra safe precaution.

app/assets/javascripts/scopeTabNavigation.js
jimleroyer: Yeah, it's not supported by IE11, which we have to support AFAIK, although I'd be more than happy to drop this hopefully soon. Using polyfills could be a workaround such [as this one](https://www.npmjs.com/package/keyboardevent-key-polyfill). We could have an API front-end of our own for these.

app/templates/partials/nav/gc_header_nav.html
jimleroyer: Yes we need to. 😓 

Let me handle this in a separate PR though. I would like designer's review on our hardcoded alt images to know if these are proper description as they are.

jimleroyer:
There is now a) a divider between items, b) a vertical sideline on selected page, c) removal of the horizontal line on selected page, d) the menu overlay on top of the admin section now, e) coding related PR comments have been addressed too.

There have been no adjustment in spacing/alignment.

jimleroyer:
[NAV Bug: "More" / "Plus" menu option when there are many items](https://trello.com/c/773NHG0t/333-nav-bug-more-plus-menu-option-when-there-are-many-items)

app/templates/notify_template.html
```diff
@@ -22,5 +22,8 @@
...
```
AntoineAugusti, line 25:
```suggestion
    <script src="{{ asset_path }}javascripts/main.min.js"></script>
```
You already include this file afterwards, I think you can delete that

AntoineAugusti (approved):
LGTM, a quick comment to address before merging 🚢

app/templates/notify_template.html
jimleroyer: My mistake, I'll remove.

PR #930, opened 2021-03-08
Tweak platform admin menu

anikbrazeau:
Make platform admin labels clearer and shorter, reorder to bring organisations up.

-- Search for service rather than Find service by name
-- Search for user rather than Find user by email
-- Organisations


AntoineAugusti (approved):
Thanks!

PR #933, opened 2021-03-09
fix add swap img to switch services link anchor tag

dorianjp:
Trello
https://trello.com/c/pUzoUtfo/329-nav-bug-include-arrows-icon-in-link

Added img to inline with link for switch services.


anikbrazeau (approved):
Looks great in Heroku

PR #935, opened 2021-03-09
feat: enable customizing message limit when turning live

AntoineAugusti:
## Context

At the moment, daily limits are 50 messages/day if you're in trial mode and 10k per day when turning a service live. With the new "Request to go live" process, clients share with us their expected volume. We want to assign them limits fitting their need, no need to give them 10k messages/day if they plan to send less than 1k notifications/month.

## Problem
When turning a service live, the limit is always set to 10k, even if you picked another number.

We should be able to change the limit before turning a service live and then turning it live, without overriding the custom value. Service owners will receive an email that their service is now live with the daily limit.

## Done when
We can change the message limit when a service is in trial mode, turn it live and then keep the custom value

Card: https://trello.com/c/vS357Jw7/364-be-able-to-customize-message-limit-before-turning-live

---
## How to test

Use the review app, create a trial service, change the message limit, turn the service live.


app/main/views/service_settings.py
```diff
@@ -339,7 +339,15 @@ def service_switch_live(service_id):
     )
 
     if form.validate_on_submit():
-        current_service.update_status(live=form.enabled.data)
+        live = form.enabled.data
+        message_limit = current_app.config["DEFAULT_SERVICE_LIMIT"]
+        if live:
```
QuanNguyen-CDS, line 344: I'm not sure I understand this logic, when i turn on and off my live status, the value in the database gets set to some value representative of the limit, can we not just set message_limit = current_service.message_limit?

QuanNguyen-CDS:
Maybe a wild edge case, but now that message_limit is settle-'able' (and I guess controlled my us?) do we protect against out of range numbers? (-negative?)

AntoineAugusti:
@QuanNguyen-CDS Message limit is set by platform admins only but there is not a protection in place to prevent negative numbers at the moment. https://github.com/cds-snc/notification-admin/blob/e1325d64d3c12ea8a68f3bfe46ac6b4c501229ff/app/main/forms.py#L727-L733

I'm happy to take care of this in this PR if you want.

QuanNguyen-CDS:
> @QuanNguyen-CDS Message limit is set by platform admins only but there is not protection in place to prevent negative numbers at the moment.
> 
> https://github.com/cds-snc/notification-admin/blob/e1325d64d3c12ea8a68f3bfe46ac6b4c501229ff/app/main/forms.py#L727-L733
> 
> I'm happy to take care of this in this PR if you want.

That'd be awesome if you can put `NumberRange(min=0)` as an additional validator

app/main/views/service_settings.py
QuanNguyen-CDS: sounds good

AntoineAugusti:
> That'd be awesome if you can put `NumberRange(min=0)` as an additional validator

@QuanNguyen-CDS Did just the last in my last commit as suggested 👌

QuanNguyen-CDS (approved):
Thank you, LGTM

PR #943, opened 2021-03-11
fix: display custom sending domain in settings if set

AntoineAugusti:
Use the custom sending domain in service settings, if it set, in the "Sending email address" field in service settings.

At the moment there is a bug where the displayed domain is always the default Notify domain, even if the service uses a custom sending domain.

Fixes https://github.com/cds-snc/notification-admin/commit/366e065e14773d0343f4a346a0a47bd84827515c

### DIFF
--- a/app/templates/views/service-settings.html
+++ b/app/templates/views/service-settings.html
@@ -6,7 +6,7 @@
 {% endblock %}
 
 {% set change_txt = _('Change') %}
-{% set display_email_from = current_service.email_from +'@' + sending_domain %}
+{% set display_email_from = current_service.email_from + '@' + sending_domain %}
 
 {% block maincolumn_content %}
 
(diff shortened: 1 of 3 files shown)


app/templates/views/service-settings.html
```diff
@@ -410,8 +410,7 @@ <h2 class="heading-medium">{{ _('Platform admin settings') }}</h2>
           {% call row() %}
           {% set txt = _('Sending domain') %}
           {{ text_field(txt) }}
-          {% set txt = _('{}') %}
-          {{ text_field(txt.format(current_service.sending_domain)|safe if current_service.sending_domain else sending_domain) }}
+          {{ text_field(sending_domain) }}
```
jimleroyer, line 413: I assume we always expect a sending domain to be configured? Is there a check in configuration to make this required?

AntoineAugusti: Most of the time it will be set to `null`. The new code in service settings takes care of the fallback value

```py
sending_domain=current_service.sending_domain or current_app.config["SENDING_DOMAIN"]
```

jimleroyer: Discussed with Antoine on this topic, there is a fallback default value so we should be good.

jimleroyer (approved):
🇸🇪 👨‍🍳

PR #949, opened 2021-03-16
Add conversion method for uploaded CSV files

RenDavidDimen:
# Summary | Résumé

Added method for attempting to convert non-utf8 files to a utf8 encoding.

Uploading a file should now try to deduce what encoding the file is currently in a try to convert it to utf8. Should the encoding not be certain, an error is thrown.

----


app/utils.py
```diff
@@ -319,6 +320,7 @@ def get_extension(filename):
 
     @staticmethod
     def normalise_newlines(file_content):
+        file_content.stream.seek(0)
```
AntoineAugusti, line 323: Wondering why this is needed

app/utils.py
```diff
@@ -339,6 +341,9 @@ def from_file(cls, file_content, filename=''):
         extension = cls.get_extension(filename)
 
         if extension == 'csv':
+            utf8_format = convert_to_utf8(file_content.read())
+            file_content.stream.seek(0)
```
AntoineAugusti, line 345: Wondering why this is needed

app/utils.py
```diff
@@ -370,6 +375,41 @@ def as_excel_file(self):
...
 
+def convert_to_utf8(file_data):
+    # Detect File Encoding
+    encoding_result = chardet.detect(file_data)
+
+    if file_data == b'':
+        return file_data
+
+    # If file encoding cannot be determined
+    if encoding_result['encoding'] is None:
```
AntoineAugusti, line 386: Can put the low confidence part here as well, to avoid duplicating code when raising the exception

tests/app/test_utils.py
```diff
@@ -162,6 +163,24 @@ def test_spreadsheet_checks_for_bad_arguments(args, kwargs):
     assert str(exception.value) == 'Spreadsheet must be created from either rows or CSV data'
 
 
+def test_convert_to_utf8():
+    file_content = b'\xe9, \xe0'
+    assert convert_to_utf8(file_content) == b'\xc3\xa9, \xc3\xa0'
```
AntoineAugusti, line 168: Can you put comments in here to help others understand what's happening?

AntoineAugusti (commented):
Looks good, thanks for working on that 👏 

Wondering if you tested with the buggy ASCII file I sent you?

app/utils.py
RenDavidDimen: It seems that when we read from the temporary file, we are reading from a stream so reading the file again without resetting where we are in the stream returns an empty file. I wasn't able to find a solid source for the documentation, however you can see it being implemented in their examples here: https://docs.python.org/3/library/tempfile.html#examples

Because of this, if we have read the file content and written to the stream, it would be safest to ensure we are back at the start of the stream when reading here

app/utils.py
RenDavidDimen: Similarly to the other comment. Reading the stream would have us essentially at the end of the byte string, so in this case we're re-writing over the temporary file with the new utf8 data

RenDavidDimen:
> Looks good, thanks for working on that 👏
> 
> Wondering if you tested with the buggy ASCII file I sent you?

I was able to test with the file in the Trello card and it seems to work as intended 👍

AntoineAugusti (approved):

PR #968, opened 2021-03-24
fix: include base JS files when editing email template

AntoineAugusti:
Variables stopped to be highlighted when editing email templates in staging/production. This is a UI regression.

Screen Shot 2021-03-24 at 14 53 56

The problem was caused by the fact that the base JS files were not included, only on this specific template. The `page_script` block is overwritten and those 2 JS files are not loaded, the [highlightTags](https://github.com/cds-snc/notification-admin/blob/master/app/assets/javascripts/highlightTags.js) module was not loaded as a consequence.

This was introduced in https://github.com/cds-snc/notification-admin/pull/929 when the `page_script` block was moved to the main template. We missed the only place in the codebase were this block is used unfortunately.

My fix is to call `super()` to include the parent content and to load the 2 JS files.

Searched for other occurrences of this bug in the codebase and found none. 

## Test

Use the review app, edit an email template with variables, see that those are highlighted.


jimleroyer (approved):
LGTM / Thanks for fixing a bug I introduced.

AntoineAugusti:
> Thanks for fixing a bug I introduced.

That's not true, it's a collective mistake. It slipped through a developer/the review and tests so blame the process, not the person. 👍

PR #970, opened 2021-03-26
Clarify branding/language choice in service onboarding

anikbrazeau:
# Summary | Résumé

Update logo choice in onboarding screen to be about language instead to clarify what the branding choice is about.

Currently, in onboarding, users have to choose a logo (EN or FR) for their service.
This causes confusion, as it's not clear what the differences are. From research: “They look the same to me… Is this a trick question?”

**Content changes:**
- change heading to be more explicit about what choice means
- add string to explain impact of choice on default email branding
- change radio button options to describe the difference
- update reassurance below button to match

**Before**

Screen Shot 2021-03-29 at 4 15 53 PM
Screen Shot 2021-03-29 at 4 16 07 PM

**After**
Screen Shot 2021-03-29 at 4 13 36 PM
Screen Shot 2021-03-29 at 4 14 16 PM


anikbrazeau:
[[Trial Content][Design] Copy updates to clarify trial mode](https://trello.com/c/N5fK1VMH/390-trial-contentdesign-copy-updates-to-clarify-trial-mode)

app/main/views/add_service.py
```diff
@@ -35,7 +35,7 @@
 DEFAULT_STEP = STEP_NAME
 
 STEP_NAME_HEADER: str = _("Name your service")
-STEP_LOGO_HEADER: str = _("Choose a logo for your service")
+STEP_LOGO_HEADER: str = _("Choose the default language of your service")
```
AntoineAugusti, line 38: Do we want to put this here? In the past they where only changing the default logo. We don't have the concept of "default language" for a service. For example we don't reflect this when sending emails by setting a different `lang` attribute for example. Do we want to go this route after?

AntoineAugusti (approved):
Changes look okay, just a clarification

app/main/views/add_service.py
AntoineAugusti: The `lang` attribute on emails is always set to `en` at the moment for example https://github.com/cds-snc/notification-utils/blob/99213ad675eb3f1efb006dc4e5b7d8c25da69329/notifications_utils/jinja_templates/email_template.jinja2#L3

anikbrazeau: @AntoineAugusti, curious, is it possible to select a lang attribute based on this choice? 
would there be a way to put `en` language attribute for the Rnglish part and `fr` language attribute for the French part of a single template accordingly? 🤔 

We can maybe dig into this more next sprint with template work too.  👀

PR #972, opened 2021-03-30
feat: add a11y tracking to staging env

maxneuvians:
This PR adds a call to the a11y tracking tools into the CI pipeline. After the admin component has been deployed it will make an API call to the a11y tracking tools API that request a spidering of the public facing pages of notify's staging environment.

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -68,5 +68,5 @@ jobs:
         run: |
           curl -s https://api.a11y.cdssandbox.xyz/v1/reports\?recent\=true > /dev/null
           sleep 60
-          json='{"product": "cds-snc/notification", "revision": "${{github.sha}}", "urls":["https://staging.notification.cdssandbox.xyz"], "spider": 1}'
+          json='{"product": "cds-snc/notification", "revision": "${{github.sha}}", "urls":["https://staging.notification.cdssandbox.xyz"], "spider": 1, "ci": 1}'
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yaml
```diff
@@ -64,3 +64,9 @@ jobs:
     - uses: cds-snc/notification-pr-bot@master
       env:
         TOKEN: ${{ secrets.TOKEN }}
+    - name: Run a11y tracker
+        run: |
+          curl -s https://api.a11y.cdssandbox.xyz/v1/reports\?recent\=true > /dev/null
+          sleep 60
+          json='{"product": "cds-snc/notification", "revision": "${{github.sha}}", "urls":["https://staging.notification.cdssandbox.xyz"], "spider": 1}'
```
AntoineAugusti, line 71: I'm assuming `spider` is a depth? It's documented as a boolean for now https://github.com/cds-snc/a11y-tools#submitting-data

Any way to prevent external URLs from being crawled?

AntoineAugusti (approved):

.github/workflows/docker.yaml
AntoineAugusti: Could put `ci` to `1` after looking at your Lambda

maxneuvians: Spider is just a bool if you should spider. Spider depth is currently defined at 2 - https://github.com/cds-snc/a11y-tools/blob/main/chromium-lambda/src/index.ts#L44

It will only scan your domain and filter external links
https://github.com/cds-snc/a11y-tools/blob/main/chromium-lambda/src/index.ts#L185

PR #977, opened 2021-04-08
Revert "Add conversion method for uploaded CSV files"

RenDavidDimen:
Reverts cds-snc/notification-admin#949

Found `chardet` may identify UTF-8 files as non UTF-8 files and attempt to convert them causing issues where non should exist. Seeing as this does more harm than good, we will be reverting these changes

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -11,7 +11,6 @@ from itertools import chain
 from os import path
 
 import boto3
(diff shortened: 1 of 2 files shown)


AntoineAugusti (approved):
Can merge as soon as tests pass

PR #980, opened 2021-04-09
feat: remove guidance on how to send files on templates page

AntoineAugusti:


anikbrazeau (approved):
So fast! Thanks for doing this!

PR #981, opened 2021-04-09
fix: display custom sending domain on email from page

AntoineAugusti:
The default sending domain was always displayed on this page, even when a service has a custom sending domain set.

image

Did a similar fix on service settings recently https://github.com/cds-snc/notification-admin/pull/943


tests/app/main/views/test_service_settings.py
```diff
@@ -662,6 +663,28 @@ def test_should_redirect_after_service_name_confirmation(
     assert mock_verify_password.called is True
 
 
+@pytest.mark.parametrize('sending_domain', [None, 'test.example.com'])
+def test_service_email_from_change(
```
AntoineAugusti, line 667: Was not covered by tests before 😢

jimleroyer (approved):

PR #994, opened 2021-04-16
Automated a11y test on authenticated dashboard page

jimleroyer:
# Summary | Résumé

Added automated a11y test for the authenticated service dashboard page.

Run the test suite! 🏃 

      langues officielles?
      fonctionnalité existante?


maxneuvians:
https://a11y-tools-query.herokuapp.com/revisions/cds-snc%2Fnotification/078d2045d24aa157cd2ce82397312ed3a456f9b1

tests/app/main/views/test_dashboard.py
```diff
@@ -1152,3 +1153,23 @@ def test_get_free_paid_breakdown_for_billable_units(now, expected_number_of_mont
...
+    mocker.patch(
+        'app.template_statistics_client.get_template_statistics_for_service',
+        return_value=copy.deepcopy(stub_template_stats)
+    )
+
+    url = url_for('main.service_dashboard', service_id=SERVICE_ONE_ID)
+    response = logged_in_client.get(url)
+
+    assert response.status_code == 200
+    a11y_test(url, response.data.decode('utf-8'))
```
AntoineAugusti, line 1175: Could you pass the full URL? We have a base and this page is without a base, see https://a11y-tools-query.herokuapp.com/revisions/cds-snc%2Fnotification/078d2045d24aa157cd2ce82397312ed3a456f9b1

You can tweak the function to ensure a full URL is passed or preprend the admin base if you want.

tests/app/main/views/test_dashboard.py
```diff
@@ -1152,3 +1153,23 @@ def test_get_free_paid_breakdown_for_billable_units(now, expected_number_of_mont
             {'free': 0, 'name': 'February', 'paid': 2000, 'letter_total': 0, 'letters': [], 'letter_cumulative': 0},
             {'free': 0, 'name': 'March', 'paid': 0, 'letter_total': 0, 'letters': [], 'letter_cumulative': 0}
         ][:expected_number_of_months]
+
+
+def test_dashboard_page_a11y(
```
AntoineAugusti, line 1158: Could you call `a11y_test` in an existing test and not create a test case just to pass a HTML payload to the a11y tester? I believe other tests in this file already do similar things.

AntoineAugusti (changes requested):

tests/app/main/views/test_dashboard.py
jimleroyer: It could be a matter of style: I personally prefer to have one focused test per concern for a variety of reasons. It makes it clear we are testing the a11y output so no future maintainer will delete, or mark @ignored a changes containing this, or just modifying it in a way where the a11y is affected. It is a rule of thumb for me, there are cases I could deviate from this. 

What do you think? What do you consider the limit (in lines of code, or tested concerns) of what a unit-test should contain?

AntoineAugusti: Agreed that's something we can debate. Our current test usually contain various assertions including mocks, asserting database states, inspecting the page's content etc. I think that it's best to keep continue doing as is. Our tests suite could go or become slow if we duplicate various scenarios to test for various things like page content, database, mock calls, a11y.

You could add a [pytest custom marker](https://docs.pytest.org/en/stable/example/markers.html) though! `@pytest.mark.a11y` Be sure to look at how to register those.

tests/app/main/views/test_dashboard.py
jimleroyer: I've made use of the [`_external` parameter of the `url_for` method](https://flask.palletsprojects.com/en/1.1.x/api/#flask.url_for) to get the full URL.

maxneuvians:
https://a11y-tools-query.herokuapp.com/revisions/cds-snc%2Fnotification/7c43600c592a4e1fc3a96f26a325514af50464a9

tests/app/main/views/test_dashboard.py
jimleroyer: The `@pytest.mark.a11y` is definitely a good compromise.

To expand a bit on the topic, I understand that longer tests might be acceptable when heavy performance resources are involved, especially for integration tests, as unit-tests should not be heavy and very quick to execute. The tests for accessibility in general can benefit from mocked calls that do not involve external calls to database and network bur rather be CPU intensive and hence, quick to execute.

To my personal taste, I can see that some existing tests are quite long sometimes and harder to understand. It's possible it is hard to avoid. It's something that nodejs test libraries can do better IMO because of nested tests within nested resources loading closures, but not easy using pytest here.

We can chat offline tomorrow morning if you want. Ultimately, I offer my perspective but I'll let you decide on the style you want to see in the project.

AntoineAugusti (approved):
LGTM :shipit:

maxneuvians:
https://a11y-tools-query.herokuapp.com/revisions/cds-snc%2Fnotification/fe6a24456c0b0494c5393ffa1a19ad5fe685fb37

maxneuvians:
https://a11y-tools-query.herokuapp.com/revisions/cds-snc%2Fnotification/25710ef09d2cf0de5a9904c7ce76acb8657dcf4e

PR #1013, opened 2021-05-03
fix: error style when uploading spreadsheets

AntoineAugusti:
When uploading a CSV with errors in multiple rows, the vertical alignment of the table skews and becomes difficult to read.

Steps to reproduce
-------------------
- Create a CSV file with missing values in multiple rows
- Head to the Service > Templates > Select Template > Send > Upload a list of email addresses
- Upload CSV file that contains the missing values
- Observe errors on new page once loaded

Actual Outcome
-------------------
- Table rows are skewed

Expected Outcome
-------------------
- Table values should all be aligned properly

# New style
Screen Shot 2021-05-01 at 09 48 29

Previously we only had a border on the left side, matching the error style elsewhere in the application, for example when you don't fill required fields.

https://github.com/cds-snc/notification-admin/blob/a47ee76fa3b2752985907ed3b36c5f714d37f867/app/assets/stylesheets/components/table.scss#L100

Trello: https://trello.com/c/g10I1EUr/469-bug-table-alignment-is-off-when-uploading-a-recipients-csv-file-with-multiple-errors


willeybryan (approved):
LGTM

PR #1014, opened 2021-05-03
feat: force SMS 2FA users to log in with email every month

AntoineAugusti:
Some SMS 2FA users use a personal phone number to use Notify. We have no way to remove their access, compared to email addresses, when they leave the Government of Canada or move to another department.

Make sure that SMS 2FA users log in at least once every 30 days with an email 2FA code, proving that they still have access to their inbox.

Trello: https://trello.com/c/l2sS1gRP/487-make-sure-sms-2fa-users-log-in-with-email-every-month

### DIFF
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -161,7 +161,7 @@ class User(JSONModel, UserMixin):
 
     @property
     def email_auth(self):
-        return self.auth_type == 'email_auth' or not self.has_recent_email_login()
+        return self.auth_type == 'email_auth' or self.requires_email_login
 
     @property
     def requires_email_login(self):
--- a/app/models/user.py
+++ b/app/models/user.py
@@ -163,6 +163,10 @@ class User(JSONModel, UserMixin):
     def email_auth(self):
         return self.auth_type == 'email_auth' or not self.has_recent_email_login()
 
+    @property
+    def requires_email_login(self):
+        return self.auth_type == 'sms_auth' and not self.has_recent_email_login()
+
     def has_recent_email_login(self):
         date = user_api_client.get_last_email_login_datetime(self.id)
         return date and date >= (datetime.utcnow() - timedelta(days=30))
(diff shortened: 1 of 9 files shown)


app/models/user.py
```diff
@@ -147,18 +149,23 @@ def sign_in(self):
...
+        return self.auth_type == 'sms_auth' and self.has_recent_email_login()
 
     @property
     def email_auth(self):
-        return self.auth_type == 'email_auth'
+        return self.auth_type == 'email_auth' or not self.has_recent_email_login()
+
+    def has_recent_email_login(self):
+        date = user_api_client.get_last_email_login_datetime(self.id)
+        return date and date >= (datetime.utcnow() - timedelta(days=30))
```
QuanNguyen-CDS, line 168: sorry, i'm not quite sure I understand this comparison, if you have a redis cache TTL of 30 days why do we still have to check the date.
Shouldn't it be enough to rely on the redis cache TTL. So if it's in redis then user has logged in within 30 days. If it's not then it's because the cache key has been flushed
So this will always be true this comparison `date >= (datetime.utcnow() - timedelta(days=30))`

QuanNguyen-CDS: I guess we do this so we can unit test and change the time manually?

AntoineAugusti: You're right, it's superfluous at this point. A non null value is enough but I feel like this made the code easier to understand here and also easier to write tests.

willeybryan (commented):
"une adresse e-mail gouvernementale valide."

The GC usually refers to Email as 'courriel'

AntoineAugusti:
@willeybryan Fixed it in the next commit, thanks for spotting this! https://github.com/cds-snc/notification-admin/pull/1014/commits/56af7332de4a39d506925abbbeddb8afa4a24a9f

willeybryan (approved):

PR #1016, opened 2021-05-03
fix: rework logic for email 2FA logic once a month

AntoineAugusti:
Follow up of #1014 #1015

This time I'm not relying on the test suite mocking everything to know how it works, I actually tested it multiple times locally on top of changing the logic.

The main problem was that the logic tried to send 2 codes here

https://github.com/cds-snc/notification-admin/blob/048be5a6cf34a608d7d93ff8516a69c6b866c1f5/app/models/user.py#L150-L154

`self.email_auth` is false the first time because we didn't log in recently with an email code but the execution doesn't stop here! `self.sms_auth` is now `true` because we just registered a recent email login.

There was the same problem on the email 2FA page, displaying another message: the state was false before the redirection but is true after the redirection because a code has been sent in the meantime.

### DIFF
--- a/app/main/views/sign_in.py
+++ b/app/main/views/sign_in.py
@@ -56,11 +56,13 @@ def sign_in():
                 abort(403)
             else:
                 invited_user.accept_invite()
+        requires_email_login = user and user.requires_email_login
         if user and user.sign_in():
-            if user.sms_auth:
+            if user.sms_auth and not requires_email_login:
                 return redirect(url_for('.two_factor_sms_sent', next=request.args.get('next')))
-            if user.email_auth:
-                return redirect(url_for('.two_factor_email_sent'))
+            if user.email_auth or requires_email_login:
+                args = {'requires_email_login': True} if requires_email_login else {}
+                return redirect(url_for('.two_factor_email_sent', **args))
 
         # Vague error message for login in case of user not known, inactive or password not verified
         flash(_("The email address or password you entered is incorrect."))
--- a/tests/app/main/views/test_sign_in.py
+++ b/tests/app/main/views/test_sign_in.py
@@ -148,7 +148,15 @@ def test_process_sms_auth_sign_in_return_email_2fa_template_if_no_recent_login(
     mock_register_email_login,
     last_email_login,
 ):
-    mocker.patch('app.user_api_client.get_last_email_login_datetime', return_value=last_email_login)
+    mock_last_email_login = mocker.patch(
+        'app.user_api_client.get_last_email_login_datetime',
+        side_effect=[
+            last_email_login,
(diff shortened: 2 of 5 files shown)


tests/app/main/views/test_sign_in.py
```diff
@@ -148,7 +148,15 @@ def test_process_sms_auth_sign_in_return_email_2fa_template_if_no_recent_login(
...
 ):
-    mocker.patch('app.user_api_client.get_last_email_login_datetime', return_value=last_email_login)
+    mock_last_email_login = mocker.patch(
+        'app.user_api_client.get_last_email_login_datetime',
+        side_effect=[
+            last_email_login,
+            last_email_login,
+            datetime.utcnow() - timedelta(seconds=10),
+            datetime.utcnow() - timedelta(seconds=10),
+        ]
```
AntoineAugusti, line 158: Mimic the state change happening for real

tests/app/main/views/test_sign_in.py
```diff
@@ -160,16 +168,17 @@ def test_process_sms_auth_sign_in_return_email_2fa_template_if_no_recent_login(
         }
     )
     assert response.status_code == 302
-    assert response.location == url_for('.two_factor_email_sent', _external=True)
+    assert response.location == url_for('.two_factor_email_sent', requires_email_login=True, _external=True)
 
     mock_get_security_keys.assert_called_with(api_user_active['id'])
-    mock_send_verify_code.assert_called_with(api_user_active['id'], 'email', None, None)
+    mock_send_verify_code.assert_called_once_with(api_user_active['id'], 'email', None, None)
```
AntoineAugusti, line 174: `called_once` is important

app/main/views/sign_in.py
```diff
@@ -56,11 +56,13 @@ def sign_in():
                 abort(403)
             else:
                 invited_user.accept_invite()
+        requires_email_login = user and user.requires_email_login
         if user and user.sign_in():
-            if user.sms_auth:
+            if user.sms_auth and not requires_email_login:
```
QuanNguyen-CDS, line 61: can you not use `user.requires_email_login` directly?

app/main/views/sign_in.py
```diff
@@ -56,11 +56,13 @@ def sign_in():
...
                 return redirect(url_for('.two_factor_sms_sent', next=request.args.get('next')))
-            if user.email_auth:
-                return redirect(url_for('.two_factor_email_sent'))
+            if user.email_auth or requires_email_login:
```
QuanNguyen-CDS, line 63: same here

QuanNguyen-CDS (approved):
small comments, not gating

PR #1018, opened 2021-05-04
Fix some content focus states

AntoineAugusti:
Fix several focus states:

- Forgot your password?
- Services under "Your services"
- Templates: Use my email address link
- Templates: Upload a list of email addresses
- Settings: Contact us
- Settings: Delete buttons
- Back link
- Check messages sent per month (show more in CSS)

Screen Shot 2021-05-04 at 14 25 34
Screen Shot 2021-05-04 at 14 25 16
Screen Shot 2021-05-04 at 14 24 55
Screen Shot 2021-05-04 at 14 24 33
Screen Shot 2021-05-05 at 09 49 48


e-pang (dismissed):
Thanks for making these additional updates Antoine! 

A couple of comments.

1. Change the [Delete this service] and [Suspend service] to red destructive buttons
- we can also wait to do this as part of the buttons update. 

2. In the list of services can we update the link style to match other in-context links

Let me know if you have any questions!

AntoineAugusti:
@e-pang as discussed with you on Slack, we will wait to work on buttons do change 1) (this is only visible to platform admins and it was important to fix the style for this component no matter what). For 2), this style is used in a lot of other places and we need to see if it makes sense to change font size/bold elsewhere.

Did another fix, this part
Screen Shot 2021-05-05 at 09 49 48

app/templates/components/nav_menu_item.html
```diff
@@ -6,7 +6,7 @@
         <a
           {% if id_key %}id="{{ id_key }}"{% endif %}
           href="{{ url }}"
-          class="text-smaller text-blue bg-gray py-5 font-bold no-underline self-center hover:underline line-under visited:text-blue link:text-blue focus:bg-yellow focus:border-b-4 focus:border-blue focus:outline-none {{ css_classes }}"
+          class="text-smaller text-blue bg-gray py-5 font-bold no-underline self-center hover:underline line-under visited:text-blue link:text-blue focus:bg-yellow focus:border-b-4 focus:border-blue focus:outline-none focus:mb-0 {{ css_classes }}"
```
jimleroyer, line 9: A margin bottom of 0 was added in some places, I assume that gave trouble with spacing during positioning?

I worry that defining the margin bottom through the `css_class` parameter to the macro will be prevented in the future because margin bottom > 0 is hardcoded in there with a value that could be configured in future usage.

app/assets/stylesheets/tailwind/views/dashboard.css
```diff
@@ -27,10 +27,13 @@
   }
 
   .file-list-filename {
-    @apply text-smaller leading-tight font-bold block truncate pt-4 pb-12;
+    @apply text-smaller leading-tight font-bold block truncate pt-4 pb-12 border-b-2 border-transparent;
```
jimleroyer, line 30: 👍

app/assets/stylesheets/tailwind/components/typography.css
```diff
@@ -108,5 +109,9 @@
     font-weight: bold;
   }
 
+  .w-max-content {
+    width: max-content;
```
jimleroyer, line 113: Using latest version of tailwind, we'd have these for free.. We use 1.9 because we want to support IE: is that still something we need to do?

.gitignore
```diff
@@ -64,6 +64,7 @@ target/
 
 # Frontend dependencies and compiled assets
 app/assets/stylesheets/govuk_template/.sass-cache/
+app/assets/javascripts/*.LICENSE.txt
```
jimleroyer, line 67: oh yes

app/assets/stylesheets/tailwind/components/typography.css
AntoineAugusti: I think we're fine dropping IE support, we can double check designers/PMs and plan to upgrade soon to leverage new features and faster compilations

app/templates/components/nav_menu_item.html
AntoineAugusti: > A margin bottom of 0 was added in some places, I assume that gave trouble with spacing during positioning?

Yup, it's because the new default focus state has a border bottom of 2px. To avoid jumps in the UI I've added a negative margin bottom of -2px so that the height will stay the same.

> I worry that defining the margin bottom through the `css_class` parameter to the macro will be prevented in the future because margin bottom > 0 is hardcoded in there with a value that could be configured in future usage.

If you put `focus:mb-4` in `css_classes` this will override the value already 👌  (latter takes precedence)

app/assets/stylesheets/tailwind/components/typography.css
jimleroyer: This would really make the CSS work faster!

AntoineAugusti:
@jimleroyer Thanks for the review! I'll change margin bottom to a transparent border, thanks for the advice 🙏

jimleroyer (approved):

PR #1019, opened 2021-05-05
fix: 500 error when going back after sending a notification

AntoineAugusti:
There is a 500 error when you go to the previous page after sending a notification.

To reproduce:
- Go to Templates
- Choose an email template
- Send it to yourself
- On the check `/service/:id/notification/:id/notification/check` page, go back
- Notice the 500 error

## Why?

The session is not properly cleaned up, the relevant code is here

https://github.com/cds-snc/notification-admin/blob/17fdffc716da058b997e63fa8317884e9c4f6f9e/app/main/views/send.py#L1030-L1062

This is a `POST` for `/services/<service_id>/template/<template_id>/notification/check`. If you go back, you reach the same URL but the `GET` endpoint https://github.com/cds-snc/notification-admin/blob/17fdffc716da058b997e63fa8317884e9c4f6f9e/app/main/views/send.py#L956-L962

The error happens in this method https://github.com/cds-snc/notification-admin/blob/17fdffc716da058b997e63fa8317884e9c4f6f9e/app/main/views/send.py#L988, the `send_step` is still present so it tries to prefill stuff but `placeholders` has been cleaned in the session already so it throws an exception.

https://github.com/cds-snc/notification-admin/blob/17fdffc716da058b997e63fa8317884e9c4f6f9e/app/main/views/send.py#L840-L851

Making sure that the `send_step` is cleared out is enough, there is already a test case making sure we're redirected if the session is not populated `test_check_notification_redirects_if_session_not_populated`. Added another defense if someones arrives with a `send_step` in the session but no `placeholders`

## Demo of the fix

https://user-images.githubusercontent.com/295709/117197297-a5f64c80-adb5-11eb-9171-77bb281b0ffa.mov

### DIFF
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -956,7 +956,7 @@ def get_back_link(service_id, template, step_index):
 @main.route("/services/<service_id>/template/<template_id>/notification/check", methods=['GET'])
 @user_has_permissions('send_messages', restrict_admin_usage=True)
 def check_notification(service_id, template_id):
-    if 'send_step' in session and not 'placeholders' in session:
+    if 'send_step' in session and 'placeholders' not in session:
         session.pop('send_step')
     return render_template(
         'views/notifications/check.html',
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -956,6 +956,8 @@ def get_back_link(service_id, template, step_index):
 @main.route("/services/<service_id>/template/<template_id>/notification/check", methods=['GET'])
 @user_has_permissions('send_messages', restrict_admin_usage=True)
 def check_notification(service_id, template_id):
+    if 'send_step' in session and not 'placeholders' in session:
+        session.pop('send_step')
     return render_template(
         'views/notifications/check.html',
         **_check_notification(service_id, template_id),
@@ -1060,6 +1062,7 @@ def send_notification(service_id, template_id):
     session.pop('placeholders')
     session.pop('recipient')
     session.pop('sender_id', None)
+    session.pop('send_step', None)
 
     return redirect(url_for(
         '.view_notification',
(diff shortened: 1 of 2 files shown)


tests/app/main/views/test_send.py
```diff
@@ -3039,7 +3039,8 @@ def test_check_messages_shows_over_max_row_error(
 @pytest.mark.parametrize('existing_session_items', [
     {},
     {'recipient': '6502532223'},
-    {'name': 'Jo'}
+    {'name': 'Jo'},
+    {'send_step': 'main.send_test_step'},
 ])
 def test_check_notification_redirects_if_session_not_populated(
```
jimleroyer, line 3045: Does this test with the changes trigger the issue with the previous code?

app/main/views/send.py
```diff
@@ -956,6 +956,8 @@ def get_back_link(service_id, template, step_index):
 @main.route("/services/<service_id>/template/<template_id>/notification/check", methods=['GET'])
 @user_has_permissions('send_messages', restrict_admin_usage=True)
 def check_notification(service_id, template_id):
+    if 'send_step' in session and 'placeholders' not in session:
```
jimleroyer, line 959: Hmm can we put a comment to explain when that state would happen? why we check for it?

AntoineAugusti:
@jimleroyer Updated the PR to have the most simple resolution: removing the `send_step` from the session. If this is cleared out properly, the code will not try to fill the current user.

The other code in my first commit was too defensive and covering the fact where `send_step` has not been cleared out, testing that we don't error. This will not happen anymore if we clean `send_step` so removed this part to keep it simple.

app/main/views/send.py
jimleroyer: @AntoineAugusti replied below to this question with this 
https://github.com/cds-snc/notification-admin/pull/1019#issuecomment-834300142

tests/app/main/views/test_send.py
jimleroyer: @AntoineAugusti Can you confirm here please?

jimleroyer (approved):

PR #1017, opened 2021-05-04
feat: utils to 43.9.0, add bold, italic and h2

AntoineAugusti:
Bump utils to 43.9.0 https://github.com/cds-snc/notification-utils/pull/82.

This adds bold, italic and h2 to email templates. Updated the templates documentation with these new features and documented how to put variables in links as well.

Screen Shot 2021-05-04 at 10 57 56

Content to review for the documentation: https://notification-utils-43-9-dha6of.herokuapp.com/templates

PR #1021, opened 2021-05-06
fix: don't remove brackets when translating template previews

AntoineAugusti:
# Problem

Template previews leverage brackets `[]` to identify text to translate. Previously the logic removed brackets when it didn't match a word, meaning you could write a title like `[Action required] Foo` and it would be displayed as `Action required Foo` afterwards but the real email would include brackets.

This PRs improves this a bit: if a word is in brackets but not in the list, we keep brackets around. It does not solve yet the issue for words we do translate. This means that if you put `[phone number]` in your template, brackets would be removed and it would be translated 🙄 

This will be fixed later on in utils, after shipping #1017 by likely using a more unique sequence than brackets.

Look at https://github.com/cds-snc/notification-utils/blob/1bcce4033ce9957691bd5b530f74801c7489ee1a/notifications_utils/jinja_templates/email_preview_template.jinja2#L15 or https://github.com/cds-snc/notification-utils/blob/1bcce4033ce9957691bd5b530f74801c7489ee1a/notifications_utils/jinja_templates/sms_preview_template.jinja2#L3

## Demo of this bug
https://user-images.githubusercontent.com/295709/117322642-2c656980-ae5c-11eb-8b59-12bb3575a4ea.mov

## Test
- Create a template with `[Action required] Foo`
- Save it
- Look at the preview, see that the subject is not altered

Trello: https://trello.com/c/VpYPyzga/490-dont-remove-brackets-in-template-previews

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -387,27 +387,22 @@ def format_delta(_date):
 
 
 def translate_preview_template(_template_str):
-    translate = {
-        "From": _("From"),
-        "To": _("To"),
-        "Subject": _("Subject"),
-        "Reply to": _("Reply to"),
-        "From:": _("From:"),
-        "To:": _("To:"),
-        "phone number": _("phone number"),
-        "email address": _("email address"),
-        "hidden": _("hidden"),
-    }
-
     def translate_brackets(x):
-        g = x.group(0)
-        english = g[1:-1]  # drop brackets
-        if english not in translate:
-            return english
-        return translate[english]
+        match, word = x.group(0), x.group(1)
+        return {
+            "From": _("From"),
+            "To": _("To"),
+            "Subject": _("Subject"),
+            "Reply to": _("Reply to"),
+            "From:": _("From:"),
+            "To:": _("To:"),
+            "phone number": _("phone number"),
+            "email address": _("email address"),
+            "hidden": _("hidden"),
+        }.get(word, match)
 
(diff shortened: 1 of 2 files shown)


willeybryan (approved):
Python is weird

but LGTM

app/__init__.py
```diff
@@ -387,27 +387,22 @@ def format_delta(_date):
...
+            "Reply to": _("Reply to"),
+            "From:": _("From:"),
+            "To:": _("To:"),
+            "phone number": _("phone number"),
+            "email address": _("email address"),
+            "hidden": _("hidden"),
+        }.get(word, match)
+
+    # This regex finds words inside []
+    template_str = re.sub(r"\[([^]]*)\]", translate_brackets, _template_str)
```
QuanNguyen-CDS, line 405: how do trailing brackets get handled?

AntoineAugusti: Sorry I don't undertstand

anikbrazeau:
Does the translation bug only apply to certain words like [email address] and [phone number]?

AntoineAugusti:
@anikbrazeau Yes exactly. With the list of words you can find in changes in this PR. This is not new behaviour, it's already like that and will be fixed for good in another iteration in the next weeks.

QuanNguyen-CDS (approved):

PR #1025, opened 2021-05-12
feat: delete user cache when validating Fido key

AntoineAugusti:
When you log in using a Fido key, this endpoint is called

https://github.com/cds-snc/notification-admin/blob/d6e7b9f62a5242546263579cfd2e106aa3b231fb/app/main/views/user_profile.py#L279-L298

which calls an API endpoint to validate the Fido key, through `validate_security_keys`. This records the login date in `logged_in_at` https://github.com/cds-snc/notification-api/blob/1997ebbee52a63f28f0d73771e832ef5df8f652c/app/user/rest.py#L679-L706

Not clearing the cache means that we will get a stale value displayed when looking at a user profile. Clearing the user's cache after calling this endpoint will make sure that we get a fresh user afterwards.

Checked other code paths for sign in and 2FA using email and SMS and I confirmed that we store login events and update `logged_in_at` properly.

Trello: https://trello.com/c/2KwpmvGk/502-last-login-date-is-not-properly-recorded

### DIFF
--- a/app/notify_client/user_api_client.py
+++ b/app/notify_client/user_api_client.py
@@ -230,6 +230,7 @@ class UserApiClient(NotifyAdminAPIClient):
         endpoint = '/user/{}/fido2_keys/authenticate'.format(user_id)
         return self.post(endpoint, {})
 
+    @cache.delete('user-{user_id}')
     def validate_security_keys(self, user_id, payload):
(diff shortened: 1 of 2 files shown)


willeybryan (approved):
LGTM, 
I guess.  Like I come from the world where everything ends in a ;. So I expect there to be a line terminator at the end of 233.

PR #1031, opened 2021-05-18
feat: add flash message when updating service settings

AntoineAugusti:
Add flash messages when updating service settings as a platform admin to say that service users will receive an email.

- when turning a service live
- when updating the daily limit of a live service


willeybryan (approved):
LGTM

PR #1038, opened 2021-05-25
Updated the suggested `flask run` command

smcmurtry:
# Summary | Résumé

When I run the application with `flask run -p 6012 --host=0.0.0.0`, the terminal displays the following link:
Screen Shot 2021-05-25 at 3 43 15 PM

And if I visit `http://0.0.0.0:6012/`, the css and some images do not load because they violate the same origin policy, making the site look broken.
Screen Shot 2021-05-25 at 3 44 38 PM

But if I visit `http://localhost:6012/` everything loads correctly. Changing the host in the `flask run` command fixes the link that gets displayed in my terminal to the correct one: 
Screen Shot 2021-05-25 at 3 47 59 PM


AntoineAugusti (approved):
Thanks!

We could even have a `make sever`/`make dev` in [the Makefile](https://github.com/cds-snc/notification-admin/blob/master/Makefile) if you feel like it

README.md
```diff
@@ -104,7 +104,7 @@ source  ~/.pyenv/versions/3.9.1/bin/virtualenvwrapper.sh
 
 14.  Run the service
 
-`flask run -p 6012 --host=0.0.0.0`
+`flask run -p 6012 --host=localhost`
```
AntoineAugusti, line 107: Can you update the command in the French section as well?

smcmurtry: Yes!

PR #1039, opened 2021-05-26
Move Ready to send heading

AntoineAugusti:
Move the "Ready to send" heading to make sure it's wrapped in the condition "can you send messages".

For templates where you cannot send notifications only the heading was displayed, it was weird.

Screen Shot 2021-05-26 at 11 47 39

Trello: https://trello.com/c/hNx9JBJR/544-do-not-show-ready-to-send-without-permissions

### DIFF
--- a/app/templates/views/templates/_template.html
+++ b/app/templates/views/templates/_template.html
@@ -13,6 +13,8 @@
     </p>
   {% else %}
     {% if current_user.has_permissions('send_messages', restrict_admin_usage=True) %}
+    <h2 class="heading-medium">{{ _('Ready to send?') }}</h2>
+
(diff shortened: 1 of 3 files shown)


app/templates/views/templates/_template.html
```diff
@@ -13,6 +13,8 @@
     </p>
   {% else %}
     {% if current_user.has_permissions('send_messages', restrict_admin_usage=True) %}
+    <h2 class="heading-medium">{{ _('Ready to send?') }}</h2>
```
AntoineAugusti, line 16: Notice the line above if the good `if` statement

jimleroyer (approved):

PR #1040, opened 2021-05-26
Make the request to go live form less confusing

smcmurtry:
# Summary | Résumé

I will convert this from a draft and once the content is finalized and I fix the tests.

Trello: https://trello.com/c/4FX6Lj1K/537-make-the-request-to-go-live-form-less-confusing
Demo: https://www.youtube.com/watch?v=AvCMQOAx9zA
Screenshot: Screen Shot 2021-05-26 at 4 03 31 PM

Steps:
- Create a new service
- click the "request to go live" link
- observe there is now a submit button on the "request to go live" page
- click the submit button, and verify that the bullets in the error message match up to the incomplete steps on the checklist
- complete the checklist, verify that the submit button now allows you to submit the request


smcmurtry:
[Make the request to go live form less confusing](https://trello.com/c/4FX6Lj1K/537-make-the-request-to-go-live-form-less-confusing)

app/templates/views/service-settings/request-to-go-live.html
```diff
@@ -17,6 +18,27 @@
         _('Request to go live'),
         back_link=url_for('main.service_settings', service_id=current_service.id)
       ) }}
+      {% if error %}
+        <div class="mb-12 clear-both contain-floats">
```
AntoineAugusti, line 22: Think `clear-both contain-floats` are not needed

app/templates/views/service-settings/request-to-go-live.html
```diff
@@ -17,6 +18,27 @@
...
+          {% call banner_wrapper(type='dangerous') %}
+            <p class="font-bold">{{ _("Complete all the steps before submitting the request to go live:") }}</p>
```
AntoineAugusti, line 24: If I remember correctly, banners are already bold. We may have a `banner-title` class

app/templates/views/service-settings/request-to-go-live.html
```diff
@@ -17,6 +18,27 @@
...
+            <ul class="list-bullet">
```
AntoineAugusti, line 25: Think it would be good to avoid duplicate checklist tasks, could you consider having a list of dict with has it been done/description/url and loop over this in the table and for errors?

app/templates/views/service-settings/request-to-go-live.html
smcmurtry: you are right!

app/templates/views/service-settings/request-to-go-live.html
smcmurtry: Without that class it is not bold but the `<li>` elements are: 
Screen Shot 2021-05-27 at 11 07 26 AM

If I use `<h2 class="banner-title">` it looks like this: 
Screen Shot 2021-05-27 at 11 08 21 AM

But the way the content is written, it is not exactly a title. Arguably the error message should have a title. But I think it is okay without one since this will be autofocused by screen readers when the error occurs so it will be hard to miss.

app/templates/views/service-settings/request-to-go-live.html
smcmurtry: Good idea!

tests/app/main/views/test_service_settings.py
```diff
@@ -1080,10 +1080,10 @@ def test_request_to_go_live_use_case_page_without_permission(
 
 
 @pytest.mark.parametrize('checklist_completed, expected_button', (
-    (False, False),
+    (False, True),
     (True, True),
 ))
-def test_should_not_show_go_live_button_if_checklist_not_complete(
+def test_should_always_show_go_live_button(
```
AntoineAugusti, line 1086: Could you write an additional test to make sure the banner is shown when at least something has not been completed?

app/templates/views/service-settings/request-to-go-live.html
```diff
@@ -34,32 +49,25 @@
...
-          current_service.has_team_members_status,
-          _('Add a team member who can manage settings'),
-          url_for('main.manage_users', service_id=current_service.id),
-        ) }}
-        {{ task_list_item(
-          current_service.has_accepted_tos,
-          _('Accept the terms of use'),
-          url_for('main.terms_of_use', service_id=current_service.id),
-        ) }}
+        {% for item in current_service.go_live_checklist %}
```
AntoineAugusti, line 52: 💅

app/models/service.py
```diff
@@ -372,6 +372,31 @@ def remove_default_letter_contact_block(self):
...
+            },
+            {
+                'text': _('Accept the terms of use'),
+                'completed': self.has_accepted_tos,
+                'endpoint': 'main.terms_of_use'
+            },
+        ]
+
     @property
     def go_live_checklist_completed(self):
```
AntoineAugusti, line 401: Could you use `go_live_checklist` here?

```py
def go_live_checklist_completed(self):
    return all([checklist['completed'] for checklist in self.go_live_checklist])

```

AntoineAugusti (changes requested):

tests/app/main/views/test_service_settings.py
smcmurtry: Will do!

app/models/service.py
smcmurtry: So I tried doing this, but then it broke this test: https://github.com/cds-snc/notification-admin/blob/5a53697585769a5d46b55ddc92687c326794f02f/tests/app/models/test_service.py#L289
Because the response context is now needed to get the translated text, and that test doesn't have it. I wasn't sure how to fix that test so I just reverted this change.

tests/app/main/views/test_service_settings.py
```diff
@@ -1102,23 +1099,42 @@ def test_should_not_show_go_live_button_if_checklist_not_complete(
...
-        page.select_one('[type=submit]').text.strip() == ('Request to go live')
     else:
-        assert not page.select('form')
-        assert not page.select('[type=submit]')
         paragraphs = [normalize_spaces(p.text) for p in page.select('main p')]
         assert "Once you complete all the steps, you'll be able to submit your request." in paragraphs
 
 
+def test_should_show_error_if_go_live_not_completed(
+    client_request: ClientRequest,
```
smcmurtry, line 1117: Setting the type so that type hints work.

Before:
Screen Shot 2021-05-27 at 4 03 48 PM


After:
Screen Shot 2021-05-27 at 4 03 03 PM

tests/conftest.py
```diff
@@ -3032,95 +3032,102 @@ def os_environ():
     os.environ = old_env
 
 
+class ClientRequest:
```
smcmurtry, line 3035: I moved this class outside of its wrapper function so that it can be imported in the tests and used for static typing.  See my comment in the test file.

AntoineAugusti:
@smcmurtry Made a refactor commit to use the checklist in `go_live_checklist_completed`.

Actually there was a bug because the checklist used `has_team_members_status` and the other method `has_team_members`. The task list needs a status because an item (inviting people can be "in progress") https://github.com/cds-snc/notification-admin/blob/3d7bcebf968dd078ca0f7977e38217511c879bce/app/templates/components/task-list.html#L4 so we need to distinguish between booleans and "in progress".

Tried to use a lazy translation (`_l("foo"` instead of `_("foo")`) but had to use a test request context as you described.

AntoineAugusti (approved):
Really good!

Thanks for going the extra kilometer and refactoring things around the checklist or the `ClientRequest`!

PR #1036, opened 2021-05-21
feat: full width for stick at bottom containers

AntoineAugusti:
Make sure containers with the class `js-stick-at-bottom-when-scrolling` are full width. Previously they were only the size of the container.

Made sure that the container is resized when the window width changes

Trello: https://trello.com/c/ooQJsbYS/531-increase-division-line-for-persistent-page-buttons

PR #1042, opened 2021-05-27
feat: adjust padding left for stick at bottom

AntoineAugusti:
Addresses this commit https://github.com/cds-snc/notification-admin/pull/1036#pullrequestreview-669578399 by adding the padding left value on the `main` element to align stuff properly. The transition is also faster.

Screen Shot 2021-05-27 at 15 18 29


app/assets/javascripts/stick-to-window-when-scrolling.js
```diff
@@ -937,10 +937,18 @@
       return;
     }
 
+    var main = document.getElementsByTagName('main')[0];
```
jimleroyer, line 940: I just noticed, because it was included in previous PR... Will the `main` HTML element always be present? Should we fall back onto some other element?

app/assets/javascripts/stick-to-window-when-scrolling.js
```diff
@@ -937,10 +937,18 @@
...
+    var paddingValue = global.getComputedStyle(main).getPropertyValue('padding-left');
+
+    var mainPaddingLeft = 0;
+    if (new RegExp(/^\d+px$/).test(paddingValue)) {
+      mainPaddingLeft = parseInt(paddingValue);
```
jimleroyer, line 945: So there was an issue on evaluating measurements defined as pixel? Any other trouble we could get with different measurements (i.e. `rem`, `em`)?

app/assets/stylesheets/tailwind/components/stick-at-top-when-scrolling.css
```diff
@@ -88,7 +88,7 @@
   }
 
   .js-stick-at-bottom-when-scrolling.content-fixed {
-    @apply transition-background duration-600 ease-in-out;
+    @apply transition-background duration-300 ease-in-out;
```
jimleroyer, line 91: I won't have time to get a ☕  now in between the animation start and end... 😞 😉

jimleroyer (commented):
it looks good overall,  just have a few questions.

app/assets/javascripts/stick-to-window-when-scrolling.js
AntoineAugusti: Added a fallback mechanism!

app/assets/javascripts/stick-to-window-when-scrolling.js
AntoineAugusti: We only support summing pixels at the moment. This test is here to make sure we don't sum pixels with rem or em. If it gets converted to rem/em, the div will not be aligned on the left side but at least it will not be on the other side of the screen.

e-pang (approved):
Thanks so much for the updates, It's looking great :)!

PR #1045, opened 2021-05-31
feat: rework edit template for SMS templates

AntoineAugusti:
Put "edit" and "previous versions" under (not on the same line) for SMS templates

Screen Shot 2021-05-31 at 11 08 30


e-pang (approved):
Looks good, thanks Antoine!

PR #1047, opened 2021-05-31
New "create template" page

smcmurtry:
# Summary | Résumé

Figma: https://www.figma.com/file/fRVRKm7cM8MUuBWUJbrfgU/Sending-from-web-interface-%7C-Envoyer-par-interface-web?node-id=2803%3A3756

To do:
- [x] create a template for the new "create" page
- [x] add routes for the new page into the controller
- [x] add correct back route to the "create" page
- [x] update back routes for the "add-email" and "add-sms" pages to point to "create"
- [x] update the "Create template" button on "templates" page to point to new "create" page
- [x] remove javascript that was used to display the "email or sms" form
- [x] remove other old code
- [x] make sure it works for services without sms sending permission
- [x] add the form field to the new "create" page
- [x] connect the "Continue" button so it brings users to the correct page
- [x] update/add tests


smcmurtry:
[[Create a new template] Create template button goes to a new screen](https://trello.com/c/fe36VXXc/539-create-a-new-template-create-template-button-goes-to-a-new-screen)

app/assets/javascripts/templateFolderForm.js
```diff
@@ -300,7 +307,7 @@
     this.nothingSelectedButtons = $(`
       <div id="nothing_selected">
         <div class="js-stick-at-bottom-when-scrolling">
-          <button class="button js-button-action" value="add-new-template">${window.polyglot.t(
+          <button class="button" value="add-new-template">${window.polyglot.t(
```
smcmurtry, line 310: Removing the `js-button-action` class removes the behaviour where another element on the page is unhidden and focussed. Now we just want to nav to a new page when this is clicked.

app/templates/views/edit-email-template.html
```diff
@@ -14,7 +14,7 @@
     page_header(
       _(heading_action) + ' ' + _('email template'),
       back_link=url_for('main.view_template', service_id=current_service.id,
-      template_id=template.id) if template else url_for('main.choose_template',
+      template_id=template.id) if template else url_for('main.create_template',
```
smcmurtry, line 17: Clicking back from the edit page now goes back to the "create_template" page unless you are editing a template that already existed.

app/templates/views/templates/_move_to.html
```diff
@@ -30,10 +30,9 @@
   </div>
   <div id="add_new_template_form">
     <div class="js-will-stick-at-bottom-when-scrolling">
-    {{ radios(templates_and_folders_form.add_template_by_template_type) }}
-    </div>
-    <div class="js-will-stick-at-bottom-when-scrolling">
-    {{ page_footer(_('Continue'), button_name='operation', button_value='add-new-template') }}
+      <a class="button" href="{{ url_for('main.create_template', service_id=current_service.id, template_type=template_type, template_folder_id=current_template_folder_id) }}">
```
smcmurtry, line 33: This is for the no JS experience - if javascript is blocked, this element will not be hidden, and clicking it will nav to the "create template" page

pytest.ini
```diff
@@ -15,6 +15,6 @@ env =
 
 filterwarnings =
     error:Applying marks directly:pytest.RemovedInPytest4Warning
-addopts = -p no:warnings
+addopts = -p no:warnings -n4
```
smcmurtry, line 18: This sets the default number of parallel processors that should be used for executing the tests. This will now match what we have set over here: https://github.com/cds-snc/notification-admin/blob/master/scripts/run_tests.sh#L41

tests/__init__.py
```diff
@@ -9,9 +9,18 @@
 from flask.testing import FlaskClient
 from flask_login import login_user
 
+from app.models.service import Service
 from app.models.user import User
 
 
+class dotdict(dict):
```
smcmurtry, line 16: This extends the object returned from `service_json` so that its keys can be accessed or set like this `service.key` or like this `service["key"]`. That makes the returned object more like the `Service` class that we are trying to mock. That way, we can type the mocked services like this: https://github.com/cds-snc/notification-admin/pull/1047/files#diff-219eb25d059813c96fb4d720851a846e97a83875ce8318b92ebb85188530ffb6R1246

tests/app/main/views/test_template_folders.py
```diff
@@ -1238,38 +1239,18 @@ def test_should_be_able_to_move_a_sub_item(
             "move_to_new_folder_name": "foo",
             "move_to": None,
         },
-        # add a new template, but also select move destination
```
smcmurtry, line 1241: Deleting these, since we always want to nav to the "create template" page if "add-new-template" is clicked

tests/app/main/views/test_templates.py
```diff
@@ -1819,6 +1816,7 @@ def test_route_permissions_for_choose_template(
 @pytest.mark.parametrize(
     "route",
     [
+        "main.create_template",
```
smcmurtry, line 1819: Test that you can't see this route with invalid permissions

tests/app/main/views/test_templates.py
```diff
@@ -1757,6 +1753,7 @@ def test_should_show_page_for_a_deleted_template(
...
```
smcmurtry, line 1756: Test that you can see this route with valid permissions

app/main/forms.py
```diff
@@ -1469,6 +1469,14 @@ def validate(form, field):
     return validate
 
 
+class CreateTemplateForm(Form):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.what_type.choices = list(filter(None, [("email", _l("Email")), ("sms", _l("Text message"))]))
```
AntoineAugusti, line 1475: Is the `filter(None` useful? I don't understand why it's needed

app/main/views/templates.py
```diff
@@ -334,6 +329,34 @@ def _add_template_by_type(template_type, template_folder_id):
...
+@main.route("/services/<service_id>/templates/folders/<template_folder_id>/create", methods=["GET", "POST"])
+@main.route("/services/<service_id>/templates/<template_type>/create", methods=["GET", "POST"])
+@main.route("/services/<service_id>/templates/<template_type>/folders/<template_folder_id>/create", methods=["GET", "POST"])
+@user_has_permissions("manage_templates")
+def create_template(service_id, template_type="all", template_folder_id=None):
+    form = CreateTemplateForm()
+
+    if request.method == "POST" and form.validate_on_submit():
+        if not current_user.has_permissions("manage_templates"):
+            abort(403)
```
AntoineAugusti, line 342: I believe the `user_has_permissions` on the endpoint is enough?

app/main/views/templates.py
```diff
@@ -334,6 +329,34 @@ def _add_template_by_type(template_type, template_folder_id):
...
+        try:
+            return _add_template_by_type(
+                form.what_type.data,
+                template_folder_id,
+            )
+        except HTTPError as e:
+            flash(e.message)
```
AntoineAugusti, line 349: Is it going to be translated? My feeling is that it's coming directly from the API without translations

app/templates/views/templates/create.html
```diff
@@ -0,0 +1,27 @@
...
+  {% set header_txt = _('Create template') %}
+  {{ page_header(
+    header_txt,
+    back_link=url_for('main.choose_template', service_id=service_id, template_type=template_type, template_folder_id=template_folder_id)
+  ) }}
+ 
+  {% set label_txt = _('Name for this key') %}
+  {% set button_txt = _('Continue') %}
+    {% call form_wrapper() %}
+    {{ radios(form.what_type, disable=disabled_options, option_hints=option_hints) }}
```
AntoineAugusti, line 22: `disabled_options` is not defined

app/main/forms.py
```diff
@@ -1550,15 +1555,6 @@ def get_folder_name(self):
     add_new_folder_name = StringField(_l("Folder name"), validators=[required_for_ops("add-new-folder")])
     move_to_new_folder_name = StringField(_l("Folder name"), validators=[required_for_ops("move-to-new-folder")])
 
-    add_template_by_template_type = RadioFieldWithRequiredMessage(
-        _l("Create template"),
-        validators=[
-            required_for_ops("add-new-template"),
-            Optional(),
-        ],
-        required_message=_l("Select the type of message you want to create"),
```
AntoineAugusti, line 1559: Should be able to remove this translation string?

app/templates/views/templates/create.html
```diff
@@ -0,0 +1,27 @@
+{% extends "admin_template.html" %}
+{% from "components/page-header.html" import page_header %}
+{% from "components/page-footer.html" import page_footer %}
+{% from "components/textbox.html" import textbox %}
```
AntoineAugusti, line 4: Unused I believe

app/assets/javascripts/templateFolderForm.js
```diff
@@ -275,17 +281,18 @@
...
+        window.location.href = `${window.location.href}/create`
+      }
+
       // use dialog mode for states which contain more than one form control
-      if (
-        ["move-to-existing-folder", "add-new-template"].indexOf(
-          this.currentState
-        ) !== -1
-      ) {
+      if ( this.currentState === "move-to-existing-folder") {
```
AntoineAugusti, line 295:
```suggestion
      if (this.currentState === "move-to-existing-folder") {
```

AntoineAugusti (approved):
LGTM! A few comments/questions!

app/main/forms.py
smcmurtry: Good point - I copy pasted that from another radio button - I'll remove

app/templates/views/templates/create.html
smcmurtry: Good catch!

app/templates/views/templates/create.html
smcmurtry: I will fix!

app/main/views/templates.py
smcmurtry: I'm not sure - I haven't been able to trigger this error to test. I copied the logic from the "choose template" route: https://github.com/cds-snc/notification-admin/blob/feature/create-template-pg/app/main/views/templates.py#L141

In some other places the logic is just to `raise e`. I could do that instead if you think that makes more sense.

app/main/views/templates.py
smcmurtry: fixed!

PR #1049, opened 2021-06-01
Experiment: Formatting all python files with black

smcmurtry:
# Summary | Résumé

## What is this?!

As an experiment, I used the python formatter [Black](https://github.com/psf/black#the-uncompromising-code-formatter) to automatically format all the python files in our codebase.

## Why?!

I have always found it tedious to fix my formatting errors by hand. Our test suite tells us which lines in which files need fixing, but it is still a manual task to understand each error and then go and fix it. This adds to the long list of things that slows me down when I work on Notify. To quote the sales pitch from Black:
> Black is the uncompromising Python code formatter. By using it, you agree to cede control over minutiae of hand-formatting. In return, Black gives you speed, determinism, and freedom from pycodestyle nagging about formatting. You will save time and mental energy for more important matters.

## Downsides

- Would make submitting code upstream, or pulling GDS code downstream more challenging
- Maybe you don't like the Black code style?

## Next steps

Some discussion! If we were to go this route, ~Black would need to replace the `flake8 .` code style check, since it seems flake8 wants a different code style.~ Black and flake8 are compatible with a bit of configuration!


jimleroyer:
+1 for cleaner code. I like the idea of having a consistent code base and that would improve it. Practically speaking though, as noted, this could make import of code harder from other forks. How frequently does that happen though? A solution could be to also format their code base with the formatter into new forks to make it easier to import too if it was such a pain (or convince them to apply black or other?).

In terms of actual formatting rules, I don't mind much personally, as long as we stick to one and it is applied on the code base.

smcmurtry:
[Automatic code formatting](https://trello.com/c/jUapnOBo/548-automatic-code-formatting)

.flake8
```diff
@@ -1,7 +1,10 @@
...
 # Rule definitions: http://flake8.pycqa.org/en/latest/user/error-codes.html
 # W503: line break before binary operator
+# In most cases Black follows the rules below, but there are some exceptions
+# E203: whitespace before ':'
+# E501: line too long
 exclude = venv*,__pycache__,node_modules,cache
-ignore = W503
+ignore = W503, E203, E501
 max-complexity = 14
 max-line-length = 130
```
AntoineAugusti, line 10:
```suggestion
max-line-length = 88
```
To be consistent with `black`

Makefile
```diff
@@ -66,4 +66,8 @@ coverage: venv ## Create coverage report
 
 .PHONY: run-dev
 run-dev:
-	flask run -p 6012 --host=localhost
\ No newline at end of file
+	flask run -p 6012 --host=localhost
+
+.PHONY: format
+format:
+	black ./
```
AntoineAugusti, line 73: Could you add `black` in `requirements-test.txt`?

.flake8
```diff
@@ -1,7 +1,10 @@
 [flake8]
```
AntoineAugusti, line 1: Do we need to update our configuration with `isort`?

https://black.readthedocs.io/en/stable/guides/using_black_with_other_tools.html?highlight=isort

I feel like https://github.com/cds-snc/notification-admin/blob/master/setup.cfg may need to be updated

Makefile
AntoineAugusti: Does this format only `app` and `tests` not `venv`?

AntoineAugusti (changes requested):
Looks good!

Could you update the `run_tests.sh` to run `black --check` to make sure tests fail if formatting is off?

.flake8
smcmurtry: Good catch! I'll do that.

Makefile
smcmurtry: Black is in [`requirements_for_test.txt`](https://github.com/cds-snc/notification-admin/blob/experiment/adding-black/requirements_for_test.txt#L16)

smcmurtry: > Does this format only `app` and `tests` not `venv`?

Good call, I'll change this to `black ./app ./tests`

.flake8
smcmurtry: I set the black line length to be 130 over here: https://github.com/cds-snc/notification-admin/blob/experiment/adding-black/pyproject.toml#L2. 

But I can change them both to be 88 if you prefer.

AntoineAugusti: Sounds good, sorry I missed this, many files were changed 😅

smcmurtry:
> Could you update the `run_tests.sh` to run `black --check` to make sure tests fail if formatting is off?

Yes! Good idea.

.flake8
smcmurtry: Haha, yes this file was buried in the 182 other files that were changed.

AntoineAugusti (approved):

PR #1052, opened 2021-06-03
Attempting to upgrade py.test - what a mess

smcmurtry:
# Summary | Résumé

I started trying to do this so that we could make use of the type definitions that now exist for the latest py.test. It turns out a lot of things in the tests would need to be re-written to do this upgrade. I started doing this, but not sure if I will continue.


smcmurtry:
Maybe I will revisit this later.

PR #1053, opened 2021-06-04
Tweaks to team member permission content

anikbrazeau:
1. Clarify dashboard permission
2. Be consistent with create template terminology rather than add template
3. Change order of send message permission to be below create template


AntoineAugusti (approved):
I'm good if you're good

PR #1055, opened 2021-06-04
Experiment: Adding MyPy

smcmurtry:
# Summary | Résumé

## What?

Added [mypy](https://github.com/python/mypy) to `scripts/run_tests.sh`.
> Mypy is an optional static type checker for Python. You can add type hints (PEP 484) to your Python programs, and use mypy to type check them statically.

## Why?

With static type checking, we can catch some errors in our code automatically. 

## Limitations

Many of the libraries we are importing don't come with type definitions. To prevent these from throwing errors during the mypy check, I've ignored those libraries in the [mypy config file](https://github.com/cds-snc/notification-admin/blob/experiment/add-mypy/mypy.ini). Probably we can get type definitions working for many of these, it will just requiring updating libraries or adding some packages.

## Next steps

Some discussion!


app/navigation.py
```diff
@@ -4,8 +4,8 @@
 
 
 class Navigation:
-    mapping = {}
-    exclude = {}
+    mapping: dict = {}
+    exclude: set = set()
```
smcmurtry, line 8: Here is an example of mypy actually catching a real error. This variable was first set to be a `dict`, and later becomes a `set`.

scripts/run_tests.sh
```diff
@@ -26,17 +26,20 @@ display_result $? 1 "Requirements check"
 make babel
 
 black ./app ./tests --check
-display_result $? 2 "Code style check (Black)"
+display_result $? 1 "Code style check (Black)"
```
smcmurtry, line 29: These numbers are not meaningful so I just set them all to be `1`

app/models/user.py
```diff
@@ -615,7 +615,7 @@ class OrganisationUsers(Users):
     client = user_api_client.get_users_for_organisation
 
 
-class InvitedUsers(Users):
+class InvitedUsers(ModelList):
```
smcmurtry, line 618: Here mypy threw an error because `model` is of type `User` in the `Users` class, but we were trying to set it to an object of a different type, `InvitedUser`, in this class. By making `InvitedUsers` not inherit from `Users` this is fixed.

scripts/run_tests.sh
```diff
@@ -26,17 +26,20 @@ display_result $? 1 "Requirements check"
...
 
 flake8 .
 display_result $? 1 "Code style check (flake8)"
 
 isort --check-only ./app ./tests
-display_result $? 2 "Import order check"
+display_result $? 1 "Import order check"
+
+mypy ./
```
AntoineAugusti, line 37: Do we need to say that this should only run for `app` and `tests`?

app/models/user.py
```diff
@@ -624,6 +624,10 @@ def __init__(self, service_id):
...
 
 
-class OrganisationInvitedUsers(InvitedUsers):
+class OrganisationInvitedUsers(ModelList):
+
     client = org_invite_api_client.get_invites_for_organisation
     model = InvitedOrgUser
+
+    def __init__(self, service_id):
+        self.items = [user for user in self.client(service_id) if user["status"] != "accepted"]
```
AntoineAugusti, line 633: Why do we need to add this?

AntoineAugusti (commented):
Great! A few questions.

Can you remove your Redis dump `dump.rdb` and add it to `.gitignore`?

scripts/run_tests.sh
smcmurtry: There are some .py files outside of those directories, like `/application.py`, `/gunicorn_config.py` and the .py files in `scripts/`. I think it might make sense to change isort and black to also look at the whole project.

smcmurtry:
> Great! A few questions.
> 
> Can you remove your Redis dump `dump.rdb` and add it to `.gitignore`?

Good catch! That's done now.

app/models/user.py
smcmurtry: `OrganisationInvitedUsers` was previously a sub class of `InvitedUsers`, which was a subclass of `Users`. They all had objects of different types assigned to the `model` property, which threw a type error. So I made them all independent classes and copied the `__init__` method here from `InvitedUsers` since it was no longer being inherited.

app/models/__init__.py
```diff
@@ -51,12 +51,12 @@ def _get_by_id(self, things, id):
...
     @property
     @abstractmethod
-    def client():
+    def client(self):
         pass
 
     @property
     @abstractmethod
-    def model():
+    def model(self):
```
jimleroyer, line 59: 😓 💦

app/models/user.py
jimleroyer: This is great! It sounds like `InvitedUsers` should be part of a `User` hierarchy but I really like the fact that _mypy_ exposes our inconsistent typing assumptions.

app/models/user.py
jimleroyer: Do you think it would be worth it to remodel a bit the class hierarchy around user/users @smcmurtry ? (not in this PR but eventually)

mypy.ini
```diff
@@ -0,0 +1,71 @@
...
+ignore_missing_imports = True
+
+[mypy-requests_mock.*]
+ignore_missing_imports = True
+
+[mypy-flask_caching.*]
+ignore_missing_imports = True
+
+[mypy-newrelic.*]
+ignore_missing_imports = True
```
jimleroyer, line 71: Hope we can remove most of these _ignore missing imports_, eventually.

scripts/run_tests.sh
jimleroyer: We should try, yeah. Also, my stress tests PR adds a new folder at the project's root.

jimleroyer (approved):
It looks good to me. I see that @smcmurtry addressed @AntoineAugusti 's comments. I asked a few questions but these are not blockers, just wondering about MyPy's capacity and working.

app/models/user.py
smcmurtry: Maybe? It would be good to know what methods/properties exist on all user types. But I think the type hints should now be able to give us that with the existing classes.

app/main/forms.py
```diff
@@ -604,7 +604,7 @@ class CreateServiceStepNameForm(StripWhitespaceForm):
 
 
 class CreateServiceStepLogoForm(StripWhitespaceForm):
-    def _getSelectBilingualChoices(self) -> dict:
+    def _getSelectBilingualChoices(self):
```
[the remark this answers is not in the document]
smcmurtry: I could have replaced this with `-> List[(str, str)]` or something like that. But I wanted to stay away from adding a lot of type annotations in this PR.

app/__init__.py
```diff
@@ -21,7 +21,7 @@
     url_for,
 )
 from flask._compat import string_types
-from flask.globals import _lookup_req_object, _request_ctx_stack
+from flask.globals import _lookup_req_object, _request_ctx_stack  # type: ignore
```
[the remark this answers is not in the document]
smcmurtry: Mypy threw an error here because it didn't think `_lookup_req_object` was a function in `flask.globals`. But I looked up the source code and I was able to find it. Not sure what the issue was.

app/models/__init__.py
```diff
@@ -6,7 +6,7 @@
 
 class JSONModel:
 
-    ALLOWED_PROPERTIES = set()
+    ALLOWED_PROPERTIES: set = set()
```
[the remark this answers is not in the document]
smcmurtry: I think mypy can infer the type here, but it was still suggesting adding a type declaration as a best practice.

app/main/forms.py
```diff
@@ -416,7 +416,7 @@ def __init__(self, invited_org_user):
 )
 
 
-class PermissionsForm(PermissionsAbstract):
+class PermissionsForm(PermissionsAbstract):  # type: ignore
```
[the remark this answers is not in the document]
smcmurtry: I just added it here because mypy was complaining about `PermissionsAbstract` not being a real class. It's defined in a weird way: https://github.com/cds-snc/notification-admin/blob/experiment/add-mypy/app/main/forms.py#L412-L416

I tried to rewrite it as a class but then that caused some tests to fail, so I gave up for now.

mypy.ini
smcmurtry: Yes! I think upgrading libraries and installing the missing type definitions should fix some of these.

PR #1059, opened 2021-06-07
Adding the "write" link to the dashboard

smcmurtry:
# Summary | Résumé

Added the "write" task-based shortcut to the dashboard. This will only be displayed if the user has the "manage_templates" permission. This links to the "create template" page. The back link from this page was modified to point to the dashboard if the user comes from there. 

I will add the "send" link in a follow up PR.   

With write permission:
Screen Shot 2021-06-08 at 11 58 40 AM

Without write permission:
Screen Shot 2021-06-08 at 11 58 29 AM


smcmurtry:
[Implement changes to templates' entry point](https://trello.com/c/kZ9SC53l/541-implement-changes-to-templates-entry-point)

anikbrazeau:
This is awesome! 👍  

Curious, will the button style eventually match the format in the other PR with blue buttons, no underline, focus states applied once that's merged?

app/templates/views/dashboard/task-shortcuts.html
```diff
@@ -0,0 +1,24 @@
+{% from "components/task-shortcut.html" import task_shortcut %}
+
+<h2 class="heading-medium visuallyhidden">{{ _('Get started') }}</h2>
+
+<div class="flex space-x-gutter mb-16 mt-12">
+  {% if current_user.has_permissions('manage_templates') %}
+    {{ task_shortcut(
```
AntoineAugusti, line 7: Is it expected to take the full width where there is only 1 item? I find it's taking too much space. Feel free to disregard.

app/templates/views/dashboard/dashboard.html
```diff
@@ -14,14 +14,14 @@
...
 
-    <h1 class="visuallyhidden">{{ _('Dashboard') }}</h1>
-    {% if current_user.has_permissions('manage_templates') and not current_service.all_templates %}
-      {% include 'views/dashboard/write-first-messages.html' %}
+    <h1 class="heading-large">{{ _('Dashboard') }}</h1>
+    {% if current_user.has_permissions('manage_templates') or current_user.has_permissions('send_messages') %}
+      {% include 'views/dashboard/task-shortcuts.html' %}
     {% endif %}
-
+
```
AntoineAugusti, line 21: Remove useless spaces?

app/templates/components/task-shortcut.html
```diff
@@ -0,0 +1,18 @@
...
+        class="flex-0 w-10 h-10" 
+        aria-hidden="true" 
+        src="{{ img_src }}"
+      />
+      <h2 class="font-bold text-base ml-2 flex-1">
+        {{ title }}
+      </h2>
+    </div>
+    <p>{{ description }}</p>
+    <a class="button-blue mt-auto" href="{{ button_url }}">
```
AntoineAugusti, line 14: I think you removed the class `button-blue` in another PR?

app/templates/components/task-shortcut.html
```diff
@@ -0,0 +1,18 @@
+{% macro task_shortcut(title, img_src, description, button_url, button_text) %}
+  <div class="border border-blue p-gutterHalf flex-1 flex flex-col">
+    <div class="flex mb-8">
+      <img 
+        class="flex-0 w-10 h-10"
```
AntoineAugusti, line 5: Does `flex-0` exists?

app/templates/views/dashboard/task-shortcuts.html
```diff
@@ -0,0 +1,24 @@
+{% from "components/task-shortcut.html" import task_shortcut %}
+
+<h2 class="heading-medium visuallyhidden">{{ _('Get started') }}</h2>
```
AntoineAugusti, line 3: Curious why do you think this is needed?

app/templates/views/dashboard/task-shortcuts.html
```diff
@@ -0,0 +1,24 @@
...
+
+<div class="flex space-x-gutter mb-16 mt-12">
+  {% if current_user.has_permissions('manage_templates') %}
+    {{ task_shortcut(
+        _("Write"),
+        asset_url('images/dashboard/edit-solid.svg'),
+        _("Prepare to send messages by creating your own templates."),
+        url_for('.create_template', service_id=current_service.id, source="dashboard"),
```
AntoineAugusti, line 11: Wondering if we want to track those links, count how many times they were clicked. Do we? Could add like `?from_dashboard=True` or something

AntoineAugusti (approved):

app/templates/components/task-shortcut.html
smcmurtry: Good catch! This exists in a React library but not in Tailwind. Looks like what I want is [`flex-none`](https://tailwindcss.com/docs/flex#none). I'll fix that.

app/templates/components/task-shortcut.html
smcmurtry: Yes - I was going to change this to just be a button in a follow up, but it probably makes more sense to just do it in this PR. Will fix.

app/templates/views/dashboard/task-shortcuts.html
smcmurtry: Yes it's full width on figma: https://www.figma.com/file/fRVRKm7cM8MUuBWUJbrfgU/Sending-from-web-interface-Envoyer-par-interface-web?node-id=3690%3A17648

AntoineAugusti: Find it massive but okay then!

app/templates/views/dashboard/task-shortcuts.html
smcmurtry: This should already be working! When running locally, `url_for` returns `http://localhost:6012/services/f4536fcb-93c0-478f-960a-837c6b942750/templates/create?source=dashboard`.

AntoineAugusti: I totally missed your extra arg, sorry! Good thought!

app/templates/views/dashboard/task-shortcuts.html
smcmurtry: Yeah it's probably not necessary to have a heading for the shortcuts section. Also "Write" and "Send" are already `h2`s. I'll remove this 👍

smcmurtry:
> Curious, will the button style eventually match the format in the other PR with blue buttons, no underline, focus states applied once that's merged?

Yes!

PR #1060, opened 2021-06-08
Made the default button colour blue

smcmurtry:
# Summary | Résumé

This changes all the green buttons in the app to be blue, to match figma.
Before:
Screen Shot 2021-06-08 at 4 47 34 PM

After:
Screen Shot 2021-06-08 at 4 47 38 PM


anikbrazeau:
AMAZING! ✨ 🤩 👍 Thanks so much, Steve!

Clicked through as many pages as I could to test -- noticed there's a few instances where the focus state breaks: 
- Add reply-to address button
- Add recipients button

Maybe this should be a separate issue/PR?

AntoineAugusti (approved):
Sweet 💙

smcmurtry:
Yikes! Okay I will fix that focus issue.

smcmurtry:
Okay I tracked that issue down and it was just happening for unvisited anchor links. I made the css selectors for `a.button` more specific and that has fixed the issue. I also reduced the width of the focus outline a bit to make it more consistent with the rest of the site.

smcmurtry:
[Some focus styles are not accessible](https://trello.com/c/GTA92zTx/559-some-focus-styles-are-not-accessible)

PR #1064, opened 2021-06-11
Making part of the dashboard visible to all users

smcmurtry:
# Summary | Résumé

Here is an idea for how we could make the write and send links available for all users who have those permissions but not the "view_activity" permission. We are describing this permission as "See dashboard statistics":
Screen Shot 2021-06-11 at 12 03 07 PM

Maybe it is okay if users without this permission can see the dashboard but not the statistics?

This PR is a draft since a lot of tests need to be fixed if we merge this. Feel free to close and I can pick it up again in 3 weeks.


smcmurtry:
Closing - will revisit later.

PR #1065, opened 2021-06-11
Sending view: Fixed a nav bug, fixed view with 0 templates

smcmurtry:
# Summary | Résumé

This PR fixes 2 bugs:
## Bug 1
If a service had 0 templates, clicking on the sending view of the Templates page looked like this:
Screen Shot 2021-06-11 at 3 41 49 PM

With the fix in this PR, it now looks like this:
Screen Shot 2021-06-11 at 3 42 46 PM

## Bug 2
In the sending view on the Templates page, if you clicked a folder link, this would return you back to the non-sending view with checkboxes: https://www.youtube.com/watch?v=seHMHhAvo_k

With the fix in this PR, the sending view is maintained: https://www.youtube.com/watch?v=3ZZVMQm15iA


AntoineAugusti:
Updated to not display "Send messages" when there are no templates yet, even if you can send messages

AntoineAugusti (approved):

PR #1071, opened 2021-06-21
feat: disable the API permission for some services

AntoineAugusti:
Disabled access to API related pages on the admin for some services.

This is controlled by a new config variable, that should be set on Kubernetes to be useful (pass a comma separated list of UUIDs like `b653467a-8d11-44a3-be8e-7225c42ccf8f,69267038-6204-477a-8dbd-60e2a6adc848`.

This is not a well designed long term solution but an interim way to solve a requirement for a client.

- Pages are not accessible (they would get a 403)
- Menu items are hidden

⚠️ You need to create new env variables on Kubernetes to make this work in production. The env can be empty in staging but needs to include the relevant service ID in production ⚠️ 

Trello: https://trello.com/c/GrV6fssa/574-modify-a-single-rcmp-service-to-not-allow-them-to-create-api-keys-for-any-reason


jimleroyer (approved):
Looks amazing @AntoineAugusti

app/models/user.py
```diff
@@ -214,11 +214,14 @@ def has_permissions(self, *permissions, restrict_admin_usage=False, allow_org_us
         if not permissions and self.belongs_to_service(service_id):
             return True
 
+        from app.models.service import Service
+
+        if "manage_api_keys" in permissions and Service.from_id(service_id).api_is_disabled:
+            return False
```
jimleroyer, line 220: Hmm reviewing this on double check, I wonder if a user with the `manage_api_keys` permission on a given restricted service would get restricted further from other pages that user should have access to. Given, the solution with these changes would be to remove such permission for the user, as the service should not have this permission set anyway.

PR #1122, opened 2021-08-06
Fix horizontal scrolling in admin view

smcmurtry:
# Summary | Résumé

Trello: https://trello.com/c/KkjoPad1/648-fix-horizontal-scrolling-on-the-admin-view-of-the-all-services-table

Truncate the service name at 35 characters. This was about as high as I could go before the table was wider than the content above it. You can still see the full service name by clicking the link in the table. 

This fixes both the Trial services and Live services pages.

Before:
Screen Shot 2021-08-06 at 1 06 28 PM

After:
Screen Shot 2021-08-06 at 4 49 29 PM

### DIFF
--- a/app/templates/views/platform-admin/services.html
+++ b/app/templates/views/platform-admin/services.html
@@ -47,7 +47,7 @@
 
         {% call row() %}
           {% call field(border=False) %}
-            <a href="{{ url_for('main.service_dashboard', service_id=service['id']) }}" class="browse-list-link">{{ service['name'] }}</a>
+            <a href="{{ url_for('main.service_dashboard', service_id=service['id']) }}" class="browse-list-link">{{ service['name']|truncate(35) }}</a>
           {% endcall %}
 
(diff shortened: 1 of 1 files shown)


app/templates/views/platform-admin/services.html
```diff
@@ -47,7 +47,7 @@
 
         {% call row() %}
           {% call field(border=False) %}
-            <a href="{{ url_for('main.service_dashboard', service_id=service['id']) }}" class="browse-list-link">{{ service['name'] }}</a>
+            <a href="{{ url_for('main.service_dashboard', service_id=service['id']) }}" class="browse-list-link">{{ service['name']|truncate(35) }}</a>
```
sastels, line 50: nice

sastels (approved):
LGTM. Some issues around whether we should break in the middle of words or not, but as discussed IRL that shouldn't matter for real service names. (as opposed to random string service names generated for testing). I'm good either way.

PR #1124, opened 2021-08-12
Fix the focus style bug for "Change" links in settings

smcmurtry:
# Summary | Résumé

Looks like this was caused by inserting a `::before` pseudo element. I'm not sure what that element was actually doing, because the table layout did not appear to change after I removed it. I wondered if this might have an effect on the date picker component since this caused that css to change. But I checked and there are no html `table` elements being used there.

Removing the `display: "block"` from the anchor tag corrected the width of the focus background and made the row height slightly smaller.

I have tested this in Chrome and Firefox, ~will check IE shortly~. Update: tested on Edge and IE using Microsoft Surface and it looks the same.

Trello: https://trello.com/c/It5VYlCx/650-bug-focus-state-for-change-in-settings

Before:
Screen Shot 2021-08-10 at 10 07 50 AM

After:
Screen Shot 2021-08-12 at 9 51 36 AM


jimleroyer (commented):
Looks good. I would just revert the `scheduler.min.js` script as nothing changed in there before a merge.

jimleroyer (approved):
LGTM!

PR #1126, opened 2021-08-16
Fix for lambda API

smcmurtry:
# Summary | Résumé

When making a request to the `/service/delivered-notifications-stats-by-month-data` endpoint, data is being returned in a different format.

This is from the lambda API:
```
{
  "data": [
    {
      "count": 292322,
      "month": "2021-08-01 00:00:00+00",
      "notification_type": "email"
    },
    {
      "count": 86,
      "month": "2021-08-01 00:00:00+00",
      "notification_type": "sms"
    },
...
```

This is from the k8s API:
```
{'data': [['2021-08-01 00:00:00+00', 'email', 292322], ['2021-08-01 00:00:00+00', 'sms', 86], ...
```

Strangely, this is the only part of the api I have found so far that is behaving differently. All the other parts seem to be working the same with the lambda api:
- I can create a new user account
- I can create a new service
- I can create a template
- I can send a test message to myself from admin
- I can create an API key
- I can view the admin pages

So this is a somewhat hacky fix to parse the API response differently if we are connected to the lambda api at https://api-lambda.staging.notification.cdssandbox.xyz. It relies on "lambda" being in the URL. I could also add a new ENV variable that is just a boolean like `LAMBDA_API=1` if that is preferable.

To test:
1. check out `feature/fix-for-lambda-api` locally
2. modify your .env file with `API_HOST_NAME=https://api-lambda.staging.notification.cdssandbox.xyz`
3. start the admin server: `make run-dev`
4. visit http://localhost:6012/ and verify that everything works


app/utils.py
```diff
@@ -71,6 +71,10 @@ def get_latest_stats(lang):
     sms_total = 0
     for line in results:
         date, notification_type, count = line
+        if "lambda" in current_app.config["API_HOST_NAME"]:
```
jimleroyer, line 74: I like this type of logic to be in a utility function of its own, we can have docstring that explains better why we need this, i.e. _For detecting if the API runs in an AWS Lambda environment and the need to apply custom logic_.

It might be clearer too to put the line 73 as part of the `if` or `else` statement to stress that one or the other will get executed.

smcmurtry: Good idea! I made those changes.

sastels (approved):
Tested in staging, both endpoints work for me. Pretty hacky though, and concerned that we don't know why it's different. I think before we switch prod to lambda we should figure this out, but for now 🦑

PR #1167, opened 2021-10-14
Fix remove pill border

amazingphilippe:
# Summary | Résumé

Some tweaks to set border-none to the pill component

After
Capture d’écran, le 2021-10-14 à 10 07 38
 Before
Capture d’écran, le 2021-10-14 à 09 54 16


amazingphilippe:
I don't see the same, what browser did you test with? Maybe there's a browser specific thing to consider

amazingphilippe:
ooh. Found the issue. Th double thick border only shows when focusing an element before a selected tab! Adding a bg color to selected tabs will fix this!

PR #1174, opened 2021-10-21
Update eventlet

jzbahrai:
# Summary | Résumé

Update `eventlet`


jzbahrai:
cant update this as gunicorn hasn't cut a new version with the evenlet changes:
https://stackoverflow.com/questions/58589138/error-class-uri-eventlet-invalid-or-not-found

PR #1185, opened 2021-11-04
descriptive link text for service settings

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/76

This improves screen reader experience by announcing what the "Change" links relate to. We still visually hide the setting name from the link as we don't have the same issue visually by displaying the setting name and value in a row. 

1. Any service. On the settings page
2. With screen reader, navigate through the "Change" links
3. Each should also read what the Change link will be changing. 

**Should we also add the value of the settings in these links? ** To be tested with users. 

      langues officielles?
      fonctionnalité existante?


amazingphilippe:
I may need help with the python tests:

1. Can I run them locally?
2. Why are we testing the value of the "Change" link?
3. Can I simply add what I am expecting at the end of each expected strings?

sastels:
> 1. Can I run them locally?
yes! Have you set things up locally as described [in the README](https://github.com/cds-snc/notification-admin#local-installation-instruction)? If so, if you've switched to the virtual env you should be able to `make test`.
I can help with the setup if there are issues (or if our setup instructions are out of date 😬 )

> 2. Why are we testing the value of the "Change" link?
That test is making sure that the user and admin settings pages have the right stuff on them, essentially. Basically it's just looking at the text that gets rendered on the screen and making sure nothing is missing.

> 3. Can I simply add what I am expecting at the end of each expected strings?
I would look at the errors, but if they make sense (ie "yeah, I changed the "Change" button there to "Change Service name") then yeah, just update the text in the test.

It's kind of hacky to have content in the test like this, but it has the advantage of making the test easy-ish to understand.

jimleroyer:
For the record, I spoke to @amazingphilippe a bit offline about his questions. Regarding question #3, I had this comment:

> Can I simply add what I am expecting at the end of each expected strings?

Yes we can, but we could also use a selector to target the screen-reader vs non-screen-reader text and make appropriate tests around these. 👍

iokpala (approved):
LGTM, Overall the Camel case appears inconsistent..

amazingphilippe:
> Camel case appears inconsistent

This is an extract of the text contents from every <tr> Every table cell will start with a capital, but can contain more than one word. 

The way I append the screen-reader-only text after "Change", I don't remove the capital case. But that text is hidden. 

I'm trying to work on better test to strip hidden text, then testing the remaining text contents!

PR #1195, opened 2021-11-09
adding the aria-required attributes to textbox component

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/66

Fixes https://github.com/cds-snc/notification-planning/issues/67

But also makes sure all textboxes have the aria-required attribute set to true as well. 

1. Inspect any form element. It should have both the `required` and `aria-required="true"` attributes. Both are important

I want to crowd source this. **Do we have any optional fields?**

# Todo

- [x] Go in and mark optional fields as optional. I found a few while casually reading forms.py

      langues officielles?
      fonctionnalité existante?


amazingphilippe:
Looking through forms.py, I have not seen any fields that should be optional. Therefore all should have the aria-required="true" attribute.

app/templates/components/textbox.html
```diff
@@ -55,6 +55,7 @@
       class=field_class,
       data_module='highlight-tags' if highlight_tags else '',
       rows=rows|string,
+      aria_required='false' if optional else 'true',
```
jzbahrai, line 58: given your latest comment should we change this?

amazingphilippe: Does it make more sense to only add this line if we need optional fields in the future?

amazingphilippe: But before we hardcode this, I'd like to know how all our fields end up with the "required" attribute. (different from aria-required="true")

jimleroyer: Approving the PR but hope we resolved this thread.

jimleroyer (approved):

app/templates/components/textbox.html
amazingphilippe: I removed the optional conditional.
1. We don't have optional fields using the textbox component
2. I forgot to add the "optional" parameter to the macro signature. 

It'll make this PR be simpler and we'll do a better implementation the day we'll decide to use textbox for optional fields.

PR #1200, opened 2021-11-18
menu bar (No-ARIA)

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/75

This PR adds in better keyboard support for the account menubutton.

**This PR isn't about ARIA anymore.**
**See reading list about how aria menus are bad for a11y.** 

**tl;dr: No ARIA is good ARIA** 

It **still** fixes the missing closing mechanisms: 
1. When a user clicks outside of the menu
2. Pressing esc works from anywhere, not just within the menu. 

1. While logged in
2. With your keyboard, toggle the menu button. The enter and space keys should toggle when focused on the button
3. Navigate through the menu items. Tab and shift-tab should allow ou to move forwards and backwards in the order of menu items
4. Pressing escape while focused on the menu items, on the menu button, or anywhere in the window should close the menu
5. Clicking outside of the menu items should close the menu. 
6. Clicking on the menu button should toggle the menu items visibility
7. Tabbing through the header without toggling the menu button should skip the menu items

- [ ] We probably want to add tests to make sure we don't break it in the future.
- [x] There is an issue with voiceover that I'm not sure we are causing. Menu items are read twice and simultaneously. Could be a VO bug too..

      langues officielles?
      fonctionnalité existante?


amazingphilippe:
This works as expected by ARIA, but I have fortunately stumbled upon vast amounts of critique of this behavior. Menu buttons and menubars strictly are meant for application like behaviour. What we have is not a menubutton, but a simple navigation that can be toggled. 

Sources

https://adrianroselli.com/2017/10/dont-use-aria-menu-roles-for-site-nav.html
https://adrianroselli.com/2019/06/link-disclosure-widget-navigation.html
https://marcus.io/blog/menu-or-not

In this last revision, I pulled out the menu and menuitem roles and toned down the keymapping back to what it was before. 
I did revise the markup to wrap the whole thing in a nav, and wrap the links in a unordered list.

I followed this example from inclusive components. https://inclusive-components.design/menus-menu-buttons/

app/assets/javascripts/menu.js
```diff
@@ -3,52 +3,119 @@
...
+    // Click anywhere but the menu items to close
+    $("body").click((e) => {
+      if (e.target != $items) {
+        close($menu, $items);
+      }
+    });
+
+    // Key handlers for toggle button
+    $menu.keydown((e) => {
+      let flag = false;
```
jimleroyer, line 78: Do we really need this flag? It seems it occurs in only one case.

app/assets/javascripts/menu.js
```diff
@@ -3,52 +3,119 @@
...
+        case "Escape":
+          close($menu, $items);
+          break;
+        default:
+          break;
+      }
+      if (flag) {
+        e.stopPropagation();
+        e.preventDefault();
+      }
```
jimleroyer, line 96: This could be a utility function if we reuse it elsewhere as in [line 109](https://github.com/cds-snc/notification-admin/pull/1200/files#diff-7aac0d62bb9e3d7b5f83f420cacda92dd22201248e8e29b4303512c7b3cec254R109).

app/templates/partials/nav/account_menu.html
```diff
@@ -1,60 +1,44 @@
...
-          type="button">
-      {{ _('Your account') }}
-      <img
-        alt="{{ _('Your account') }}"
-        aria-hidden="true"
-        class="account-menu-arrow ml-2 inline arrow self-center"
-        src="{{ asset_url('images/arrow-licorice.svg') }}"
-        width="15"
-      />
+<nav class="hidden md:visible md:flex relative justify-end inline-block text-left w-full text-blue visited:text-blue
```
jimleroyer, line 2: Cool to use `nav` instead.

app/templates/partials/nav/account_menu.html
```diff
@@ -1,60 +1,44 @@
...
+      border-transparent focus:outline-none focus:bg-yellow focus:border-blue hover:bg-gray-50 hover:underline"
+    id="account-menu"
+    data-module="menu"
+    data-menu-items="account-menu-options">
+      {{ _('Your account') }} 
+      <img alt="{{ _('Your account') }}" aria-hidden="true" class="account-menu-arrow ml-2 inline arrow self-center" src="{{ asset_url('images/arrow-licorice.svg') }}" width="15"/>
   </button>
-
-  <div id="account-menu-options"
+  <ul
```
jimleroyer, line 16: Is it better to use a list for the menu items?

app/assets/javascripts/menu.js
```diff
@@ -3,52 +3,119 @@
...
+    if (!this.hasFocus) {
       $items.toggleClass("hidden", false);
+      $items.removeAttr("hidden");
+      const $arrow = $menu.find(".arrow");
+      if ($arrow.length > 0) {
+        $arrow.toggleClass("flip", true);
+      }
+
+      $menu.attr("aria-expanded", true);
+      this.hasFocus = true;
```
jimleroyer, line 17: This is great to track this so behavior is improved. 👍

jimleroyer (changes requested):
**Breaking bug**

The mouse click on the menu is broken when in desktop view.

**Potential improvements** (issue also present in production env): 

Pressing the `escape` button once the menu is expanded should collapse the menu on desktop view. This works fine in mobile view it seems.

amazingphilippe:
> The mouse click on the menu is broken when in desktop view.

Oof missed this one!

app/templates/partials/nav/account_menu.html
amazingphilippe: If the CSS styles are disabled, lists are a good way to leverage HTML to organize the content of the page. Otherwise this has no impact on anything else!

app/assets/javascripts/menu.js
amazingphilippe: Maybe you can help me on this one. Is there a difference if this code is executed within the switch, or after? I'm not sure I'll be able to see the difference.

jimleroyer: I don't think there will be any difference if you'd put the flag code within the switch. Moving the code inline and with a function call would make it more readable though IMO.

jimleroyer:
Potential improvements: Pressing the escape button once the menu is expanded should collapse the menu on mobile view when no tab happened yet.

The switch code can be simplified as well with the usage of the flag which unnecessarily moves the logic outside of its contextual scope.

I will approve the PR though as this really improves the functionality of the menu overall. Phil and I can talk about next steps if he'd prefer to deal with these in another PR or now.

jimleroyer (approved):
Thanks for improving the functionality around this, especially to detect if it has current focus or not. 👍

amazingphilippe:
> The switch code can be simplified as well with the usage of the flag which unnecessarily moves the logic outside of its contextual scope.

Further work also revealed that the switch case isn't needed as space and enter are the default key mapping for a button element.

PR #1206, opened 2021-12-21
bug/Detect api lambda vs k8s better

sastels:
# Summary | Résumé

In #1126 we use the presence of "lambda" in `API_HOST_NAME` to detect whether we're connecting to the api lambda. This won't work in the canary (or after) when `api.staging.notification.cdssandbox.xyz` can go to either the lambda or or k8s cluster. We switch to a test that doesn't depend on the hostname.

Run locally with your `API_HOST_NAME` set to 
https://api.staging.notification.cdssandbox.xyz
https://api-k8s.staging.notification.cdssandbox.xyz
https://api-lambda.staging.notification.cdssandbox.xyz

and make sure you can still reach the home page in each case.

Why is there a difference? we never figured it out before.

      langues officielles?
      fonctionnalité existante?


app/utils.py
```diff
@@ -79,7 +71,7 @@ def get_latest_stats(lang):
     emails_total = 0
     sms_total = 0
     for line in results:
-        if is_lambda_api():
+        if isinstance(line, dict):  # the lambda returns info in a different format than the k8s cluster.
```
jimleroyer, line 74: I'd keep the `is_lambda_api` function as it is more telling and self-documenting code. 

How we implement detection then is encapsulated and hidden in the function itlself.

app/utils.py
```diff
@@ -63,12 +63,12 @@
...
 
 
-def is_lambda_api():
+def from_lambda_api(line):
     """
     We need to detect if we are connected to the lambda api rather than the k8s api,
     since some response data is different
     """
-    return "lambda" in current_app.config["API_HOST_NAME"]
+    return isinstance(line, dict)
```
jimleroyer, line 71: I think the detection of `AWS_LAMBDA_RUNTIME_API` env var would be better to detect ([replicating current strategy in our  lambda entrypoint shell script](https://github.com/cds-snc/notification-api/blob/dd1a62454618f840945ef5cc4f814dc0b6672417/bin/entry.sh#L2)), but if that works it's fine with me.

jimleroyer (approved):
LGTM

app/utils.py
sastels: The problem is that the admin could connect to either k8s or lambda - we'd need a way for the api to surface its `AWS_LAMBDA_RUNTIME_API` variable in the query. I figured this hack is fine for now since eventually it'll be 100% lambda and we can get rid of this check.

PR #1212, opened 2022-01-06
Fix the footer links and "top of page" link

pcraig3:
# Summary | Résumé

Footer links layout were breaking on mobile (too wide), but no longer. And then the "Top of page" link was using another background colour, so I changed it. All CSS changes, so no functional changes here.

Edit: made sure the hit state of the footer links was the minimum size of 44px ([according to WCAG](https://www.w3.org/TR/WCAG22/#target-size-enhanced))

| before | after |
|--------|-------|
|   Screen Shot 2022-01-06 at 14 42 37    |   Screen Shot 2022-01-06 at 14 42 43   |

### DIFF
--- a/app/templates/components/links.html
+++ b/app/templates/components/links.html
@@ -45,7 +45,7 @@
 {% macro nav_link_light(label="TODO: LABEL THIS LINK", href="/", class=None, external_link_img=None) %}
 <a
     href="{{ href }}"
-    class="p-2 text-blue text-small no-underline line-under border-b-2 border-transparent visited:text-blue link:text-blue hover:underline focus:border-blue focus:outline-none focus:text-blue decoration-clone flex items-center min-h-target {% if class %} {{ class }} {% endif %}"
+    class="p-2 text-blue text-small no-underline line-under border-b-2 border-transparent visited:text-blue link:text-blue hover:underline focus:border-blue focus:outline-none focus:text-blue decoration-clone min-h-target {% if class %} {{ class }} {% endif %}"
     {% if external_link_img %}
         target="_blank"
         aria-label="{{ label }} ({{ _('Opens in a new tab') }})"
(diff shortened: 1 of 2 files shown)


sastels (approved):
tested Heroku app on Chrome using dev tools to simulate phone screen, awesome! 👍

app/templates/components/links.html
```diff
@@ -45,7 +45,7 @@
 {% macro nav_link_light(label="TODO: LABEL THIS LINK", href="/", class=None, external_link_img=None) %}
 <a
     href="{{ href }}"
-    class="p-2 text-blue text-small no-underline line-under border-b-2 border-transparent visited:text-blue link:text-blue hover:underline focus:border-blue focus:outline-none focus:text-blue decoration-clone flex items-center min-h-target {% if class %} {{ class }} {% endif %}"
+    class="p-2 text-blue text-small no-underline line-under border-b-2 border-transparent visited:text-blue link:text-blue hover:underline focus:border-blue focus:outline-none focus:text-blue decoration-clone min-h-target {% if class %} {{ class }} {% endif %}"
```
amazingphilippe, line 48: `min-h-target` loses its effect here when we remove the`flex` class. Which simply applies display:flex. 

To fix the mobile issue, could we group the bullet + link instead? Like so:

```html
<div class="flex-shrink mr-5 flex">
  <img class="self-center lg:hidden inline mr-3">
          
  <a class="p-2 text-blue text-small no-underline line-under border-b-2 border-transparent visited:text-blue link:text-blue hover:underline focus:border-blue focus:outline-none focus:text-blue decoration-clone min-h-target " >
    Confidentialité
  </a>
</div>
```

Or any simpler solutions. Like a ::before or anything you can think of to keep the minimum target size!

pcraig3:
Thanks Steve! Merging now.

app/templates/components/links.html
pcraig3: Thanks for the suggestion, Phil! Added `flex` to the container divs.

PR #1213, opened 2022-01-06
Create external links by passing in a boolean rather than passing in an image

pcraig3:
External links have an icon beside them, a little square with an arrow coming out of it.

Screen Shot 2022-01-06 at 16 13 29

So far, in our macros we are usually (but not always) passing an image path for the "external.svg" icon when a link points somewhere external.

I say _usually_ because it changes sometimes. eg:
-    `is_external_link`: in the mobile header nav: [`nav_menu_item_mobile(documentation_url(), _('API documentation'), False, is_external_link=True)`](https://github.com/cds-snc/notification-admin/blob/78ec5496228d0e2d1ebb7654fa87fbb685105f6e/app/templates/partials/nav/gc_header_nav_mobile.html#L36)
-    `external_link_img`: in the regular header nav: [`nav_menu_item(documentation_url(),_('API documentation'),'header--'+header_navigation.is_selected('documentation'), external_link_img=asset_url('images/external.svg'))`](https://github.com/cds-snc/notification-admin/blob/78ec5496228d0e2d1ebb7654fa87fbb685105f6e/app/templates/partials/nav/gc_header_nav.html#L70)

`external_link_img` isn't really needed: the image path could easily be hard-coded in the macro itself (like [in the mobile header nav link macro](https://github.com/cds-snc/notification-admin/blob/c0b10714a598ab9c0f86239ca56dea165c61a493/app/templates/components/nav_menu_item_mobile.html#L19-L23)) and that makes the macro easier to use. It also makes it easier to build menus from pure data, incidentally.

There are no functional changes from this PR. It just simplifies these macros, and in the case of `secondary_button_transparent_white` removes this option altogether since no one is using it.

### DIFF
--- a/app/templates/components/buttons.html
+++ b/app/templates/components/buttons.html
@@ -36,24 +36,12 @@
 </a>
 {% endmacro %}
 
-{% macro secondary_button_transparent_white(label="TODO: LABEL THIS BUTTON", href="/", class=None, external_link_img=None) %}
+{% macro secondary_button_transparent_white(label="TODO: LABEL THIS BUTTON", href="/", class=None) %}
 <a
     href="{{ href }}"
     class="p-4 line-under text-base text-default underline text-white border-b-2 border-transparent link:text-white visited:text-white hover:no-underline focus:text-blue focus:outline-none focus:border-b-2 focus:border-white focus:no-underline {% if class %} {{ class }} {% endif %}"
-    {% if external_link_img %}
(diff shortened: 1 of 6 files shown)


amazingphilippe (approved):
Looks good to me! Although we might remove that icon if we end up with the API docs on the same site... in which case this refactor still works

pcraig3:
Awesome! Merging now

PR #1216, opened 2022-01-12
Add 2 new image domains to CSP: *.canada.ca and *.cdssandbox.xyz

pcraig3:
# Summary

Add 2 new `img-src` domains (our safelist for approved image domains) so that we can load images across environments if they are uploaded to GC Articles. 

## Description

We are planning to receive content from the GC articles API, including images. However, currently images can only come from `"self"` or the asset domain for a particular URL. 

So far we have been linking to images already hosted on Notify from GC Articles. For example, we can create an `` tag and reference an image on the Notify staging site. For example, we do this: ``

That works [on staging](https://staging.notification.cdssandbox.xyz/features-wp) but it doesn't work locally or on Heroku, and it won't work on the live domain.

This PR safelists 2 domains: `*canada.ca` and `*.cdssandbox.xyz`, so that we load images hosted between these domains. This makes building rich pages with GC Articles a lot easier, allowing us to upload images directly to GC Articles, or host images from on Notify domain to other Notify domains.

## Screenshots

| before | after |
|----------|-----------|
|   images fail to load on local dev and error in the console      |   images show up normally, no console error       |
|  Screen Shot 2022-01-12 at 11 31 50      |    Screen Shot 2022-01-12 at 11 32 17       |


amazingphilippe (approved):
Works very well locally!

pcraig3:
Thanks Phil!

PR #1210, opened 2022-01-05
Add feature to pull page content and navigation from GC Articles

timarney:
# Important 
This initial PR sets things up to transition pages incrementally to GC Articles.   This PR doesn't replace any existing pages or navigation. 

# Summary 

This pull request replaces the content for 2 static pages on the Notify homepage. Rather than replacing the pages outright, it **creates 2 new URLs for them**, so we can see them work side-by-side **without** causing a risk to Notify's existing workflows.  

The 2 new URLs are:
- a new "Features" page (`/features-wp`) 👉 https://notification-feature-he-8zanmf.herokuapp.com/features-wp
- a new "Why Notify" page (`/why-notify-wp`) 👉  https://notification-feature-he-8zanmf.herokuapp.com/why-notify-wp

There are 2 API requests being made:

1. Get all the content for the page
2. Get the menu items for the main nav and the dark blue footer

| current page (this is what is live right now) | page with content replaced using GC Articles API | page with content replaced using GC Articles API (colourized) |
|--------|-------|-------|
|  notification canada ca_features   |  localhost_6012_features  | localhost_6012_features copy |

## GC Articles

For this PR, we are pulling content from the staging environment for GC Articles — currently more stable than our "prod" environment. The URL for the staging environment is: https://articles.cdssandbox.xyz/notification-gc-notify/

You can see the pages we are re-creating here:
- https://articles.cdssandbox.xyz/notification-gc-notify/features
- https://articles.cdssandbox.xyz/notification-gc-notify/why-notify

The page content and the menu items are also editable in the GC Articles backend. Once you have an account, you can get in and change things.

| GC Articles Notify pages | URL (EN) | URL (FR) |
|----------|-----------|-----------|
|  Home page      |     https://articles.cdssandbox.xyz/notification-gc-notify/      | https://articles.cdssandbox.xyz/notification-gc-notify/fr/      |
|  Login Page       |    https://articles.cdssandbox.xyz/notification-gc-notify/sign-in-se-connecter/       |  (same)  |
|  All Pages      |   https://articles.cdssandbox.xyz/notification-gc-notify/wp-admin/edit.php?post_type=page    | (same) |
|  Edit "Features" page      |   https://articles.cdssandbox.xyz/notification-gc-notify/wp-admin/post.php?post=11&action=edit&lang=en        | https://articles.cdssandbox.xyz/notification-gc-notify/wp-admin/post.php?post=27&action=edit&lang=fr |
[...]

PR #1217, opened 2022-01-12
Add type-hinting to request_content

pcraig3:
# Summary | Résumé

This PR 
- adds type-hinting to the `request_content` method 
- manually raise a `NotFound` exception if we try to pull content from the API that doesn't exist

### Adds type-hinting

Following up on @jimleroyer's [review from yesterday](https://github.com/cds-snc/notification-admin/pull/1210#discussion_r782605719), this PR adds type-hinting to `request_content`. 

I discovered that pulling a page from the API, we get back a `list`, vs pulling a menu back we get a `dict`, so the type-hinting specifies that we can return `dict`, `list`, or `None`.

 ## raise `NotFound` exceptions manually

If we request a page that doesn't exist from the API, or a menu that doesn't exist, we get different results, but both are valid JSON responses. Since we don't want to treat these as successful responses, I added logic to ensure that our returned responses are actually valid, and raise a `NotFound` exception if not. If this exception is raised, it is caught lower in the function and either a cached response is returned or `None`.

### DIFF
--- a/app/main/views/index.py
+++ b/app/main/views/index.py
@@ -259,7 +259,7 @@ def page_content():
     response = request_content("wp/v2/pages", {"slug": slug, "lang": get_current_locale(current_app)})
 
     # show technical difficulties if no result
-    if response == "":
+    if response is None:
         abort(500)
 
     nav_url = "menus/v1/menus/notify-admin"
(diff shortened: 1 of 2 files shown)


jimleroyer (approved):
LGTM

PR #1218, opened 2022-01-14
Add route handling for preview pages

timarney:
Given we want to be able to load a page preview from GC Articles this adds the ability to render content from GC Articles via an page ID --- and the /preview path.
 
https://notification-feature-pr-o5vizv.herokuapp.com/preview?id=11
https://notification-feature-pr-o5vizv.herokuapp.com/preview?id=15

We need to make some updates on GC Articles to move this along further but this takes care of the loading + rendering of the content.

PR #1219, opened 2022-01-17
Add GC Articles JWT auth

dsamojlenko:
# Summary | Résumé

To enable Previews from GC Articles, will need to be able to pull Draft content from the API, which requires authentication.

To enable this, we've added JWT (Json Web Token) Authentication to GC Articles.

To interact with protected endpoints, the notify client will need to:
- Request a JWT from GC Articles using a dedicated user and generated Application Password
- On subsequent requests, pass the JWT as a Bearer Token
- When the JWT expires, request a fresh one for future requests


timarney:
Let's make some further updates on this after https://github.com/cds-snc/notification-admin/pull/1218 is in

We can add auth for only the preview routes + check for a logged in user to ensure previews don't become public

```
if current_user and current_user.is_authenticated:
```

app/articles/__init__.py
```diff
@@ -18,19 +23,58 @@ def set_active_nav_item(items=[], url="") -> None:
     for item in items:
         item["active"] = True if item["url"] == url else False
 
+def validate_token(token, auth_endpoint=GC_ARTICLES_AUTH_API_ENDPOINT):
+    base_endpoint = current_app.config["GC_ARTICLES_API"]
+    url = f"https://{base_endpoint}{auth_endpoint}/validate"
+
+    headers = {
+        'Authorization': ('Bearer ' + token)
```
timarney, line 31: I'm seeing a Type error here in VSCode might want to use this (as seen in some other Notify code)

```
 "Authorization": "Bearer {}".format(api_token)
```

app/articles/__init__.py
```diff
@@ -18,19 +25,58 @@ def set_active_nav_item(items=[], url="") -> None:
...
+
+    cache.set(GC_ARTICLES_AUTH_TOKEN_CACHE_KEY, parsed['token'], timeout=GC_ARTICLES_AUTH_TOKEN_CACHE_TTL)
+
+    return parsed['token']
 
 def request_content(endpoint: str, params={"slug": ""}) -> Union[dict, None]:
     base_endpoint = current_app.config["GC_ARTICLES_API"]
     lang_endpoint = ""
     lang = get_current_locale(current_app)
     cache_key = "%s/%s" % (lang, params["slug"])
```
timarney, line 66: Maybe not part of this PR but there's an issue here:

If we hit a valid URL i.e. 

http://localhost:6012/preview?id=261 --- if you enter a bad url after that http://localhost:6012/preview?id=2610000 you'll end up with the content cached from the original http://localhost:6012/preview?id=261

app/articles/__init__.py
```diff
@@ -18,19 +25,58 @@ def set_active_nav_item(items=[], url="") -> None:
...
 
+    token = authenticate()
+
+    headers = {
```
timarney, line 70: Same comment here:
https://github.com/cds-snc/notification-admin/pull/1219/files#r787160944

app/articles/__init__.py
```diff
@@ -6,7 +6,14 @@
 
 from app import cache, get_current_locale
```
jimleroyer, line 8: Didn't noticed you folks made an `articles` module, this is great. 🙏

app/articles/__init__.py
jimleroyer: We are trying to move more and more to [f-strings](https://realpython.com/python-f-strings/) for new code submissions, so slightly the same:
```
"Authorization": f"Bearer {api_token}"
```

app/articles/__init__.py
```diff
@@ -18,19 +25,58 @@ def set_active_nav_item(items=[], url="") -> None:
     for item in items:
         item["active"] = True if item["url"] == url else False
 
+def validate_token(token, auth_endpoint=GC_ARTICLES_AUTH_API_ENDPOINT):
+    base_endpoint = current_app.config["GC_ARTICLES_API"]
```
jimleroyer, line 29: Can we move the GC_ARTICLES_API env reading into a constant like those around lines 11-16? To have all the constants from env vars declared in the same place and have standalone functions relying less on these environment varas and more on parameters?

app/articles/__init__.py
```diff
@@ -18,19 +25,58 @@ def set_active_nav_item(items=[], url="") -> None:
...
+    if cache.get(GC_ARTICLES_AUTH_TOKEN_CACHE_KEY):
+        token = cache.get(GC_ARTICLES_AUTH_TOKEN_CACHE_KEY)
+        if validate_token(token):
+            return token
+
+    # Otherwise get a fresh one
+    res = requests.post(url=url, data={
+        "username": username,
+        "password": password
+    })
```
jimleroyer, line 54: What will happen in the call hierarchy if the request fails?

jimleroyer (commented):
This looks good overall. 

My main concern with these changes are what happens if the JWT token retrieval fails and the lack of tests around the new feature.

app/articles/__init__.py
timarney: Low risk here as this will only be used for /preview routes --- would need to check but I think this will result in a 500 at the moment

timarney: Will be even looks risk in the next PR --- we'll add that the user needs to be logged into Notify

timarney: Adding some error checking in a followup PR https://github.com/cds-snc/notification-admin/pull/1222/commits/b74f9531098a9dbea354bdf4c3a65c8785cd03a1#diff-426038280d05044cecc48d58f7e010751be4122863bbbaf69b40a6f458e3f61bR81

dsamojlenko:
Closing as this is superceded by #1222

PR #1220, opened 2022-01-18
Banner for page previews 

pcraig3:
# Summary | Résumé

Adds a "preview" overlay for previewing pages from the CMS. The two parts of this are:
1. a 'phase-banner' type component (similar to [the "Alpha" banner in the header of Forms](https://forms-staging.cdssandbox.xyz/))
2. A full-screen yellow border 

The design goal here is that users should _always_ know they are previewing a page, not seeing live content, but the 'preview' interface itself shouldn't interfere with the intended layout of the content. (You don't want a 'preview' page that doesn't look how the published page will look.)

Screen Recording 2022-01-19 at 09 58 30

## Features

- Only on pages that use the "/preview" sub-path, not used anywhere else.
- Contains an "edit" link (Note that this will 404 for anyone not logged into GC Articles)
- English and French
- Updates the page title in the browser tab so that it's abundantly clear when it's a preview

(Note that this isn't actually working for "draft" content: currently we are using the `/preview` url to preview already-published content.)

## To test this

- See what it looks like on the "Features" page preview here: https://notification-preview-ba-wfkr8y.herokuapp.com/preview?id=11
- If you're logged into WordPress, you can edit the page here: https://articles.cdssandbox.xyz/notification-gc-notify/wp-admin/post.php?post=11&action=edit&lang=en


app/translations/csv/fr.csv
```diff
@@ -1510,4 +1510,7 @@
 "Can see","Peut voir"
 "main footer","pied de page principal"
 "secondary footer","pied de page secondaire"
-"Select the {name} {type}","Sélectionner le {type} {name}"
\ No newline at end of file
+"Select the {name} {type}","Sélectionner le {type} {name}"
+"This is a preview of drafted content. Information on this page is subject to change.","Il s'agit d'un aperçu du contenu provisoire. Les informations contenues dans cette page sont susceptibles d'être modifiées."
```
amazingphilippe, line 1514: Just reworking this translation a little bit:

Cette page est un aperçu de contenu en mode brouillon. Les informations contenues dans cette page sont susceptibles de changer.

amazingphilippe (approved):
Just a suggestion for the banner content in French, otherwise it looks good to me!

app/translations/csv/fr.csv
pcraig3: Updated, thanks for that!

app/main/views/index.py
```diff
@@ -374,6 +375,12 @@ def page_content(path=""):
 
     if path == "preview":
         page_id = request.args.get("id")
+        # 'g' will set a global variable for this 1 request
+        g.preview = True
+        # URL for editing a page
+        g.preview_url = (
+            f"https://articles.cdssandbox.xyz/notification-gc-notify/wp-admin/post.php?post={page_id}&action=edit&lang=en"
```
timarney, line 382: We should use an environment variable here

pcraig3: Good point, adding in the `current_app.config["GC_ARTICLES_API"]` here since it already has the root URL we want.

PR #1221, opened 2022-01-19
Update: Switch from master to main branch

iokpala:
# Summary | Résumé

Going forward, we are adopting main as our default branch,

This commit updates the workflows YAML files, replacing master with  main branch

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -2,7 +2,7 @@ name: Build, push to AWS ECR, and deploy
 on:
   push:
     branches:
-      - master
+      - main
 
 env:
   AWS_REGION: ca-central-1
@@ -70,7 +70,7 @@ jobs:
         app_id: ${{ secrets.PRBOT_APP_ID }}
         private_key: ${{ secrets.PRBOT_PRIVATE_KEY }}
 
-    - uses: cds-snc/notification-pr-bot@master
+    - uses: cds-snc/notification-pr-bot@main
       env:
         TOKEN: ${{ steps.notify-pr-bot.outputs.token }}
 
--- a/.github/workflows/secret.yaml
+++ b/.github/workflows/secret.yaml
@@ -5,6 +5,6 @@ jobs:
     name: seekret-scanning
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@main
     - name: docker://cdssnc/seekret-github-action
       uses: docker://cdssnc/seekret-github-action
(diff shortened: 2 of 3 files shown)


jimleroyer:
Spoke with @iokpala who started on the effort. This is a low hanging fruit task to do in-between tasks, while waiting for new ones.

.github/workflows/docker.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
         app_id: ${{ secrets.PRBOT_APP_ID }}
         private_key: ${{ secrets.PRBOT_PRIVATE_KEY }}
 
-    - uses: cds-snc/notification-pr-bot@master
+    - uses: cds-snc/notification-pr-bot@main
```
sastels, line 73: This is referencing the pr-bot's default branch, which is `master`. ☝️  is a good idea though, just that we'll have to first change pr-bot so it has a `main` branch.

.github/workflows/secret.yaml
```diff
@@ -5,6 +5,6 @@ jobs:
     name: seekret-scanning
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@master
+    - uses: actions/checkout@main
     - name: docker://cdssnc/seekret-github-action
       uses: docker://cdssnc/seekret-github-action
```
sastels, line 10: I'm not sure if this "master" refers to the admin repo. 🤔  Wouldn't that mean that in CI we never checked the PR branch, just the default branch? @maxneuvians you set all this up, what's going on here?

sastels (commented):
I'm not sure if there's any changes we need before proceeding, just a couple things to work out first.

.github/workflows/docker.yaml
iokpala: Good catch.. Cheers Now has a `main`  branch.. https://github.com/cds-snc/notification-pr-bot/tree/main

.github/workflows/secret.yaml
iokpala: Good catch!! Cheers

sastels: Chatted with Max, the master there does refer to the admin repo. So we could switch to "main" to replicate the behaviour. BUT this is scanning the default branch, NOT the branch in this PR. So we should switch it to
```
- uses: actions/checkout@v2
```
and then we'd be scanning for secrets added in the PR (which is better than not knowing until the PR is merged)

jimleroyer (approved):

sastels:
Looks good! As mentioned above we should switch the seekret scanner to either be
```
    - uses: actions/checkout@main
```
to keep the original functionality
or 
```
    - uses: actions/checkout@v2
```
to do the scanning on the PR branch rather than the default branch.

sastels:
Note that merging this PR will probably trigger a deploy to staging (since we're now deploying when we merge to main)

iokpala:
@sastels noted

PR #1223, opened 2022-01-20
Homepage styles and editor experience

amazingphilippe:
# Summary | Résumé

This PR adds in styles to accommodate the home page design in GC Articles. It also uncovers interesting challenges with some WordPress blocks. Each have a workaround, but come with some trade-offs.

 ## Blocks
**Column block**: super nice to use when using percentage units. 
- Could benefit from some block settings to set max width within WordPress. 
- The workaround is to add a class using available tailwind sizes. I say available because we purge unused `max-w-x` declarations and their responsive variants. Therefore, using a max-width from WordPress also requires we safelist the declaration in the code. 

**Buttons**
- These are currently not useful because they wrap anchor tags in a div. It is also impossible to add classes to the anchor tag when it is an inner container. 
- The best workaround is to use custom html blocks. 

**Covers**
- Useful but are tricky to work with:
- Theming the covers is a bit fiddly. I've set up color themes in our css, but it requires that we pass a class through the advanced block settings. 

**Semantic html**
- lists and headings are not styled unless we provide the correct classes in the advanced block settings. I think this could be seen as desirable to offer as much control to the editor as possible. 

## WordPress API
- We need a way to show or hide the title on content coming from the API. This is currently only useful for the homepage. 
- We could use the slug to set a class on the main-content wrapper for content coming from WordPress. Something like this:
```html
<main role="main" class="wp-page-slug ...">
  ...
</main>
```
- Theme json file on the WordPress side to help theme cover blocks

1. Review [https://staging.notification.cdssandbox.xyz/](https://staging.notification.cdssandbox.xyz/) and [https://notification-cover-bloc-a9fyyy.herokuapp.com/preview?id=2](https://notification-cover-bloc-a9fyyy.herokuapp.com/preview?id=2) side by side
2. Review small, medium and large breakpoints

You'll notice some differences. Some of them are deliberate:
1. Hiding and showing images and headings as a reordering technique: Introduces too much complexity in the editor experience. Therefore the order on small screens will be different. 
2. Font: We could add the `font-sans` class to our headings in WordPress, but I didn't.
3. Spacing is slightly better on this new version on smaller screen. This is due to the simpler, non-reordering layout. 
4. Missing the live number section.


timarney:
@amazingphilippe 

If you can give me a list of colours 

hex + colour name I can setup a theme.json file.

amazingphilippe:
We keep our colours in [tailwind.config.js](https://github.com/cds-snc/notification-admin/blob/master/tailwind.config.js)

Would this be a one time thing? Or could we edit the json through WordPress too?

```js
colors: {
      red: {
        default: "#b10e1e",
        hover: "#990c1a",
        border: "#6a0812",
        mellow: "#df3034"
      },
      white: {
        default: "#FFF"
      },
      blue: {
        lighter: "#B2E3FF",
        default: "#26374A",
        border: "#1A3152",
        selected: "#75b9e0",
        lightblue25: "#d5e8f3",
        lightblue: "#0154B0",
        slightlight: "#284162"
      },
      gray: {
        default: "#eee",
        button: "#dee0e2", /* grey3 */
        selected: "#e1e4e7",
        hover: "#d0d3d6",
        border: "#b5babe",
        grey1: "#595959",
        grey2: "#bfc1c3",
        grey4: "#f8f8f8",
        lightgrey: "#C0C1C3",
        visitedlight: "#929AA4",
        visiteddark: "#C8CDD1"
      },
      yellow: {
        default: "#ffbf47"
      },
      green: {
        default: "#00672F",
        darker: "#00703C",
        hover: "#00692f",
        border: "#003618",
        green: "#006435"
      },
      black: {
        default: "#000"
      },
      transparent: {
        default: "transparent"
      },
}
```

timarney:
RE: buttons

This PR will filter the core/buttons block --- remove the div and add the class if applied to the a tag 

https://github.com/cds-snc/gc-articles/pull/478

timarney:
@amazingphilippe 

Added a starter theme.json file we can test with here

https://github.com/cds-snc/gc-articles/pull/480/files#diff-0cd7d1145ebdae78ef3bfd6200e4a651c469e002f44f1830aa5f238b02501bf0R13

It will be up in the next release in about 1hr including https://github.com/cds-snc/notification-admin/pull/1223#issuecomment-1020294270

jimleroyer:
Regarding the changes, I trust you more than I trust myself to evaluate these options. My main feedback, styling wise, is that it would be good to test all of our changes with accessibility in mind, i.e. with different browser font size default, just to see if that breaks the new wp styles, especially with gutters and percentage usage.

amazingphilippe:
> test all of our changes with accessibility in mind

I made several tweaks and tested using our a11y manual testing checklist. I fixed the issues that I could find. I think this is in a good spot to merge. It will be helpful for testing more content as it gets added. 

With this PR, the homepage should be considered equal or better than production (I fixed some spacing issues)

But! there are things to do still. Notably the live stats section. I think I want to tackle this in a new PR and keep this one as an initial commit for wp styles.

jimleroyer (approved):
LGTM

PR #1225, opened 2022-01-26
Language switcher for pages being returned from the API

pcraig3:
# Summary | Résumé

Gets the language switcher working properly for pages being returned from the API. 

There are 2 types of URLs that we need to account for: 
1. returning pages by slug
2. returning pages by id

There are also 2 scenarios here
1. Returning a page that _already has_ a translated page
2. Returning a page that _does not have_ a translated page 

This PR accounts for all of these cases. The only slightly weird one is if you are asking for a page by ID that doesn't have a translated ID, you get sent to a 404 page, and you have to use the "back" button in the browser to get back to where you were.

Also adds a bunch of tests for the new functions I've written.

### addendum 

Two more changes added in: 

1. Remove the black dashed border from "previews" to cut down on visual noise, and add a `z-index` so that the preview sits on top of fullscreen elements.
2. Add a `.wp` body class to API-driven pages, so that we can target elements like the page title or site header or footer if needed.

## gifs

<details open>
<summary>
<strong>Successfully going between pages that have EN and FR</strong>
</summary>

Screen Recording 2022-01-26 at 13 46 48

</details>

<details>
<summary>
<strong>Trying to get translations for an EN-only page</strong>
</summary>

Screen Recording 2022-01-26 at 13 52 11

</details>


timarney (approved):

jimleroyer (approved):
LGTM

PR #1227, opened 2022-02-03
Try again if API slug returns 404

pcraig3:
# Summary | Résumé

This pull request adds a fix for a problematic behaviour that @amazingphilippe brought up. 

Notify uses a language parameter in the user session, so this will change from person to person. It means that if _my_ browser is in French for Notify, and I send `/une-page` to Tim, whose browser is in English, he will get a `404 Not Found`, even though the page exists.

The logic in here is:
- when we ask for a page by slug
- _if_ it comes back empty
- try again with the other language
  - _if_ page found, return a string and redirect
  - _it_ *no* page found, 404 as normal

I also added a couple tests to make sure those steps happen, one for each case.

### DIFF
--- a/app/main/views/index.py
+++ b/app/main/views/index.py
@@ -371,7 +371,6 @@ def old_page_redirects():
 # --- Dynamic routes handling for GCArticles API-driven pages --- #
 @main.route("/<path:path>")
 def page_content(path=""):
-    nav_items = get_nav_items()
     page_id = ""
     auth_required = False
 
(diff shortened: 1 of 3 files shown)


app/main/views/index.py
```diff
@@ -386,11 +385,17 @@ def page_content(path=""):
 
     response = request_content(f"wp/v2/pages/{page_id}", {"slug": path}, auth_required=auth_required)
 
+    if isinstance(response, str):
```
jimleroyer, line 388: Can we add a one or two lines comment to say that a `str` returns means we are doing a language redirect?

I feel the `request_content` might get a bit complicated as it does different things and the return inform of what happened. The types of these returns aren't straighforward on what that means exactly, not without reading the function  itself and follow up code at least.

app/main/views/index.py
```diff
@@ -386,11 +385,17 @@ def page_content(path=""):
...
+        return redirect(response)
+
     if response:
         title = response["title"]["rendered"]
         slug_en = response["slug_en"]
         html_content = response["content"]["rendered"]
+
+        nav_items = get_nav_items()
```
jimleroyer, line 396: 👍

jimleroyer (approved):
LGTM -- although I would like a 1-2 lines comment / more documentation on the modified behavior of `request_content`.

timarney:
Yup looks good -- agree with @jimleroyer here https://github.com/cds-snc/notification-admin/pull/1227#discussion_r799761566

pcraig3:
Added a comment to clear it up. I think we're good here. 👍

PR #1228, opened 2022-02-04
added dynamic section at the end of homepage only

amazingphilippe:
# Summary | Résumé

1. This adds back the counter section to the homepage
2. Since the section comes before the CTA, I've made 2 partials that only display if `slug_en` equals "home" while rendering `page-content.html`
3. I've also removed the CTA from the content in GC Articles. If someone has an idea to insert a partial in the middle of the wordpress content (or better: feed dynamic values to the wordpress content), please share! I've tried the format and replace filters from jinja, but with unexciting results.

1. visit /home and /accueil on the deployment. Don't forget to toggle the language.


EliseKa (commented):
wow it looks so great! I wonder though, why are the numbers different than in production? Is it pulling numbers from staging instead of prod?

amazingphilippe:
> Is it pulling numbers from staging instead of prod?

That would be my guess!

pcraig3:
Added some CSS rules for lists in general, and then some homepage specific stuff. It's pretty subtle but it looks better on this page (and others).

Screen Shot 2022-02-04 at 13 36 13

app/main/views/index.py
```diff
@@ -398,6 +398,7 @@ def page_content(path=""):
             nav_items=nav_items,
             slug=slug_en,
             lang_url=get_lang_url(response, bool(page_id)),
+            stats=get_latest_stats(get_current_locale(current_app)),
```
jimleroyer, line 401: It seems that will call the get stats function on rendering of all pages, even these that do not require stats. Is my assumption wrong? If that is correct, could we avoid extra calls? There could be a short lived cache in the `get_latest_stats` function.

amazingphilippe: Your assumption is correct! Thanks for the function idea, I'll try that instead

app/assets/stylesheets/tailwind/toolkit/wp.css
```diff
@@ -105,22 +104,33 @@
...
+  .wp .page-content ul {
     @apply list list-bullet;
   }
 
-  .page-content blockquote {
+  .wp .page-content ul li {
+    @apply pt-4;
+  }
+
+  .wp .page-content blockquote {
```
amazingphilippe, line 115: The `.wp`class is too specific in here, I noticed it broke spacing between figures and ul on the smaller breakpoint. Was there a particular need for your changes to use .wp?

app/main/views/index.py
jimleroyer: Tell me if you need assistance on it.

app/assets/stylesheets/tailwind/toolkit/wp.css
pcraig3: Yes, we already discussed in slack, but putting here for visibility. I was noticing some of the generic element styles in "wp.css" were being overruled by more general-looking selectors that were above them in the file. I assumed that we wanted the styles targeting lists or quotes or whatever to be more specific so that they wouldn't get overwritten.  No worries though, seems like a mistaken assumption.

I've since reverted this change, so the top margins are back.

app/main/views/index.py
```diff
@@ -405,6 +405,7 @@ def page_content(path=""):
             nav_items=nav_items,
             slug=slug_en,
             lang_url=get_lang_url(response, bool(page_id)),
+            stats=get_latest_stats(get_current_locale(current_app)) if slug_en == "home" else None,
```
amazingphilippe, line 408: @jimleroyer I'm not sure if this is better. We'll still have a "stats" on every page even when we don't need it.

app/templates/partials/page-content/counter.html
```diff
@@ -0,0 +1,29 @@
+{% from "components/links.html" import content_link_light with context %}
+
+{% if stats %}
```
amazingphilippe, line 3: And here I check if stats is something. Not 100% sure of my correct use of `None`

app/main/views/index.py
pcraig3: If `stats` is None, it will be similar to if stats isn't set at all. If you are using an `if` check for it, both will return negative.

jimleroyer (approved):
LGTM -- after everyone's comments are addressed. 😄 Good work!

PR #1229, opened 2022-02-10
add Google Tag Manager to public facing pages

sastels:
# Summary | Résumé

Adds Google Tag Manager to public facing pages.

* Got rid of the `unsafe-inline` from the `script-src` by injecting a request nonce to propagate from the headers into the `<script/>` tags.
* Added Google Tag Manager domains in the content security policy tag to allow loading and execution of its resources.
* Added the `GOOGLE_ANALYTICS_ID` and `GOOGLE_TAG_MANAGER_ID` configuration variables. These could be sourced from the environment from an override but the default values are the ones we currently rely upon.
* Added IDs on elements to track as per the [GTM even tracking document](https://docs.google.com/document/d/1eB8efXRfJUhQWKrGK7YC0khaTloofGzmjfj3liR5Iwk/edit?pli=1#heading=h.jj00kvfr2rcm). **We will have to implement these IDs as well on the GCArticles** side as the staging environment will override the static content for which these changes apply.

**A few misc. items:**

* Sorted the configuration variables to easily have an overview. Deleted 1 or 2 that were not used anymore in the process.
* Removed the `interest-cohort` header that was meant to opt out of an intrusive Google advertising platform. Google deprecated this headers following some uproar. The tag was removed from Google Chrome support so it's now useless. We will have to keep an eye on this as Google is rebranding their effort and might come back with a similar product and opt-out mechanism (hopefully there will be one).

I could not get rid of `unsafe-eval` for the `script-src` policy unfortunately. Notify is making use of Javascript libraries that make use of it, such as for loading templates. We need to update these dependencies and hope that latest iteration of these scripts are using better javascript practices.

* Head toward the [Heroku instance link](https://notification-add-google-atjim1.herokuapp.com/).
* Fire up Google Tag Manager with the tags you have defined and the IDs that have been defined in the static template file. These IDs are all listed in [GTM even tracking document](https://docs.google.com/document/d/1eB8efXRfJUhQWKrGK7YC0khaTloofGzmjfj3liR5Iwk/edit?pli=1#heading=h.jj00kvfr2rcm).
* Test the tags from GTM using the Preview section.


jimleroyer:
Good idea to fill in a quick & dirty implementation for testing on a deployed Heroku.

app/config.py
```diff
@@ -15,55 +15,33 @@
 
 
 class Config(object):
-    LANGUAGES = ["en", "fr"]
```
sastels, line 18: Just sorting config keys in alphabetical order I assume?

app/__init__.py
```diff
@@ -247,25 +248,32 @@ def record_start_time():
...
 
 
-def convert_to_boolean(value):
-    if isinstance(value, string_types):
-        if value.lower() in ["t", "true", "on", "yes", "1"]:
-            return True
-        elif value.lower() in ["f", "false", "off", "no", "0"]:
-            return False
-
-    return value
```
sastels, line 268: I guess this function just isn't used anymore?

app/templates/components/buttons.html
```diff
@@ -28,16 +30,27 @@
...
 
-{% macro primary_button_blue_lighter(label="TODO: LABEL THIS BUTTON", href="/", class=None) %}
+{% macro primary_button_blue_lighter(
+    label="TODO: LABEL THIS BUTTON",
+    href="/",
+    class=None,
+    id_key=None
+) %}
 <a
     href="{{ href }}"
```
sastels, line 40: Do you need a `{% if id_key %}id="{{ id_key }}"{% endif %}` here?

app/templates/components/buttons.html
```diff
@@ -5,9 +5,11 @@
   lang=None,
   class=None,
   link_img=None,
-  link_img_class=None
+  link_img_class=None,
+  id_key=None
```
sastels, line 9: just curious, why call it `id_key` instead of `id`?

app/templates/components/buttons.html
jimleroyer: Oh yes, seems like I missed that one.

app/__init__.py
jimleroyer: No it wasn't, seems the configuration got some type checking/conversion a long time ago and it was dropped. I think we have a more solid solution with what @patheard brought.

app/config.py
jimleroyer: Yes, the only new environment vars are the two new `GOOGLE_*` around GTM and GA.

app/templates/components/buttons.html
jimleroyer: This might have caused some trouble in Python as the `id` is an existing function name, IIRC.

app/templates/components/links.html
```diff
@@ -42,7 +56,12 @@
...
 
-{% macro nav_link_light(label="TODO: LABEL THIS LINK", href="/", class=None, is_external_link=False) %}
+{% macro nav_link_light(
+    label="TODO: LABEL THIS LINK",
+    href="/", class=None,
+    id_key=None,
+    is_external_link=False
+) %}
 <a
     href="{{ href }}"
```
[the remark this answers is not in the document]
jimleroyer: Yes absolutely, thank you.

jimleroyer:
@sastels Added missing id_key output in components where it was missing and reworked logging level. I'll wait for your explicit approval before merging (as the roles are reversed here).

jimleroyer (approved):
Adding approval on Steve's behalf.

PR #1237, opened 2022-03-03
use correct key from tailwind tokens

amazingphilippe:
# Summary | Résumé

Missed an "s" when calling `theme("screens.lg")`


EliseKa (commented):
Looks good! No more extra white space. There's just the paragraph spacing that seems to change when narrowing the window - I don't know if it's expected when headings get resized?

PR #1240, opened 2022-03-07
Update GC Articles API integration caching

dsamojlenko:
# Summary | Résumé

Includes a major refactor of the original request_content function and wildcard/preview routes. The relevant functions were simply doing too much with too much branching logic.

Also converts the FlaskCache-based caching to RedisClient-based caching. (**NOTE**: This means you'll need redis running and configured for use when working locally)

For pages, there are two types of cache:
- Fallback - long-lived (7 days) cache in case GC Articles is not available
- Default - shorter (1 day) default cache to reduce latency from network requests to the GC Articles API

The TTL of these two types of cache can be adjusted as desired, as long as the fallback cache TTL is longer than the default.

- Does not cache `preview` endpoints.
- Menus may be cached for a longer period (vs content).

Also included the ability to clear all caches using the existing "Clear cache" form in Notify Admin.


timarney:
@dsamojlenko --- just getting this back on the radar .... conflict needs to be updated.

jimleroyer (commented):
Very nice refactoring and caching work. Have a few questions and started 1-2 threads, with potentially one necessary fix for the `None` response on invalid URL hits.

app/articles/api.py
```diff
@@ -0,0 +1,114 @@
...
+
+        # Long-term "Fallback" cache
+        if cacheable:
+            lang = get_current_locale(current_app)
+            slug = params.get("slug")
+            cache_key = f"{GC_ARTICLES_FALLBACK_CACHE_PREFIX}{endpoint}"
+            if slug:
+                cache_key += f"/{lang}/{slug}"
+            current_app.logger.info(f"Saving to cache: {cache_key}")
+            redis_client.set(cache_key, json.dumps(parsed), ex=GC_ARTICLES_FALLBACK_CACHE_TTL)
```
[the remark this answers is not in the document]
dsamojlenko: Great catch, yeah agreed that a None response shouldn't make it to the cache. Will push a commit this aft.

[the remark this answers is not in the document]
dsamojlenko: Here's where the fix for that ended up: https://github.com/cds-snc/notification-admin/pull/1240/commits/5edee7607504068f1e3ddf47126238d1991a3100

tests/app/articles/test_request_content.py
```diff
@@ -45,57 +43,14 @@ def test_request_content_fr(app_, mocker, capsys):
...
-            assert mock.call_count == 2
-            assert mock.request_history[0].url == f"{notify_url}?slug=la-page&lang=en"
-            assert mock.request_history[1].url == f"{notify_url}?slug=la-page&lang=fr"
-
-            # response should be url for the language switcher, as a string
-            assert isinstance(response, str)
-            assert response == "/set-lang?from=/la-page"
-
-
-def test_request_content_en_404_but_fr_doesnt_exist(app_, mocker, capsys):
```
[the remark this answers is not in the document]
dsamojlenko: Yeah, just not sure how to write tests around the http/routing layer.

tests/app/articles/test_fallback_cache.py
```diff
@@ -0,0 +1,75 @@
...
+        with requests_mock.mock() as mock:
+            """ ConnectionError during retrieval """
+            mock.get(notify_url, exc=requests.exceptions.ConnectionError)
+            
+            get_content("pages", {"slug": "mypage", "lang": "en"}, cacheable=True)
+
+            assert mock_redis_method.get.called
+            assert mock_redis_method.get.called_with('gc-articles-fallback--pages/en/mypage')
+            assert mock_redis_method.get('gc-articles-fallback--pages/en/mypage') == json.dumps(response_json)
+
```
jimleroyer: Nice set of tests!

tests/app/articles/test_pages.py
```diff
@@ -0,0 +1,67 @@
...
+import requests_mock
+from werkzeug.exceptions import Forbidden
+from app.articles import GC_ARTICLES_CACHE_PREFIX, GC_ARTICLES_FALLBACK_CACHE_PREFIX
+
+from app.articles.api import get_content
+from app.articles.pages import get_page_by_slug_with_cache
+from tests import MockRedis
+
+gc_articles_api = "articles.alpha.canada.ca/notification-gc-notify"
+notify_url = f"https://{gc_articles_api}/wp-json/pages"
```
jimleroyer: That might be better named gc_articles_url, as I was wondering why the mock was checking for the notify URL. 😅

tests/app/articles/test_pages.py
```diff
@@ -69,6 +70,32 @@ def test_get_page_by_slug_with_cache_miss_with_fallback(app_, mocker):
...
+
+            get_page_by_slug_with_cache(endpoint, params)
+
+            assert request_mock.called
+
+            assert mock_redis_method.get.called
+            assert mock_redis_method.get.call_count == 2
+            assert mock_redis_method.get.called_with(cache_key)
+            assert mock_redis_method.get.called_with(fallback_cache_key)
+            assert not mock_redis_method.set.called
```
jimleroyer, line 101: Thanks for including that one!

jimleroyer (approved):
LGTM -- just left a suggestion on var rename but all good.

app/main/views/index.py
```diff
@@ -368,44 +374,92 @@ def old_page_redirects():
...
+        response = get_page_by_slug_with_cache(endpoint, params=params)
 
-    if path == "preview":
-        if not request.args.get("id") or not current_user.is_authenticated:
-            abort(404)
+    if not response:
+        return _try_alternate_language(endpoint, params)
+
+    if not response:
+        abort(404)
```
andrewleith, line 418: Is this code reachable?

andrewleith (approved):
LGTM - one little comment below.  The feature works well, caches as expected and the routes fall through over to GCArticles when none are matched.  Great work!

@dsamojlenko As an aside, are there any guides or docs covering the steps for creating content in GCArticles that will then show up in GCNotify?

app/main/views/index.py
dsamojlenko: Good catch, you're right _try_alternate_language will abort(404) if there's still no response, so no need here.

dsamojlenko:
> LGTM - one little comment below. The feature works well, caches as expected and the routes fall through over to GCArticles when none are matched. Great work!
> 
> @dsamojlenko As an aside, are there any guides or docs covering the steps for creating content in GCArticles that will then show up in GCNotify?

I'm not sure if anyone else prepared something, but none that I'm aware of.

PR #1242, opened 2022-03-16
a11y fixes for textbox, textbox_search and live_search components. 

amazingphilippe:
# Summary | Résumé
Fixes [notification-planning/issues/475](https://github.com/cds-snc/notification-planning/issues/475)
Fixes [notification-planning/issues/481](https://github.com/cds-snc/notification-planning/issues/481)

- Removes nested items in textbox label
- Places hint and error message right after the label
- Adds an id to the hint and error message
- Adds an aria-describedby attribute to the label

- Look at the accessibility tree
- On any text input field:
- Make sure the label contains the hint or error message when relevant.
- Make sure a `<input>` does not have the "row" attribute
- Make sure that required and invalid attributes are set when relevant. 

Capture d’écran, le 2022-03-17 à 13 17 03
Capture d’écran, le 2022-03-17 à 13 19 29

# Reviewers needed

A dev
- [x] code review

A11y checks on a representative page ("create a template" is a good one) @adriannelee 
- [x] A11y tree review on inputs
- [x] Automated checks. (ARC toolkit, Axe or wave)
- [x] Voice over with safari

# Future work

Also sets the work for:
- [notification-planning/issues/477](https://github.com/cds-snc/notification-planning/issues/477) We can set whether a field is required or not. We just have to review where we have optional fields. Required is the default.


amazingphilippe:
I'm having trouble with wtforms or jinja. Not sure which. 

hen we call the input field we're just giving it every parameter we might need. this results in invalid markup, like "rows" on a text input, or aria-required on optional fields. 

```njk
{{ field(
      class=field_class,
      data_module='highlight-tags' if highlight_tags else '',
      rows=rows|string,
      aria_required='true',
      **kwargs
    ) }}
```

In this PR, I need to add aria-describedby, but only when it has a value. aria-describedby cannot be empty, we need to not have it at all. 

Is there a way to unroll an object into the field parameters? If the object is empty, it wont add any parameters. If the oibject contains data, they will be passed as key=value to the field.

amazingphilippe:
If anyone is curious, the answer lies here: https://wtforms.readthedocs.io/en/2.3.x/widgets/#widget-building-utilities

> In addition, the values True and False are special:
> - attr=True generates the HTML compact output of a boolean attribute, e.g. checked=True will generate simply checked
> - attr=False will be ignored and generate no output.

adriannelee (commented):
For the VoiceOver on Safari, everything seemed to be fixed. But, "This cannot be empty" repeats twice, once on the text and then again on the input field. Is that ok? @amazingphilippe

app/templates/components/live-search.html
```diff
@@ -14,7 +14,8 @@
             form.search,
             width='w-full',
             label=search_label,
-            autocomplete="off"
+            autocomplete="off",
+            **kwargs
```
sastels, line 18: where is `kwargs` coming from? Should it be in the `live_search` parameters?

amazingphilippe: In this instance, `kwargs` would come from when `live_search` is called. It is any keyword arguments that are given but unused. 

Then, since `live_search` only calls `textbox`, we don't have to be super explicit and repeat everything defined as defaults in `textbox`

That's my understanding, but I could be wrong. Here are some references:

- https://jinja.palletsprojects.com/en/2.11.x/templates/#macros
- https://wtforms.readthedocs.io/en/2.3.x/fields/#fields-with-custom-constructors

sastels: oh cool, I didn't know that! 💯

sastels (approved):
LGTM

yaelberger-commits:
Reviewed, approved just needs merging

PR #1246, opened 2022-03-22
fix: update to match notification-api devcontainer

patheard:
# Summary 
Use the same devcontainer base image and Dockerfile structure
as the [`cds-snc/notification-api` devcontainer](https://github.com/cds-snc/notification-api/tree/main/.devcontainer).

Also updates to use a single Docker container since the docker-compose
setup is no longer required (shared resources are now in the
notification-api devcontainer stack).


sastels (approved):
Works for me! thanks for fixing this up!

PR #1249, opened 2022-03-24
feat: add Config logging on startup

patheard:
# Summary
Add logging that outputs non-sensitive config to the logs when
the app is created.  This will allow us to validate that the app
is configured as expected.

# Related
* cds-snc/notification-planning#421
* cds-snc/notification-planning#503


app/__init__.py
```diff
@@ -192,6 +193,9 @@ def get_locale():
     # make sure we handle unicode correctly
     redis_client.redis_store.decode_responses = True
 
+    # Log the application configuration
+    application.logger.info(f"Notify config: {config.get_safe_config()}")
```
jimleroyer, line 197: Interesting, were we not logging it before?

app/config.py
```diff
@@ -115,6 +115,37 @@ class Config(object):
...
+        for attr in dir(cls):
+            attr_value = "***" if attr in sensitive_config else getattr(cls, attr)
+            if not attr.startswith("__") and not callable(attr_value):
+                config[attr] = attr_value
+        return config
+
+    @classmethod
+    def get_safe_config(cls) -> dict[str, Any]:
+        "Returns a dict of config keys and values with sensitive values masked"
+        return cls.get_config(cls.get_sensitive_config())
```
jimleroyer, line 147: IIRC, we did it in the API component as well? some of this code could be shared via notification-utils it appears to me, right?

tests/config/test_config.py
```diff
@@ -0,0 +1,51 @@
+import importlib
+import os
+
+import pytest
+
+from app import config
+
+
+@pytest.fixture
```
jimleroyer, line 9: Can we add these tests in a related test class? We started doing it that way. I know it's very small here but if that grows, our latest best practice will get followed at least.

tests/config/test_config.py
```diff
@@ -0,0 +1,51 @@
...
+
+    yield
+
+    os.environ = old_env
+    importlib.reload(config)
+
+
+def test_get_config(reload_config):
+    config.Config.API_HOST_NAME = "http://foo.bar"
+    config.Config.NOTIFY_USER_ID = "bueller"
```
jimleroyer, line 24: Like in Ferris? 😎

app/__init__.py
patheard: Yup, exactly - we only had it in API.

tests/config/test_config.py
patheard: image

app/config.py
patheard: Yup, that's a great idea.  Do you mind if we address in a future PR so we get the logging in place first?

tests/config/test_config.py
patheard: Ah, my bad - I didn't notice the existing `tests/app/test_config.py` file.  I've moved the new tests there.

jimleroyer (approved):
LGTM

PR #1250, opened 2022-03-28
Added request nonce to all script tags with inline code

jimleroyer:
# Summary | Résumé

Providing the request nonce to all scripts for allowing inline definition of scripts.

Get into the authenticated part of the application where the scripts are not loading.

### DIFF
--- a/app/templates/main_template.html
+++ b/app/templates/main_template.html
@@ -36,7 +36,7 @@
     property="og:image"
     content="{{ asset_url('images/product/notify-main.jpg') }}"
   />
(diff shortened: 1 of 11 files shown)


patheard (approved):
for science 🧪

iokpala (approved):
LGTM
Screenshot 2022-03-28 at 11 20 17

jimleroyer:
Tested on the heroku instance and many javascript errors are gone. Still see 2 js messages but it might be unrelated.

PR #1253, opened 2022-04-06
A11y/banner role

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/496

There are a few banners you can test with. Here is one:

1. On the API integration page
2. Navigate to keys
3. Create a testing key
4. When revoking a key, an alert banner should appear with an action button
5. Revoking the key also reveals a status banner
6. On the API integration page
7. Navigate to the safelist page
8. Enter invalid values in a number of fields
9. An error summary alert should appear after submit


andrewleith (approved):
Looks good!

✅  Voiceover correctly announces the warning banner when revoke API key button is pressed, and when validation triggers on the safelist page  
✅  `role="alert"` is now on the warning banner

PR #1255, opened 2022-04-07
Forced password reset flow

jzbahrai:
# Summary | Résumé

---

---


app/templates/views/new-password.html
```diff
@@ -18,6 +18,7 @@ <h1 class="heading-large">
       <p>
         {{ _('You can now create a new password for your account.') }}
       </p>
+
```
sastels, line 21: probably can get rid of this extra line? Unless we added on purpose.

andrewleith: Jumana removed it!

app/assets/stylesheets/tailwind/elements/details.css
```diff
@@ -45,9 +45,9 @@
   details .arrow {
     @apply text-small hidden;
   }
-
```
andrewleith: @jzbahrai  I've just added prettier to run on css files on save for the future :). Any other file types we run prettier on?  Or any other extensions like this I should install?

app/main/views/sign_in.py
```diff
@@ -36,6 +36,10 @@ def sign_in():
             "location": _geolocate_ip(get_remote_addr(request)),
         }
 
+        user = User.from_email_address_or_none(form.email_address.data)
+        if user and user.password_expired:
```
andrewleith, line 40: The way we wrote this logic means you can do this mini-login process with ANY password, i.e. no matter what you type in as a password for a user that is marked as having an expired password, the app will let you login and have the system send the reset email.  Is this what we want?

andrewleith: Probably related, there is also no explicit 2FA process here as we are short-circuiting the login process.  The user does get a reset email though, so perhaps that is sufficient?  This may warrant a discussion.

andrewleith (commented):
All flows work as designed:
- ✅  Admin user is now able to reset password
- ✅  Resetted user receives message upon login and is sent reset email
- ✅  User can not login with old password after this process is complete

A few notes commented below in the review:
- the expired user's old password is not verified, so they can technically enter anything and have a reset email sent to the email address in the db
- the way this feature short circuits the login process, there is no email code sent to the user (2FA). There is, however, the email reset link, so perhaps this suffices.
- Possibly out of scope of this PR, but you can use your previous password again when you reset your password.

app/main/views/forgot_password.py
```diff
@@ -29,3 +29,10 @@ def forgot_password():
         return render_template("views/password-reset-sent.html")
 
     return render_template("views/forgot-password.html", form=form)
+
+
+@main.route("/forced-password-reset/<email_address>", methods=["GET"])
+def forced_password_reset(email_address):
+    email_address = email_address.replace(" ", "+")
```
sastels, line 36: We should probably url encode the email address before we send it in, then we won't have to do this replace.

jzbahrai:
@andrewleith 

> All flows work as designed:
> 
> ✅ Admin user is now able to reset password
> ✅ Resetted user receives message upon login and is sent reset email
> ✅ User can not login with old password after this process is complete
> A few notes commented below in the review:
> 
> the expired user's old password is not verified, so they can technically enter anything and have a reset email sent to the email address in the db
This is 100p true but works exactly the same way as the "forgot password reset flow" and we agreed that as it has the same security concerns we are okay to build this feature in the same way. Good catch.
> the way this feature short circuits the login process, there is no email code sent to the user (2FA). There is, however, the email reset link, so perhaps this suffices.
Same as above.
> Possibly out of scope of this PR, but you can use your previous password again when you reset your password.
This is a good call but out of scope here. As we are changing the salt here and this is about the encrypted pass we are okay for this PR.

app/main/views/sign_in.py
jzbahrai: The second part is true and good catch, let me add that in.

amazingphilippe:
All looks good to me. Just a few comments

- I'm not receiving the correct email when an admin expires a password. I'm receiving an email with a password reset token. 
- The thing about expiring vs resetting a password. I've set up a little overview of what needs to be changed in Figma. Look for the pink outlines:
Capture d’écran, le 2022-04-11 à 11 02 29

- Missing the success banner in Admin panel after expiring a password


And then some nitpicks:

- Maybe take a look into [this PR for the details-summary](https://github.com/cds-snc/notification-admin/pull/1259) to see if we want to do (or not do) the changes in this PR instead. Or close the other PR and ignore this for now. @andrewleith 
- On the danger link styles: Their focus styles look different from the danger links we find on the template page. That's on me, I didn't look around when I styled them in the previous PR: 
Capture d’écran, le 2022-04-11 à 09 45 42

Capture d’écran, le 2022-04-11 à 10 41 50

app/main/views/forgot_password.py
sastels: ok, we tried this and it didn't work - let's keep it as it is 👍

sastels (approved):
I think this is good! Let's get it into staging and kick the tires a bit more.

amazingphilippe:
Chat with @jzbahrai 

- [ ] change the one "expiring" wording visible in the details-summary
- [x] The other "expiring" vocabulary not relevant in the admin feature, it won't work like the mass expiration. Same comment about the email it sends. 
- [ ] We can work to merge #1259 Shouldn't break anything here.
- [ ] Separate PR for the danger links
- [x] No success banner for the other admin danger links : No banner for the reset password link

PR #1259, opened 2022-04-08
detail summary component

amazingphilippe:
# Summary | Résumé

Adds a border on the expanded contents of details-summary components. 

We don't need this for the password reset feature, but it brings the summary border to all the components, regardless if they have a "details-body" class. For example in the admin panel:

Capture d’écran, le 2022-04-11 à 09 56 28

This would duplicate some of the work done in #1255. So we would need to carefully merge the CSS component

1. In admin
2. Open the "Apply filters" summary
3. See the border.


app/assets/stylesheets/tailwind/elements/details.css
```diff
@@ -42,7 +42,25 @@
...
+  details[open]:not(:last-child) {
+    @apply mb-gutterHalf;
+  }
+
   details .arrow {
     @apply text-small hidden;
   }
+
+  details > *:not(summary) {
+    padding-left: 25px;
```
andrewleith, line 54: Should there be a `border-left` rule here?

amazingphilippe: Possibly! I'm using a shadow on a pseudo element. Although a border would work nicely here. I'm second-guessing my pseudo styles. I don't remember why I did that...

amazingphilippe: I remember:

If direct children have margin, the border or shadow will skip the margin. So the pseudo trick makes sure the border is applied to the parent container, minus the height of summary (Which is a problem...)
Capture d’écran, le 2022-04-11 à 15 09 37
Capture d’écran, le 2022-04-11 à 15 12 57

amazingphilippe: So ultimately, I think the macro approach is probably the most robust approach. We can't get too clever with CSS on this one.

andrewleith: I'm not clear on what you're suggesting, do you want to continue with trying to target all child elements of `details` (i.e. `details >*`), or use a wrapper element/class?

If we want to continue with the child element targeting, I noticed a few gotchas:
- text elements aren't matched by this syntax (not that anyone should really do this..)
- inline elements float weirdly if they're first
- this technique tries style each child element, so it would need to override their styles, which may have unintended consequences (i.e. trying (but failing) to change the `button`'s padding value below)

image
```
<details open="">
    <summary>This is the title</summary>
    Random floating text
    <br />
    more floating text!
    <br />
    <button type="submit" style="margin-bottom:20px" class="button">Filter</button>
    <div>div tag</div>
    <p>p tag</p>
    <span>span tag</span>
 </details>
```

amazingphilippe: > do you want to continue with trying to target all child elements of details (i.e. details >*), or use a wrapper element/class?

I'm suggesting to use a wrapper element, in a reusable macro. For exactly the reasons you mentioned. 

And also because a pseudo element cannot know the height of the `summary` element. Therefore it's impossible to give the pseudo element a height value.

I'll make a new push and we can discuss

amazingphilippe:
So with more investigation, I found that that there is a polyfill for the details element that expects a div to wrap the content. It gives it an id of `details-content-x`

So I'm proposing we use this id to select the content wrapper and style it. 

See this file:
https://github.com/cds-snc/notification-admin/blob/b5019a4b5b641240cc9488da5bd019688ee8ba99/app/assets/javascripts/detailsPolyfill.js#L99-L107

andrewleith:
Interesting, yes that should work.  LGTM!

andrewleith (approved):
Good to go.

PR #1263, opened 2022-04-14
Update: Notify Security Headers

iokpala:
[Zenhub](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/96)

# Summary | Résumé

Notify's admin requires Referrer-Policy and Strict-Transport-Security, security headers respectively.

Within are 2 commits that add and test the aforementioned

---

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -616,6 +616,7 @@ def save_service_or_org_after_request(response):
 
 #  https://www.owasp.org/index.php/List_of_useful_HTTP_headers
 def useful_headers_after_request(response):
+    response.headers.add("Strict-Transport-Security", "max-age=63072000; includeSubDomains; preload")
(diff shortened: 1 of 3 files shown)


mohamed-cds (approved):
LGTM 🚀 ✅  when this lands in staging i'd verify analytics/metrics services especially if their endpoints are HTTP. the referrer policy will result in the `referer` header only containing the domain and will strip out path and query parameters. HTTPS is not impacted.

e.g. HTTPS (https://www.example.com/foo?bar=baz) will result in the referer of https://www.example.com if the next request is a HTTP endpoint. if it's a HTTPS endpoint the referer will be https://www.example.com/foo?bar=baz

jimleroyer (commented):
Good stuff @iokpala 👍

PR #1264, opened 2022-04-14
refactor how the email address is treated

sastels:
# Summary | Résumé

- store the email address in the browser session rather than the url
- only send the email if there is an email address of a user in the session and the user's password is expired.
- add more tests

Make sure the whole expire password flow still works.


andrewleith (approved):
Looks good:
- No longer able to spam reset emails
- Email address removed from query params

PR #1267, opened 2022-04-19
Fix/login event

andrewleith:
# Summary | Résumé

Create a LoginEvent when a user is forced to update their password, or requests a password reset, as this process automatically logs the user in. The user will see this even on the Sign-in History page when they log in.

See cds-snc/notification-planning#615 for more details including test plan.


jzbahrai (commented):
looks good to me but will also want a second review.

jzbahrai (approved):

app/main/views/authenticator.py
```diff
@@ -15,7 +15,7 @@ def __enter__(self) -> User:
 
         # Check if coming from new password page
         if "password" in session.get("user_details", {}):
-            self.user.update_password(session["user_details"]["password"])
+            self.user.update_password(session["user_details"]["password"], session["user_details"]["loginData"])
```
sastels, line 18: Should also check if `loginData` is a key in `session.get("user_details", {})`

sastels (changes requested):
Code looks good with one minor request. Haven't tested it yet, will do that ASAP.

app/main/views/authenticator.py
andrewleith: Done.

sastels (approved):
Tested locally, verified that login_events table is getting new rows for forced password reset. Good job!

jimleroyer (commented):
Great job @andrewleith ! 👏

PR #1272, opened 2022-04-28
feat: PR review environment

patheard:
# Summary
Creates a PR review environment using Lambda and function URLs.
Each PR against the `main` branch will create a unique review
environment hooked up to the Staging API.

# Related
* cds-snc/notification-planning#625

### DIFF
--- a/.github/workflows/test-admin-remove.yaml
+++ b/.github/workflows/test-admin-remove.yaml
@@ -7,7 +7,8 @@ on:
     types: [closed]
 
 env:
-  FUNCTION_NAME: "notify-admin"
+  AWS_DEFAULT_REGION: ca-central-1
+  FUNCTION_NAME: "notify-admin-pr"
   GITHUB_SHA: ${{ github.sha }}
   IMAGE: notify/admin
   REGISTRY: 239043911459.dkr.ecr.ca-central-1.amazonaws.com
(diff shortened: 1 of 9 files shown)


jzbahrai (approved):

sastels:
The only thing is that we do use redis in admin, for the initial sign on flow and for the template reviews. Is there any way to hook this up to redis somewhere?

sastels (approved):
So awesome!!!!

.github/workflows/test-admin-remove.yaml
```diff
@@ -0,0 +1,42 @@
+name: Remove test admin deployment
+
+on:
+  pull_request:
+    branches:
+      - main  
+    types: [closed]
```
jimleroyer, line 7: That's very handy, to remove a preview app on closed PR. 

We got a few Snyk PR getting opened though and I wonder how many images/lambdas we'll have deployed, considering we are not disciplined at the moment (we plan to change that). Could we apply a filter? or a time based deletion of the lambdas? (which then might require a manual run of the workflow action if we want to spin one up again).

jimleroyer (commented):
Nicely done @patheard

.github/workflows/test-admin-remove.yaml
patheard: Yeah, I've been thinking about that idea that the environments are deleted after a certain amount of time and if you need them back, the workflow can just be manually re-run.

patheard: There, added this feature in #1274

PR #1275, opened 2022-05-03
Load GC Articles provided content but keep named routes

dsamojlenko:
# Summary | Résumé

Swap out static pages for GC Articles provided content, preserving named routes. 

The simpler version of this would be to delete the routes and just let the path pattern fall through to the wildcard route, but in some cases we want to preserve the named routes for use in templates. 

To preserve these named routes, we'll just leave them in place but use the page_content() method of the wildcard route to render the page.


amazingphilippe:
Just fixed a few things. Here's the list:

1. Removed conflicting named routes
2. replaced `url_for()` with `_()` and added translations where routes were referenced. 
3. Fixed the personalisation-guide fr redirect

Just realized we did not need to update all `url_for`s. When the routes are redirecting, we can use the named routes just fine.

app/main/views/index.py
```diff
@@ -95,11 +89,6 @@ def verify_mobile():
     return render_template("views/verify-mobile.html")
 
 
-@main.route("/privacy")
```
andrewleith, line 98: Why did you remove some routes entirely to allow pass-through to GCArticles (like this one), but then modify others to explicitly check the language (like `/format`)?  What is the difference with these two approaches?

app/templates/partials/nav/footer.html
```diff
@@ -14,8 +14,8 @@
               {% endif %}
             {% endfor %}
           {% else %}
-            {{ nav_link_dark(id_key='nav-footer-primary-why-notify', label=_('Why GC Notify'), href=url_for('main.why-notify'), class="-ml-2") }}
-            {{ nav_link_dark(id_key='nav-footer-primary-features', label=_('Features'), href=url_for('main.features'), class="-ml-2") }}
+            {{ nav_link_dark(id_key='nav-footer-primary-why-notify', label=_('Why GC Notify'), href=_('/why-notify'), class="-ml-2") }}
```
andrewleith, line 17: This seems to work well for app-based pages.  However, when I am viewing static content on GCArticles, the french menus all have English URLs.  I tried to update them but it didn't seem to work.

andrewleith (commented):
Overall this looks good.  I am curious what @jimleroyer thinks about having 2 ways to link to URLs going forward:
1. For app pages, we would continue to use`url_for()` as usual
2. For static content served up from GCArticles, we would create translations, and link directly using `<a href="_('/some-path-in-translation-file')">...</a>`

app/main/views/index.py
amazingphilippe: That's a good point.
- I left the `/format` route and alike because we were forwarding it to a new path in both languages: `/formatting-guide` and `guides-reference`. My rationale is that we might want to keep these ones to redirect users if we forget a link somewhere. Basically preventing link rot. 
- I removed the `privacy` routes because they were causing problem when linking from FR pages. Instead of adding a new route, or doing something hacky in the route, we decided to remove it entirely and use relative path strings that we would manually translate.

app/templates/partials/nav/footer.html
amazingphilippe: You're right, I can see the French menu in Articles, but it doesn't seem to get picked up on fr pages. This might already be an existing issue

app/main/views/index.py
andrewleith: Gotcha - thanks for the explanation.

.github/workflows/test-admin-deploy.yaml
```diff
@@ -132,7 +132,8 @@ jobs:
                 REDIS_ENABLED=False,\
                 SENDING_DOMAIN=staging.notification.cdssandbox.xyz,\
                 API_HOST_NAME=https://api.staging.notification.cdssandbox.xyz,\
-                ADMIN_BASE_URL=$URL
+                ADMIN_BASE_URL=$URL,\
+                GC_ARTICLES_API=articles.alpha.canada.ca/notification-gc-notify
```
jimleroyer, line 136: Spoke with Andrew, we'll undo this and go through the manifest .env files.

app/main/views/index.py
```diff
@@ -48,14 +49,8 @@ def index():
     if current_user and current_user.is_authenticated:
         return redirect(url_for("main.choose_account"))
 
-    return render_template(
-        "views/signedout.html",
```
jimleroyer, line 52: One thing I just noticed is that we changed the hierarchy of the template. Initially, the notify_template.html was the base template for public pages and the main_template.html was the one for authenticated pages. It seems we switched to the latter, following the call hierarchy and template called by `page_content`. 

I wonder what we lose an gain from switching to the main_template.html file. I know that @andrewleith had to add back some metatags and the hidden ascii art easter egg already.

image

app/main/views/index.py
```diff
@@ -71,7 +66,7 @@ def robots():
 
 @main.route("/.well-known/security.txt")
 def security_txt():
-    security_policy = url_for("main.security", _external=True)
+    security_policy = url_for("main.index", _external=True).rstrip("/") + _("/security")
```
jimleroyer, line 69: Maybe one slight improvement would be to have a util function in the `articles` module, something called `gc_url_for` for example that we could call this way:

```python
gc_url_for('/security')
```

app/main/views/index.py
```diff
@@ -374,6 +318,16 @@ def old_page_redirects():
     return redirect(url_for(request.endpoint.replace("redirect_", "")), code=301)
 
 
+# --- GCA Redirects --- #
+@main.route("/a11y", endpoint="/accessibility")
+@main.route("/why-notify", endpoint="/why-gc-notify")
+@main.route("/personalise", endpoint="/personalisation-guide")
+@main.route("/format", endpoint="/formatting-guide")
+@main.route("/messages-status", endpoint="/formatting-guide")
```
jimleroyer, line 326: Note: message-status is redirecting to formatting-guide, check if that's right.

app/main/views/index.py
```diff
@@ -374,6 +318,16 @@ def old_page_redirects():
...
+def gca_redirects():
+    return redirect(_(request.endpoint.replace("main.", "")), code=301)
```
jimleroyer, line 328: Nice handy function!

app/templates/content_template.html
```diff
@@ -6,6 +6,22 @@
...
+      <meta name="keywords" content="{{ keywords }}"/>
+      <meta property="og:title" content="{{ meta_title }}"/>
+      <meta property="og:description" content="{{ description }}"/>
+      <meta
+        property="og:image"
+        content="{{ asset_url('images/product/notify-main.jpg') }}"
+      />
+      <meta property="og:url" content="{{ admin_base_url }}"/>
+    {% endif %}
+  {% endblock %}
```
jimleroyer, line 24: If we'd be using the notify_template.html file as the base hierarchy instead of main_template.html, filling in the gap with the meta might not be necessary.

app/translations/csv/fr.csv
```diff
@@ -1561,4 +1561,14 @@
...
+"/privacy","/confidentialite"
+"/security","/securite"
+"/why-gc-notify","/pourquoi-gc-notification"
+"/features","/fonctionnalites"
+"/guidance","/guides-reference"
+"/accessibility","/accessibilite"
+"/terms","/conditions-dutilisation"
+"/formatting-guide","/guide-mise-en-forme"
+"/personalisation-guide","/guide-personnalisation"
+"/message-delivery-status","/etat-livraison-messages"
```
jimleroyer, line 1574: Moving to a hypothetical `gca_url_for` function might remove the need to be in the translation file, but it would require a dictionary making this mapping. It would move content related translations from bilingual URLs. I wonder if placing such mapping in the navigation.py might be better. /brainstorming

tests/app/articles/test_pages.py
```diff
@@ -99,3 +100,49 @@ def test_bad_slug_doesnt_save_empty_cache_entry(app_, mocker):
...
+    """
+    This test ensures that the pages we renamed properly provide a permanent redirect, and exist in GCA
+    """
+
+    # ensure each url is a permenent redirect
+    client_request.get_url(url, _expected_status=301)
+
+    # ensure target location exists and returns content
+    page = client_request.get_url(url, _follow_redirects=True)
+    assert page.find("span", id="gc-title").text.strip() == "GC Notify"
```
jimleroyer, line 116: Very nice. 👍

tests/app/main/views/test_index.py
```diff
@@ -96,7 +96,8 @@ def test_security_txt(client):
     response = client.get(url_for("main.security_txt"))
     assert response.headers["Content-Type"] == "text/plain"
     assert response.status_code == 200
-    security_policy = url_for("main.security", _external=True)
+    security_policy = url_for("main.index", _external=True).rstrip("/") + "/security"
```
jimleroyer, line 99: We could use the hypothetical `gca_url_for`.

tests/app/main/views/test_index.py
```diff
@@ -136,6 +128,29 @@ def test_static_pages(
...
+        "/message-delivery-status",
+    ],
+)
+def test_static_pages_gca(
+    client_request,
+    mock_get_organisation_by_domain,
+    url,
+):
+    page = client_request.get_url(url, _follow_redirects=True)
+    assert not page.select_one("meta[name=description]")
```
jimleroyer, line 151: ♥

app/__init__.py
```diff
@@ -508,7 +507,9 @@ def format_notification_status_as_field_status(status, notification_type):
 
 
 def format_notification_status_as_url(status, notification_type):
-    url = partial(url_for, "main.messages_status")
+    # url = partial(url_for, "main.messages_status")
```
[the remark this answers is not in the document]
andrewleith: Done.

tests/app/main/views/test_index.py
andrewleith: Done.

.github/workflows/test-admin-deploy.yaml
andrewleith: Done.

app/main/views/index.py
andrewleith: Done.

app/templates/content_template.html
andrewleith: Will look into this next.

app/main/views/index.py
andrewleith: Fixed.

app/templates/partials/nav/footer.html
andrewleith: I think we are all set on this now, menus are now working as expected.

app/main/views/index.py
andrewleith: @jimleroyer  Digging a little further, it looks like only `signedout.html` used `notify_template.html` - every other page used `admin_template.html` even pages that are not signed in.  We are not introducing anything new here in regards to templates inheritance.

andrewleith:
Looking into getting the ASCII art working again on the home page, I noticed the following condition in `app/templates/partials/cds_ascii_art.html`:

```
{% if is_cds_domain or is_review_app %}
```

Not sure why we need that, but it just means you won't see it when running locally.  I've left it there for now.

tests/app/articles/test_pages.py
```diff
@@ -99,3 +100,49 @@ def test_bad_slug_doesnt_save_empty_cache_entry(app_, mocker):
...
+    """
+    This test is ensures the content we link to exists on GCA.  If someone removes a page on GCA, this will fail.
+    """
+
+    # ensure target location exists and returns content
+    page = client_request.get_url(url, _expected_status=200)
+    assert (
+        page.find("span", id="gc-title").text.strip() == "GC Notify"
+        or page.find("span", id="gc-title").text.strip() == "GC Notification"
+    )
```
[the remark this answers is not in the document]
andrewleith: Yes, probably.  Shall I remove these tests?

app/articles/routing.py
```diff
@@ -0,0 +1,44 @@
+from flask import current_app, url_for
+
+from app.articles import get_current_locale
+
+GC_ARTICLES_ROUTES = {
+    "home": {"en": "/home", "fr": "/acceuil"},
```
jimleroyer, line 6: Is there a typo in `accueil` or it's really written this way in the app?

jimleroyer:
```suggestion
    "home": {"en": "/home", "fr": "/accueil"},
```

andrewleith: That is a typo.  I will fix and add a test to verify the URLs in this dict 😊

tests/app/main/views/test_index.py
```diff
@@ -96,7 +97,7 @@ def test_security_txt(client):
     response = client.get(url_for("main.security_txt"))
     assert response.headers["Content-Type"] == "text/plain"
     assert response.status_code == 200
-    security_policy = url_for("main.index", _external=True).rstrip("/") + "/security"
```
jimleroyer: 👍

.github/workflows/test-admin-deploy.yaml
```diff
@@ -132,8 +132,7 @@ jobs:
                 REDIS_ENABLED=False,\
                 SENDING_DOMAIN=staging.notification.cdssandbox.xyz,\
                 API_HOST_NAME=https://api.staging.notification.cdssandbox.xyz,\
-                ADMIN_BASE_URL=$URL,\
-                GC_ARTICLES_API=articles.alpha.canada.ca/notification-gc-notify
```
jimleroyer: 💥

jimleroyer (commented):
There might be a typo about accueil but else it looks excellent.

tests/app/articles/test_pages.py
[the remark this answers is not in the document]
jimleroyer: We could add an annotation for future usage and not lose track of these, so we could do something along the lines of what we did for [concurrent vs serial tests](https://github.com/cds-snc/notification-api/blob/ff0b04a435bde73bdf4815c06292a9da439fca82/scripts/run_tests.sh#L43-L48). But we'd just mark these for now without taking more actions on how to run. Pytest documentation: https://docs.pytest.org/en/6.2.x/mark.html

app/main/views/index.py
jimleroyer: Thanks @andrewleith !

tests/app/articles/test_pages.py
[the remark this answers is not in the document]
andrewleith: Are you suggesting marking them as `skip` - or marking them as integration tests and letting them continue to run?

app/articles/routing.py
andrewleith: Fixed.

jimleroyer (approved):
LGTM! :shipit:

PR #1276, opened 2022-05-19
fix: PR review environment function concurrency

patheard:
# Summary
Add Lambda reserved concurrency to the Notify Admin PR review
environment function.  This is to prevent Denial of wallet attacks.

Also updates the function role ARN to use the IAM role managed
by the `notification-terraform` project.

# Related
* cds-snc/notification-terraform#482


jimleroyer (approved):
👍

PR #1278, opened 2022-05-24
Update GC Articles tests

andrewleith:
# Summary | Résumé

As Jimmy pointed out, some of the tests are actually integration tests that reach over to GC articles.  This PR:
-  marks these as `skip` for now until we have a proper place to run integration tests
- fixes some of the other tests that were mistakenly calling out to GCA and instead mocks those calls

### DIFF
--- a/tests/app/articles/test_pages.py
+++ b/tests/app/articles/test_pages.py
@@ -113,36 +113,3 @@ def test_gca_redirects_work(client_request, mocker, url):
 
     # ensure each url is a permenent redirect
     client_request.get_url(url, _expected_status=301)
-
-
-@pytest.mark.skip(reason="This is an integration test, should be elsewhere")
-@pytest.mark.parametrize(
-    "url",
-    [
-        "/why-gc-notify",
-        "/features",
-        "/guidance",
-        "/security",
-        "/privacy",
-        "/accessibility",
-        "/terms",
-        "/pourquoi-gc-notification?lang=fr",
-        "/fonctionnalites?lang=fr",
-        "/guides-reference?lang=fr",
-        "/securite?lang=fr",
-        "/confidentialite?lang=fr",
-        "/accessibilite?lang=fr",
-        "/conditions-dutilisation?lang=fr",
-    ],
-)
-def test_gca_content_exists_en_fr(client_request, mocker, url):
-    """
-    This test is ensures the content we link to exists on GCA.  If someone removes a page on GCA, this will fail.
-    """
-
-    # ensure target location exists and returns content
-    page = client_request.get_url(url, _expected_status=200)
-    assert (
-        page.find("span", id="gc-title").text.strip() == "GC Notify"
-        or page.find("span", id="gc-title").text.strip() == "GC Notification"
-    )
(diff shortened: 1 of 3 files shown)


tests/app/articles/test_routing.py
```diff
@@ -29,11 +29,13 @@ def test_gca_url_for_creates_asbolute_url(app_, mocker, route, lang, expectedURL
     assert route == expectedURL
 
 
+@pytest.mark.skip(reason="This is an integration test, should be elsewhere")
 @pytest.mark.parametrize("route", list(GC_ARTICLES_ROUTES))
 def test_ensure_all_french_gca_routes_in_GC_ARTICLES_ROUTES_exist(client_request, mocker, route):
 
     mocker.patch("app.articles.routing.get_current_locale", return_value="fr")
     mocker.patch("app.main.views.index.get_current_locale", return_value="fr")
+    mocker.patch("app.main.views.index.get_page_by_slug", return_value={"some": "value"})
```
sastels, line 38: Do we need this mock if we're skipping the test? With this mock the test wouldn't really be testing anything I think?

tests/app/articles/test_routing.py
```diff
@@ -42,11 +44,13 @@ def test_ensure_all_french_gca_routes_in_GC_ARTICLES_ROUTES_exist(client_request
     assert render_article.called
 
 
+@pytest.mark.skip(reason="This is an integration test, should be elsewhere")
 @pytest.mark.parametrize("route", list(GC_ARTICLES_ROUTES))
 def test_ensure_all_english_gca_routes_in_GC_ARTICLES_ROUTES_exist(client_request, mocker, route):
 
     mocker.patch("app.articles.routing.get_current_locale", return_value="en")
     mocker.patch("app.main.views.index.get_current_locale", return_value="en")
+    mocker.patch("app.main.views.index.get_page_by_slug", return_value={"some": "value"})
```
sastels, line 53: same question as above.

sastels (approved):
LGTM just a couple questions about whether we should add those mocks to the skipped tests/

tests/app/articles/test_pages.py
```diff
@@ -105,17 +105,17 @@ def test_bad_slug_doesnt_save_empty_cache_entry(app_, mocker):
...
+
     # ensure each url is a permenent redirect
     client_request.get_url(url, _expected_status=301)
 
-    # ensure target location exists and returns content
-    page = client_request.get_url(url, _follow_redirects=True)
-    assert page.find("span", id="gc-title").text.strip() == "GC Notify"
-
 
+@pytest.mark.skip(reason="This is an integration test, should be elsewhere")
```
jimleroyer, line 118: Can we mark these with an integration tag? Do we want to skip these because they will fail more often than we'd like?

tests/app/articles/test_routing.py
andrewleith: I think i did this when I was trying to see if I could not call out to GCA - but in the end the test is pointless if it doesn't call out to GCA.

tests/app/articles/test_pages.py
andrewleith: Yes, I will mark them with an integration tag. 

I think I misunderstood your earlier comments, so I marked these as skipped.  I didn't anticipate them failing at all 
- unless GCA went down.  I am still not clear why its returning a 404 sometimes here.

As discussed yesterday, maybe we should give them one more try now that we are pointing to PROD GCA and see if its less flaky?

tests/app/articles/test_routing.py
andrewleith: Fixed.

andrewleith:
After discussion with Jimmy, I'm going to let the tests simmer for a few days as-is now that the the correct URL for GCA is being used in staging.  I think (hope?) the failures may have been due to us pointing to their staging environment.

Looking closer at the failures, it seems to happen on a schedule: every night around 10:40EST, which is a bit odd.

PR #1280, opened 2022-05-26
Gc articles bugfixes

andrewleith:
# Summary | Résumé

Bug fixes for the test/hack session.

## QA
- [x] relative links used on some pages (why-notify, features)
- [x] Links in formatting guide should not leave site
- [x] "Why notify" menu item not active when on page
- [x] Image on `/message-delivery-status` pointing to staging
- [x] "Your account" menu going underneath blue hero banner on home page
- [x] Mobile language switch not working
- [x] Incorrect links on `/security` page
- [x] "More menu" (seen in tablet view) going underneath blue hero banner on home page

### DIFF
--- a/app/assets/javascripts/moreMenu.js
+++ b/app/assets/javascripts/moreMenu.js
@@ -62,6 +62,8 @@
       .children(":gt(" + fitCount + ")")
       .not("#more-menu");
 
+    collectedSet.find(" > div").addClass("z-50");
+
     // Empty the more menu and add the out of space menu items in
     // a special set.
     var $moreMenuItems = $("<div/>")
--- a/app/assets/javascripts/moreMenu.js
+++ b/app/assets/javascripts/moreMenu.js
@@ -62,14 +62,12 @@
       .children(":gt(" + fitCount + ")")
       .not("#more-menu");
 
-    collectedSet.find(" > div").addClass("z-50");
-
     // Empty the more menu and add the out of space menu items in
     // a special set.
     var $moreMenuItems = $("<div/>")
       .attr("id", "more-menu-items")
       .addClass(
-        "absolute right-0 mr-gutterHalf flex flex-col flex-shrink-0 text-right bg-gray divide-y divide-gray-grey2  shadow"
+        "absolute right-0 mr-gutterHalf flex flex-col flex-shrink-0 text-right bg-gray divide-y divide-gray-grey2  shadow z-50"
       );
     $menuContainer.html($moreMenuItems);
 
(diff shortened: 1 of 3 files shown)


andrewleith:
Better question: where is that menu? :) How do I get there?

jimleroyer:
The images from the `/message-delivery-status` page now points to production. Is that what we want? We didn't want these to target on staging env from review/staging environments targeting staging?

app/assets/javascripts/moreMenu.js
```diff
@@ -62,6 +62,8 @@
       .children(":gt(" + fitCount + ")")
       .not("#more-menu");
 
+    collectedSet.find(" > div").addClass("z-50");
```
jimleroyer, line 65: hmm that could break easily

jimleroyer:
The more menu fix is a bit messed up compared to production:

In review environment:

image

In production environment:

image

andrewleith:
> The images from the `/message-delivery-status` page now points to production. Is that what we want? We didn't want these to target on staging env from review/staging environments targeting staging?

Both notify staging and notify PROD point to GCA PROD, which I think is correct.  

What we should probably do, however, is create a notify-staging site in GCA prod, and point our staging to it so we can make changes there that don't affect our prod content.

What we didn't want is to have any our environments pointing to a staging version of GCA, since it would not be reliable.

andrewleith:
> The more menu fix is a bit messed up compared to production:
> 
> In review environment:
> 
> image
> 
> In production environment:
> 
> image

Looks like i targeted the wrong element in my fix, just fixed it again, please re-check!

amazingphilippe:
Menus looks good on my end:
Capture d’écran, le 2022-05-26 à 14 00 30
Capture d’écran, le 2022-05-26 à 14 00 44

andrewleith:
@jimleroyer good on your end?

jimleroyer:
> Both notify staging and notify PROD point to GCA PROD, which I think is correct.

> What we should probably do, however, is create a notify-staging site in GCA prod, and point our staging to it so we can make changes there that don't affect our prod content.

> What we didn't want is to have any our environments pointing to a staging version of GCA, since it would not be reliable.

Historically, we had staging environments to target staging assets (and still have for most content not served by GCArticles). 

Practically, targeting production assets in staging environment will not be handy for testing nor consistency (we'll have assets from both staging and prod environments mixed together in the staging environment).

I understand that's because we are now serving GCArticles that is in production, hence that production targets production assets. That makes sense... I wonder if there is no way out of this though to keep testability and consistency. That will be harder to test new assets that way.

I can approve this though as this is a current design limitation and outside the scope of this PR.

jimleroyer (approved):
:shipit:

PR #1282, opened 2022-06-03
Setting INFO log level for production env

jimleroyer:
# Summary | Résumé

Looking at logging in production, there are a few DEBUG level statements around the nonce injection that pollutes the logs. Reviewing the debug level statements in the app, these are the only ones present in the admin component. It would be useful to keep the DEBUG level out of the production system and have these enabled locally. We can switch back the logging value for STAGING back to DEBUG once this has been tested in that environment.

Test in staging environment first and look at the AWS CloudWatch logs to compare before/after with expected missing log statements and non-expected ones.

### DIFF
--- a/app/utils.py
+++ b/app/utils.py
@@ -782,7 +782,7 @@ def _geolocate_lookup(ip):
         with urllib.request.urlopen(request) as f:
             response = f.read()
     except urllib.error.HTTPError as e:
-        current_app.logger.debug("Exception found: {}".format(e))
+        current_app.logger.warning("Exception found: {}".format(e))
         return ip
     else:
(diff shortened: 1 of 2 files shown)


app/utils.py
```diff
@@ -782,7 +782,7 @@ def _geolocate_lookup(ip):
         with urllib.request.urlopen(request) as f:
             response = f.read()
     except urllib.error.HTTPError as e:
-        current_app.logger.debug("Exception found: {}".format(e))
+        current_app.logger.warning("Exception found: {}".format(e))
```
jimleroyer, line 785: Switching that debug statement to keep it in the logs, for when geolocate lookup fails. I might switch it to an INFO level if there are too many of these.

PR #1293, opened 2022-07-07
A11y/fix mobile menus

andrewleith:
# Summary | Résumé
- Focuses on first menu item when menu is opened
- Labels menu properly
- Adds proper role for dialog

### DIFF
--- a/app/assets/javascripts/menuOverlay.js
+++ b/app/assets/javascripts/menuOverlay.js
@@ -13,8 +13,8 @@
     if (show) {
       $content.toggleClass("hidden", false);
       // a11y: focus on the first anchor tag when the menu opens
-      if ($content.find('a').length > 0) {
-        $content.find('a')[0].focus();
+      if ($content.find("a").length > 0) {
+        $content.find("a")[0].focus();
       }
     }
     // Hide the menu..
--- a/app/assets/javascripts/menuOverlay.js
+++ b/app/assets/javascripts/menuOverlay.js
@@ -12,6 +12,10 @@
     $menuButton.attr("aria-expanded", show);
     if (show) {
       $content.toggleClass("hidden", false);
+      // a11y: focus on the first anchor tag when the menu opens
+      if ($content.find('a').length > 0) {
+        $content.find('a')[0].focus();
+      }
     }
     // Hide the menu..
     else {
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/menuOverlay.js
```diff
@@ -12,6 +12,10 @@
     $menuButton.attr("aria-expanded", show);
     if (show) {
       $content.toggleClass("hidden", false);
+      // a11y: focus on the first anchor tag when the menu opens
+      if ($content.find("a").length > 0) {
+        $content.find("a")[0].focus();
+      }
```
jimleroyer, line 18: So simple and yet so much better. Should allow to use the escape button right away if I get this right, as right now, one needs to tab to focus on first element.

app/templates/partials/nav/gc_header_nav_mobile.html
```diff
@@ -1,6 +1,8 @@
 {% from "components/nav_menu_item_mobile.html" import nav_menu_item_mobile with context %}
 
 <div
+  role="dialog"
+  aria-label="Menu"
```
amazingphilippe, line 5: We could maybe use the same label as the main navigation `"Main menu"` We could also use `aria-labelledby` and target the "Menu" text on lines 15-17

I think we're missing one prop according to [this guide](https://www.w3.org/WAI/ARIA/apg/patterns/dialogmodal/): 
- `aria-modal` set to true on the dialog container

amazingphilippe (commented):
Just a prop to add. 

And also something I noticed when testing on a real device: Voice over allows you to continue selecting elements that should be inert when the dialog is open. So we can't loop within the dialog. Unsure if there is anything to do here. Even the WAI example has the same bug.

app/templates/partials/nav/gc_header_nav_mobile.html
andrewleith: I prefer `aria-label` as it is a little more straight forward when reading the HTML manually, especially since we already have a lang key for this.  I'll update it to use the that one.
Interesting about `aria-modal` it wasn't mentioned on MDN but I will certainly add that in.

amazingphilippe (approved):

PR #1294, opened 2022-07-07
Update docker.yaml

maxneuvians:
Removes the a11y tracker CI step that is being retired

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -82,13 +82,6 @@ jobs:
         project_name: notification-admin/docker
         project_type: docker
 
-    - name: Run a11y tracker
-      run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/docker.yaml
```diff
@@ -82,13 +82,6 @@ jobs:
         project_name: notification-admin/docker
         project_type: docker
 
-    - name: Run a11y tracker
```
andrewleith, line 85: I didn't even realize that was there.  Will we replace this with anything? Or just continue manual a11y checks?

jzbahrai (approved):

.github/workflows/docker.yaml
maxneuvians: Scan websites is the replacement :)

PR #1297, opened 2022-07-11
Upgrade flask to v2 (admin)

andrewleith:
# Summary | Résumé

## Package bumps
This PR upgrades:
- Flask to 2.0.3
- Flask-WTF to 1.0.1
- itsdangerous to 2.0.1

And adds some packages to keep mypy happy:
- types-python-dateutil  2.8.2
- types-pytz 2021.1
- types-requests 2.25.1

## Flask changes
- the `send_file` method renamed the parameter `filename_or_fp` to `path_or_file`

## Jinja changes
- The `autoescape` and `with_` extensions are now built into jijna and do not need to be referenced as extensions

## Issue with freezegun
As noted by GDS:
> Add itsdangerous to the libraries which freezegun ignores to avoid errors. In tests where we freeze time, the code in the test function will get the frozen time but the  fixtures will be using the current time. This causes itsdangerous to raise an exception - when the session is decoded it appears to be created in the future.

# Testing
- [ ] Ensure pages that use `service_download_agreement()` work
- [ ] Ensure pages that use `public_agreement()` work
- [x] Ensure pages that use GC Articles (i.e. `/home`) works

### DIFF
--- a/babel.cfg
+++ b/babel.cfg
@@ -1,4 +1,1 @@
 [python: app/**.py]
-#[jinja2: app/templates/**.html]
-[jinja2: app/templates/**/**.html]
-extensions=jinja2.ext.autoescape,jinja2.ext.with_
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -44,8 +44,7 @@ awscli-cwlogs>=1.4.6,<1.5
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous==2.0.1  # pyup: <1.0.0
 
-#git+https://github.com/cds-snc/notifier-utils.git@47.0.0#egg=notifications-utils
-git+https://github.com/cds-snc/notifier-utils.git@be1a983ee91f8b3a0ea426d46e91fb5293159e42
+git+https://github.com/cds-snc/notifier-utils.git@48.0.0#egg=notifications-utils
 
 rsa>=4.1 # not directly required, pinned by Snyk to avoid a vulnerability
 
--- a/requirements-app.txt
+++ b/requirements-app.txt
@@ -2,8 +2,8 @@
 # with package version changes made in requirements-app.txt
 
 timeago==1.0.15
-Flask==1.1.2
-Flask-WTF==0.14.3
+Flask==2.0.3
+Flask-WTF==1.0.1
 Flask-Login==0.5.0
 Flask-Caching==1.10.1
 environs==9.5.0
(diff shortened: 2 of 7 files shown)


tests/__init__.py
```diff
@@ -12,6 +13,12 @@
 from app.models.service import Service
 from app.models.user import User
 
+# Add itsdangerous to the libraries which freezegun ignores to avoid errors.
+# In tests where we freeze time, the code in the test function will get the frozen time but the
+# fixtures will be using the current time. This causes itsdangerous to raise an exception - when
+# the session is decoded it appears to be created in the future.
+freezegun.configure(extend_ignore_list=["itsdangerous"])  # type: ignore
```
jimleroyer, line 20: Interesting... thanks for the added comment.

requirements-app.txt
```diff
@@ -42,8 +42,13 @@ redis==3.5.3 # pinned for now
 awscli-cwlogs>=1.4.6,<1.5
 
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
-itsdangerous==0.24  # pyup: <1.0.0
+itsdangerous==2.0.1  # pyup: <1.0.0
 
-git+https://github.com/cds-snc/notifier-utils.git@47.0.0#egg=notifications-utils
+git+https://github.com/cds-snc/notifier-utils.git@48.0.0#egg=notifications-utils
```
jimleroyer, line 47: Tests still ✅ after referring to the new utils release of 48.0.0.

tests/__init__.py
andrewleith: This is from GDS! This was the cause of all of our tests marked with `freeze_time` failing.

andrewleith:
> This pull request **introduces 28 alerts** when merging [d2ca86c](https://github.com/cds-snc/notification-admin/commit/d2ca86c4b7eb13d1a0e26b1b120881b5222c9ebd) into [f654339](https://github.com/cds-snc/notification-admin/commit/f6543391f0fd3478374525859e46683a8c90aece) - [view on LGTM.com](https://lgtm.com/projects/g/cds-snc/notification-admin/rev/pr-bcf6ec8cb5b55868874b3463a55cbe8a73b5775e)
> 
> **new alerts:**
> 
> * 28 for Missing call to `__init__` during object initialization

@jimleroyer any idea what these alerts mean?  They all seem to be related to form classes in `app/main/forms.py`, although none of this code has changed so I am a little puzzled why these alerts would crop up in this PR.

babel.cfg
```diff
@@ -1,4 +1 @@
 [python: app/**.py]
-#[jinja2: app/templates/**.html]
-[jinja2: app/templates/**/**.html]
-extensions=jinja2.ext.autoescape,jinja2.ext.with_
```
sastels, line 4: why delete these lines?

jimleroyer:
@andrewleith Good question... could that be the Flask library upgrade now requires a different super call? We can take a look at it this pm together over remote vscode and gathertown.

andrewleith:
> @andrewleith Good question... could that be the Flask library upgrade now requires a different super call? We can take a look at it this pm together over remote vscode and gathertown.
> 
> How does it look like locally? does it run okay? does form validation runs fine?

Yes I've tested a few forms, and their validation is working correctly.

sastels:
I'm getting an error with the `service_download_agreement()`
https://sentry.io/organizations/canadian-digital-service/issues/3420379979/?project=1522933

However, I don't believe we use this (@jimleroyer ?)  (for example, it requires an env var `MOU_BUCKET_NAME` that we don't define in prod and has default value `""`).

andrewleith:
> I'm getting an error with the `service_download_agreement()` https://sentry.io/organizations/canadian-digital-service/issues/3420379979/?project=1522933
> 
> However, I don't believe we use this (@jimleroyer ?) (for example, it requires an env var `MOU_BUCKET_NAME` that we don't define in prod and has default value `""`).

On that note, I couldn't for the life of me figure out exactly how those methods were called, so this is somewhat reassuring. :)

sastels:
also public agreement isn't working on staging or prod anyway (ex https://staging.notification.cdssandbox.xyz/agreement/crown then try to download the file)
I think we can ignore this code.

sastels (approved):
all looks good to me and things are still working!

jimleroyer:
I'll create a card to slash these two pages / URL endpoints. This could be used to raise it to the product / policy team and maybe nuke 💥 it.

PR #1299, opened 2022-07-12
Update moment.js due to security vulnernability

andrewleith:
# Summary | Résumé

Updates momentjs to 2.29.4 to solve 2 vulnerabilities:
- https://www.cve.org/CVERecord?id=CVE-2022-31129 
- https://www.cve.org/CVERecord?id=CVE-2022-24785

## Testing notes
- Moment.js is used throughout the application, wherever we display a date into an element with the class `.local-datetime-short`.  I would think if it still works in one of these places, it will work in all as the underlying code that gets and places the dates in the HTML has not changed:
   - Service dashboard
   - Revoked api keys list
   - Complaints (admin)
   - And more!

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "hogan": "1.0.2",
     "jquery": "^3.6.0",
     "jquery-migrate": "^3.3.2",
-    "moment": "^2.29.1",
+    "moment": "^2.29.4",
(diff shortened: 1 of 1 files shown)


sastels (approved):
LGTM 🚀

jimleroyer (approved):

PR #1300, opened 2022-07-14
Remove test that calls out to the internet and is making our CI fail.

andrewleith:
# Summary | Résumé
Removes the test `test_blocklist1_can_contact_hibp_api` which is calling out to the internet and now failing.


sastels (approved):
LGTM 👍

PR #1301, opened 2022-07-18
Nested nav fix

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/776

We had a nav nested within a nav. That's illegal. 

Instead, I made it so we have one nav, and all the descendant links are part of a single structured list. 

Test account and language menus both when signed in and signed out. This PR should only affect the desktop version of this nav.


app/templates/partials/nav/account_menu.html
```diff
@@ -1,5 +1,5 @@
 <!-- Large screen menu dropdown -->
-<nav class="hidden md:visible md:flex relative justify-end inline-block text-left w-full text-blue visited:text-blue
+<li class="hidden md:visible md:flex relative justify-end inline-block text-left w-full text-blue visited:text-blue
```
andrewleith, line 2: 👍 Much better as an unordered list!

andrewleith (approved):
LGTM

PR #1302, opened 2022-07-19
A11Y: Fix up account menu labelling

andrewleith:
Summary:

- Move the `aria-label` off of the div (it isn't allowed on a non-interactive element) and down to the menu button instead
- Remove `aria-label` from `img` tag


jimleroyer (approved):
LGTM!

amazingphilippe:
Also looks good tome! Although I did notice that we don't treat both modal menus with the same aria attributes. I'll open up a new PR to demonstrate.

PR #1305, opened 2022-07-27
Remove code for returning a list from the api

jzbahrai:
# Summary | Résumé

Given all the changes we were doing in API, we had to https://github.com/cds-snc/notification-admin/pull/1298/files
introduce that PR as the return type changes from list(list) to list(dicts).
We have now found the API to be stable and are removing the code that handled list(list) from the API and instead only have list(dict)

## Tested
Review app shows no errors.


jimleroyer (approved):
LGTM! We could wait for @sastels though as he crafted that logic.

sastels (approved):
LGTM!

PR #1310, opened 2022-08-05
feat: report csp violations to the CDS report service

mohamed-cds:
Browsers support sending content security violations to the server identifed in the content security policy header returned by the app.

We recently launched a [new service](https://csp-reports.security.cdssandbox.xyz/) to collect these reports. The goal is to make them available to CDS teams so they can view violations users may be experiencing. This app can only be accessed by CDS employees after authenticating via Google SSO.

Information sent by the browser is the following:
- document-uri
- referrer
- violated-directive
- original-policy
- blocked-uri

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -629,7 +629,7 @@ def useful_headers_after_request(response):
     asset_domain = current_app.config["ASSET_DOMAIN"]
     response.headers.add(
         "Content-Security-Policy",
-        (   
+        (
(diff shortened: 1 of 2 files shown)


patheard (approved):
Confirmed that `report-uri` has been added to the CSP response header:
image

andrewleith:
Here is an interesting repo with "weird" CSP reports and some analysis on what they mean: https://github.com/nico3333fr/CSP-useful/blob/master/csp-wtf/explained.md, in case we come across reports we can't quite put a finger on.

PR #1313, opened 2022-08-17
Add WAF secret to header for calls to api

sastels:
# Summary | Résumé

We want to rate limit calls to api, but we don't want to accidentally rate limit ourselves 😅 
Here we add a secret to the request header that will be used by the WAF to let all these requests through.

Check that when admin makes a request to api then the `WAF-secret` header is there.

Can use the [review app](https://xy6um3xjfs3aws3aycbzkmwm6y0xreay.lambda-url.ca-central-1.on.aws/), and do this Athena query on staging:
```
select
    from_unixtime(timestamp / 1000e0) as time,
    httprequest.clientip, header.value as waf_secret,  httprequest.uri, httprequest.headers
from waf_logs
cross join UNNEST(httprequest.headers) AS t(header)
where header.name='waf-secret'
order by timestamp desc
```


jzbahrai (approved):

YedidaZalik:
Already merged, all good

PR #1315, opened 2022-08-18
Remove broken redirects

jzbahrai:
# Summary | Résumé

Remove broken redirects.


jzbahrai:
Removed the broken redirects as stated in the zendesk ticket! Broken Redirects don't work anymore - return page not found

app/main/views/index.py
```diff
@@ -304,16 +304,11 @@ def activity_download():
     )
 
 
-# --- Interal Redirects --- #
```
sastels, line 307: We run a tight ship here 😅

sastels (approved):
🚀

PR #1318, opened 2022-08-31
details-summary focusable elements

amazingphilippe:
# Summary | Résumé

This fixes a small bug found while QAing our summary-detail component. If it contains interactive elements like links, buttons or inputs, some users might be able to still tab to an invisible interactive element while the details are hidden.

Also because you generally don't want focusable stuff in a hidden parent container. https://dequeuniversity.com/rules/axe/4.4/aria-hidden-focus

References for the fix
https://zellwk.com/blog/keyboard-focusable-elements/
https://www.a11yproject.com/posts/how-to-use-the-tabindex-attribute/

This is how it works:

1. Get all the focusable elements
2. Set tabindex to either -1 if state is closed or 0 if state is open
3. When toggling the summary, toggle tabindex on each focusable element

1. Run Axe or ARC toolkit on the upload file step of the send flow
2. Inspect the link in the steps to see that the tabindex prop is being toggled properly. 0 is normal focus order, -1 is not focusable


amazingphilippe:
Not sure why there are 9 fiels here. I'll close and start again

PR #1320, opened 2022-08-31
Big number failure link

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/828

This adds a link styling + a small question mark icon to prompt users to investigate failures from the dashboard. 

Hopefully with a dashboard containing some amount of failures, you should see something like this:

Capture d’écran, le 2022-08-31 à 16 01 09

It could be red or green depending on the percentage.


app/templates/components/big-number.html
```diff
@@ -39,12 +39,13 @@
...
     {% if show_failures %}
-      <div class="big-number-status{% if danger_zone %} bg-red{% endif %}">
+      <div class="big-number-status{% if danger_zone %} bg-red{% endif %}{% if not failures %} p-6{% endif %}">
         {% if failures %}
           {% if failure_link %}
-            <a class="big-number-status-link" href="{{ failure_link }}">
+            <a class="p-6 big-number-status-link flex justify-between items-end border-0" href="{{ failure_link }}">
               {{ "{:,}".format(failures) }}
               {{ _('failed') }} – {{ failure_percentage }}%
+              <i aria-hidden="true" class="p-1 fa-solid fa-circle-question"></i>
```
andrewleith, line 48: This works well and now it is apparent that cell is clickable 🎉

andrewleith (approved):
Looks great overall, much better usability than before. One thing though is that the help icon made me think I should hover over it for some sort of explanation.  Not a show stopper, but maybe we want to use another icon?
image

amazingphilippe:
We could use an arrow, like we already use when we want to lead users somewhere. Like in the card above. Although I try to use this icon as a last resort because it is almost meaningless. 
Capture d’écran, le 2022-09-02 à 14 10 14

Here are a few options we could consider. My case for the ? was that you could link to investigate. I found this magnifting galss icon that could be interesting too. We're limited with the free font awesome regular icons [If you want to explore more](https://fontawesome.com/search?p=2&o=r&m=free&s=solid)!

Capture d’écran, le 2022-09-02 à 14 09 58

amazingphilippe:
Maybe the warning triangle icon is a bit more redundant to the colour red... Which would be a good thing!

jimleroyer:
I like the warning triangle icon. ⚠️

sastels:
Maybe for a lot of failures we could use
image

PR #1325, opened 2022-09-02
add waffles support for GCA routes

sastels:
# Summary | Résumé

add the GCA routes to an env var `EXTRA_ROUTES` so that waffles can pull them out to test.

run this branch together with waffles from https://github.com/cds-snc/notification-utils/pull/130


app/config.py
```diff
@@ -19,6 +21,10 @@
 
 
 class Config(object):
+
+    # for waffles: pull out the routes into a flat list of the form ['/home', '/accueil', '/why-gc-notify', ...]
+    EXTRA_ROUTES = [item for sublist in map(lambda x: x.values(), GC_ARTICLES_ROUTES.values()) for item in sublist]
```
jimleroyer, line 26: Based on previous Slack conversation with @andrewleith , would GC_ARTICLES_ROUTES cover all of existing ones for GCNotify?

sastels: Possibly not - for this approach to work we should find any other routes and add them to the `GC_ARTICLES_ROUTES` object.

jzbahrai (approved):

PR #1327, opened 2022-09-06
Feat/daily sms limit

andrewleith:
# Summary | Résumé
This PR adds the sms daily limit to admin settings so it can be viewed by all and edited by admins.  It also updates the wording around the existing daily limit to specify that it is for emails.

## NOTE: Feature flag
You need a feature flag in order to see these screens.  It is on in the PR review environment, but locally you'll have to add this to your `.env`:
```
FF_SPIKE_SMS_DAILY_LIMIT=False
```

### Service settings changes
image

### Platform admin changes
image

---

### SMS daily limit
- [ ] Ensure daily sms limit can be seen from a service's settings page
- [ ] Ensure daily sms limit can be seen from a service's settings page within the platform admin section
- [ ] Ensure user receives an email when the daily sms limit is modified

### Email daily limit
- [ ] Ensure the daily email limit is correctly labelled to refer to emails and not all notifications in the service's settings page
- [ ] Ensure the daily email limit is correctly labelled to refer to emails and not all notifications in the service's settings page within the platform admin section
 - [ ] Ensure user receives an email when the daily email limit is modified


app/templates/views/service-settings.html
```diff
@@ -98,13 +98,22 @@ <h2 class="heading-small">{{ _('Your service is in trial mode') }}</h2>
         {% endcall %}
 
         {% call row() %}
-          {% set txt = _('Daily message limit') %}
+          {% set txt = _('Daily email message limit') %}
```
sastels, line 101: right now `message_limit` is still for notifications as a whole - we should either wrap this in the feature flag or just leave as "Daily message limit" everywhere for now.

andrewleith: Ah good point, got a little ahead of myself.

andrewleith: All set now.

smcmurtry:
The form pages seem to be working to set the values in settings: https://www.youtube.com/watch?v=vn9-09ucz5M

However I don't get an email to notify me that a change was made. I also do not get an email when I change the message limit of my service on https://staging.notification.cdssandbox.xyz/. So not sure if that is a problem or not.

It looks like the French translation is not showing up:
Screen Shot 2022-09-14 at 3 54 36 PM

smcmurtry:
Also I have run this locally and can confirm that the feature is invisible if I do not set the feature flag in `.env`.

app/main/views/service_settings.py
```diff
@@ -1056,9 +1061,25 @@ def set_message_limit(service_id):
...
+@main.route("/services/<service_id>/service-settings/set-sms-message-limit", methods=["GET", "POST"])
+@user_is_platform_admin
+def set_sms_message_limit(service_id):
+
+    form = SMSMessageLimit(message_limit=current_service.sms_daily_limit, other_data="HI")
+
+    if form.validate_on_submit():
+        service_api_client.update_sms_message_limit(service_id, form.message_limit.data)
+        if current_service.live:
+            flash(_("An email has been sent to service users"), "default_with_tick")
```
smcmurtry, line 1077: When I update the SMS limit on a live service, I receive the following email:
Screen Shot 2022-09-14 at 4 36 49 PM

The 10,002 number is actually the email limit, not the SMS limit. I think we need to make a 2nd email template here and insert the correct number.

andrewleith: Yes I think you're right.  Initially I had planned to update this template to include both numbers, but with the feature flag it will be more tricky.  

Looks like this sort of thing was done in https://github.com/cds-snc/notification-api/blob/main/migrations/versions/0319_warn_daily_limits.py so I will follow this example.

andrewleith:
> The form pages seem to be working to set the values in settings: https://www.youtube.com/watch?v=vn9-09ucz5M
> 
> ~However I don't get an email to notify me that a change was made. I also do not get an email when I change the message limit of my service on https://staging.notification.cdssandbox.xyz/. So not sure if that is a problem or not.~ Okay this is only for live services.
> 
> It looks like the French translation is not showing up: Screen Shot 2022-09-14 at 3 54 36 PM

Just added the translation.  @amazingphilippe, does that read OK in French?

Old text:
> Limite d'envoi quotidienne

New text:
> Limite d'envoi courriel quotidienne
> Limite d'envoi SMS quotidienne

app/main/views/service_settings.py
andrewleith: I'll add a card for this work since it will be done outside of this PR.

smcmurtry: Okay sounds good!

amazingphilippe:
@andrewleith We use consistently "text messages" and "messages texte" So I would suggest the following translations:

Daily email limit, Limite quotidienne d'envoi par courriel
Daily text message limit, Limite quotidienne d'envoi par message texte

I know its a mouthful, but it sounds correct. We can work on shortening the French a bit! Also let me know if it doesn't fit in the table 🙃

smcmurtry (approved):
Looks great!

app/main/views/service_settings.py
andrewleith: New card here: https://github.com/cds-snc/notification-planning/issues/868

PR #1329, opened 2022-09-13
Add opp.ca to list of allowed domain names

andrewleith:
# Summary | Résumé

Allow opp.ca folks to create accounts.


smcmurtry (approved):
👍

PR #1331, opened 2022-09-13
links that open in a new tab

amazingphilippe:
# Summary | Résumé

I had an idea to re-use how we localise content within Javascript files to add some interesting styles behaviour.

This PR proposes to use CSS to automatically add a screen-reader-only "Opens in a new tab" to all target=_blank links. 

Here's how it works:

Just like we set polylang strings in main_template.html, we can set CSS strings as CSS variables. Later, we can use them in our CSS files for things like the use case above. 

This has the benefit of simplifying some translation keys, avoiding the need to add in some gnarly html strings. (See how we had to add `<span class='sr-only'> (Opens in a new tab)</span>` within the translation strings. 

I tested it with voiceover on safari and it works super nicely. We should do some research to see if this technique works with other screen readers. This is where I encountered this [technique](https://ashleemboyer.com/blog/should-i-use-a-button-or-a-link#links-to-external-resources)

Sidenote: there are a few more files for a "minor" idea, but its essentially the same change in many files. 

# Discussion

This is super low priority, but I just wanted to put the idea out there!


smcmurtry:
Where should I look in the review app to see this change? I looked at the "API documentation" link in the footer (where I think this change should show up) but I didn't see the `<span>` you added in the DOM:

Screen Shot 2022-09-13 at 4 52 58 PM

Screen Shot 2022-09-13 at 4 52 28 PM

amazingphilippe:
This is exactly where it would appear! If you're logged out, it'll also show in the main navigation.

I _removed_ the `<span>` in fr.csv, on lines 1554 and 1556

But the purpose of this PR is more to suggest that we treat external link consistently with screen readers. Each link opening in a new tab should announce this with some visually hidden text.

Therefore, instead of duplicating the `<span>` in our translation strings, we can have something a bit more automatic.

smcmurtry:
Okay I understand - the text doesn't show up when I inspect the DOM because it is inserted with CSS. Looks like it works just the same as before:
Current Prod: https://youtu.be/1cKuUoo4vvg
This PR: https://youtu.be/NSkcVRiBWXg

I think this new approach is better because it is less code and we are less likely to miss a link since the same css can target many links. Also it looks like this approach is [recommended by W3C](https://www.w3.org/TR/WCAG20-TECHS/G201.html).

I say go for it!

app/translations/csv/fr.csv
```diff
@@ -1551,9 +1551,9 @@
...
+"A template allows you to send the same message to many people. To tailor messages for each recipient, visit the <a href='{}' target='_blank'>Personalisation guide<img aria-hidden='true' role='presentation' class='inline w-6 h-6 ml-2 self-center' src='{}'/></a>.","Un gabarit vous permet d’envoyer le même message à de nombreuses personnes. Pour adapter le message à chaque destinataire, utilisez le <a href='{}' target='_blank'>guide de personnalisation<img aria-hidden='true' role='presentation' class='inline w-6 h-6 ml-2 self-center' src='{}'/></a>."
```
andrewleith, line 1554: 👍 This is so much cleaner not to be in the language files!

app/templates/main_template.html
```diff
@@ -92,6 +92,11 @@
     }
 
   </script>
+  <style title="CSS strings">
```
andrewleith, line 95: It's too bad we have to put these here in an HTML file instead of inside the `.css` files.  I guess that's because you need it to translate on the fly?

app/assets/stylesheets/tailwind/components/typography.css
```diff
@@ -38,6 +38,11 @@
     @apply text-blue;
   }
 
+  a[target="_blank"]::after {
```
andrewleith, line 41: This is excellent, now no one can miss marking up an external link accessibly, and they are all consistent!  🏅

andrewleith (approved):

app/templates/main_template.html
amazingphilippe: That's it! Since we don't (and shouldn't) localize our css files!

app/templates/components/links.html
```diff
@@ -9,21 +9,11 @@
     {% if id_key %}id="{{ id_key }}"{% endif %}
     class="my-0 p-2 line-under text-base border-b-2 border-transparent link:text-blue visited:text-blue focus:bg-yellow focus:border-blue focus:no-underline focus:outline-none hover:no-underline decoration-clone {% if class %} {{ class }} {% endif %}"
     href="{{ href }}"
-    {% if is_external_link %}
-        target="_blank"
-        aria-label="{{ label }} ({{ _('Opens in a new tab') }})"
-    {% else %}
-        aria-label="{{ label }}"
-    {% endif %}
+    {% if is_external_link %}target="_blank"{% endif %}
```
jimleroyer, line 12: 🥇

jimleroyer (approved):
Looks nice!

PR #1332, opened 2022-09-14
remove fake tabs, use navs

amazingphilippe:
# Summary | Résumé

This updates our pill component to feel just like a nav to screen reader users. The pill component already worked like a nav, so the markup doesn't really changes. I did simplify the component a bit, especially for the label and did some CSS tweaks that should be unoticeable. 

(Fixes) https://github.com/cds-snc/notification-planning/issues/498

These are the endpoints you can use to view the different pill component instances

- /monthly
- /notifications/email?status=delivered
- /templates/email

1. Check that the nav is labelled with aria-label, in both languages
2. Check that the links work, and land you where you expect

A11y QA:
- [x] Scan with ARC toolkit or Axe devtools. For issues related to the pill
- [x] Test the pill with voiceover on Safari
- [ ] Test the landmark menu with voiceover on Safari


app/assets/stylesheets/tailwind/components/pill.css
```diff
@@ -81,7 +83,7 @@
   .pill-separate-item:hover,
   .pill-separate-item:focus,
   .pill-separate-item:link:focus {
-    @apply text-blue-lightblue25;
+    @apply bg-yellow text-blue;
```
smcmurtry, line 86: This is causing somewhat weird behaviour where only the background of the selected link of the pill turns blue when you hover, so now the hover styles are inconsistent. Before, all links in the pill turned blue background on hover. The links already have an underline that appears on hover, so maybe we can just remove the background colour hover style?

Here's a demo - 1st our current app then your review app: https://youtu.be/XK3egR4w1Vw

app/templates/components/pill.html
```diff
@@ -4,38 +4,27 @@
   items=[],
   current_value=None,
   big_number_args={'smaller': True},
-  show_count=True
+  show_count=True,
+  label=""
 ) %}
-  <div role='tablist' class='pill flex'>
+  <nav class='pill flex' aria-label="{{ label }}">
```
smcmurtry, line 10: Nice!

smcmurtry (changes requested):
This is great @amazingphilippe! I like the way the `<nav>` is working, it seems much clearer using voiceover:

Current prod: https://youtu.be/-rie7OE7ne4
Your review app: https://youtu.be/nygf3HhpcLY

The only thing I noticed is an issue with the focus styles - can you take a look at that before merging?

app/templates/components/pill.html
```diff
@@ -4,38 +4,27 @@
   items=[],
   current_value=None,
   big_number_args={'smaller': True},
-  show_count=True
+  show_count=True,
+  label=""
```
jimleroyer, line 8: Interesting, looks simpler now in the implementation. 👍

app/assets/stylesheets/tailwind/components/pill.css
amazingphilippe: Oh yeah, that is weird. I wanted to simplify this CSS a bit. Good catch!

amazingphilippe: Should be fixed! I reviewed CSS specificity, and found out about LVHA ordering. added a comment for future reference.

smcmurtry (approved):
Looks good to me!

PR #1334, opened 2022-09-15
Adding SMS message parts to the "Review before sending" pages

smcmurtry:
# Summary | Résumé

Adding SMS message parts to the "Review before sending" pages.

We know that the SMS message parts are referred to as `billable_units`: https://github.com/cds-snc/notification-api/blob/main/app/delivery/send_to_providers.py#L98

And that is available through the `fragment_count` property on `template` objects. 

Interestingly, 
```remaining_messages = current_service.message_limit - sum(stat["requested"] for stat in statistics.values())```
https://github.com/cds-snc/notification-admin/blob/main/app/main/views/send.py#L633

So this should be calculated based on what has already been requested, not what has already been sent, since a job may be in progress with messages requested but not yet sent.

## Still to do

- [x] Add in the actual daily SMS limit
- [x] Wrap in feature flag
- [x] Set the feature flag for the AWS review app

Create a long SMS template and navigate to the following pages:
- "Send this message to yourself" > "Review before sending" 
- "Add recipients" > Upload csv > "Review before sending" 

The new content should be visible and the numbers should be correct. The new content should not be visible if you are not sending SMS or if the feature flag is not set.

### CODE REVIEW

**Review by jzbahrai — APPROVED** [review]
lgtm.

File: app/main/views/send.py [hunk 1]
```diff
@@ -631,6 +633,8 @@ def _check_messages(service_id, template_id, upload_id, preview_row, letters_as_
 
     statistics = service_api_client.get_service_statistics(service_id, today_only=True)
     remaining_messages = current_service.message_limit - sum(stat["requested"] for stat in statistics.values())
+    # todo: update with the sms specific limit when available
+    remaining_sms_messages = current_service.message_limit - sum(stat["requested"] for stat in statistics.values())
```
sastels, line 637: ah, it should be `sms_daily_limit` but I didn't add it to admin yet 😞 I can make a little admin PR that adds that...

sastels (changes requested):
just change that `current_service.message_limit` to `current_service.sms_daily_limit` and we should be GTG!

sastels (approved):

sastels:
Looks sharp! only question: should we hide this behind the feature flag?

PR #1335, opened 2022-09-20
A11y/no truncation

amazingphilippe:
# Summary | Résumé

This removes all (I think) instances of text truncation and allows text to flow to a new line. 

There will be some odd things happening with tables at very small sizes. I think this is okay for the scope of this fix. We can work on the tables in another PR. (To allow horizontal scrolling or row wrapping for example)

Here is a list of pages I think we can review. Check out "Départment de la température ambiante" in staging

- [x] The dashboard Templates used section
- [x] A job report, or any other job status page
- [x] The settings
- [x] There is a page where I think callbacks are listed. I was unable to generate such a page. It could be impossible to test.


smcmurtry (approved):
Looks good Phil! I was able to check all 4 pages you suggested against staging. All the ellipses have been removed and I don't even see any styling/overflow problems at small screen widths. 

Here are the many diff screenshots I took:
[Archive.zip](https://github.com/cds-snc/notification-admin/files/9619958/Archive.zip)

PR #1336, opened 2022-09-21
Reply-to adresses as a list

amazingphilippe:
# Summary | Résumé

(Fixes) https://github.com/cds-snc/notification-planning/issues/588

This pr reworks the reply-to settings page to use a list structure. I also cleaned up the styles so that things would look a bit more ordered. 

Checkout the reply-to settings page in a service that has

- [ ] 0 reply-to addresses
- [ ] 1 reply-to address
- [ ] More than 1 reply-to address


app/templates/views/service-settings/email_reply_to.html
```diff
@@ -22,24 +22,29 @@
...
       </div>
-    {% endif %}
-    {% for item in current_service.email_reply_to_addresses %}
-      <div class="user-list-item">
-        <h3>
-          <span class="heading-small">{{ item.email_address }}</span>&ensp;<span class="hint">
-            {%- if item.is_default -%}
-              ({{ _('default') }})
+    {% else %}
+    <ul class="divide-y divide-gray-300">
```
smcmurtry, line 26: Much more appropriate than the previous markup.

smcmurtry (approved):
Looks way better Phil! The skipped heading problem no longer exists since you removed the `h3`s. Also the styling is way better now.

Prod:
missing_h2


Phil's PR:
fixed


Phil's PR narrow screen:
narrow_screen

PR #1339, opened 2022-09-26
A11y/personalize title and instructions

amazingphilippe:
# Summary | Résumé

Provides a descriptive title on the personalization flow

Doing so, we can also make the input label the page's `h1`. Since we're only asking one thing. 

Todo:
- [ ] Fix tests
- [ ] Review content
- [ ] A11y scans


amazingphilippe:
I found a bug when using a trial service:

Capture d’écran, le 2022-09-29 à 09 25 58

amazingphilippe:
Fixed the issue above by placing the Trial limitation text where it makes more sense: On the Add recipients screen.

sastels (approved):
LGTM. Ran through with voiceover and axe devtools, didn't notice any issues.

PR #480, opened 2020-04-28
Fr dashboard update

timarney:
**Fixes:**
https://github.com/cds-snc/notification-api/issues/751

**Staging preview link:**
https://notification-fr-dashboa-btrmiz.herokuapp.com

Screen Shot 2020-04-28 at 1 29 02 PM

<hr>

Screen Shot 2020-04-28 at 1 29 14 PM

PR #1340, opened 2022-09-27
Adding the "Daily Usage" section to the dashboard

smcmurtry:
# Summary | Résumé

Working on https://github.com/cds-snc/notification-planning/issues/865.

Adding the "Daily Usage" section to the dashboard. This new section is hidden behind the feature flag. It is calculated using "sms notifications" instead of "sms parts" until those become available. 

We will need a follow up PR or PRs to add:
- a red/"0 SMS remaining" state
- French content
- correct sms parts number remaining

 
Screen Shot 2022-09-28 at 4 41 18 PM


app/templates/components/message-count-label.html
```diff
@@ -19,6 +19,18 @@
     {%- else -%}
       {{ _('text-messages') }}
     {%- endif -%}
+  {%- elif template_type == 'sms-parts' -%}
```
andrewleith, line 22: I noticed on the dashboard we now say "fragments" but we said "parts" on the review screens, we'll need to be consistent.

app/templates/components/message-count-label.html
```diff
@@ -19,6 +19,18 @@
...
+    {%- if count == 0 -%}
+      {%- if session["userlang"] == "fr" -%}
+        {{ _('text message fragment') }}
```
andrewleith, line 25: 🔬 Interesting that "0 items" in french is not pluralized! :til:

app/templates/views/dashboard/dashboard.html
```diff
@@ -24,11 +24,18 @@ <h2 class="heading-medium mt-8">{{ _("Scheduled sends") }}</h2>
       {{ ajax_block(partials, updates_url, 'upcoming', interval=5) }}
     {% endif %}
 
+    {% if config["FF_SPIKE_SMS_DAILY_LIMIT"] %}
+      <h2 class="heading-medium mt-8">
```
andrewleith, line 28: Perhaps a test to ensure this section is included when the FF is on?

amazingphilippe (changes requested):
A few changes I would like to see in this PR:
1. If we are now counting notifications instead of fragments, we could adjust the content for this PR. Switch it later in the next dashboard PR.
2. I want the French to be at least OK in this PR, even if we have to follow it up with changes. I'll have a look at th message-count-label file and suggest changes!

amazingphilippe:
I just added the French content to this PR. But I have a problem where the plural form doesn't apply to sms only... I have no idea why, because it seems to work for emails: 

Capture d’écran, le 2022-09-29 à 14 20 45

amazingphilippe:
> I have no idea why, because it seems to work for emails:

There was an exception put in place only for sms. After looking at what we have in production, I noticed the FR was wrong there too. So it went under our radar since #480 !! 😲 

0 message texte envoyé
0 message texte restant
2 messages texte envoyés
2 messages texte restants

tests/conftest.py
```diff
@@ -32,6 +32,16 @@
...
+if os.getenv("_PYTEST_RAISE", "0") != "0":
+
+    @pytest.hookimpl(tryfirst=True)
+    def pytest_exception_interact(call):
+        raise call.excinfo.value
+
+    @pytest.hookimpl(tryfirst=True)
+    def pytest_internalerror(excinfo):
+        raise excinfo.value
+
```
andrewleith, line 44: What's this?

tests/app/main/views/test_dashboard.py
```diff
@@ -441,6 +442,38 @@ def test_correct_columns_display_on_dashboard(
     assert len(page.select(column_name)) == expected_column_count
 
 
+@pytest.mark.parametrize(
+    "feature_flag",
+    [True, False],
+)
+def test_daily_usage_section_shown(
```
andrewleith, line 449: Looks good! 👍

andrewleith (approved):
Left a question for you and Phil in the slack thread about wording. 

Just a question below about what you put in `conftest`, but other than that good to go.

tests/conftest.py
smcmurtry: Whoops! I added that when I was trying to get the test debugger to work. I didn't mean to include it in this PR. Removing, thanks for catching that!

amazingphilippe (approved):
Current content seems to pluralize correct in both FR and EN!

PR #1341, opened 2022-09-28
set FF correctly

sastels:
# Summary | Résumé

Fix how the FF is set

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -127,7 +127,7 @@ class Config(object):
     ZENDESK_API_KEY = os.environ.get("ZENDESK_API_KEY")
 
     # FEATURE FLAGS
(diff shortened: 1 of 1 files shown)


app/config.py
```diff
@@ -127,7 +127,7 @@ class Config(object):
     ZENDESK_API_KEY = os.environ.get("ZENDESK_API_KEY")
 
     # FEATURE FLAGS
-    FF_SPIKE_SMS_DAILY_LIMIT = os.environ.get("FF_SPIKE_SMS_DAILY_LIMIT", False)
+    FF_SPIKE_SMS_DAILY_LIMIT = env.bool("FF_SPIKE_SMS_DAILY_LIMIT", False)
```
andrewleith, line 130: My bad, must have copied that from an old FF.  Thanks for fixing this up. 🙇

sastels: You probably copied it from api where I set it wrong too 😓

PR #1343, opened 2022-09-29
New sms parts error message

sastels:
# Summary | Résumé

Rough in an error in admin if you try to send too many sms on a day. Note that I have omitted the "Schedule for tomorrow" feature for now - This will require more complicated changes (and at the same time should be added to the existing general "over daily limit" flow)

image

image

image

image

image

image

- set your sms daily limit low
- send big messages

Notes:
- review apps currently crash on csv uploads - testing locally shows the same error message when the csv is too big.
- It appears that api sends the message that takes you over the limit and returns an error the next message you try. Not the fault of this PR, but something to investigate in api.
- right now we're just using the number of sms messages previously sent, not the number of message fragments, to decide whether to show the error for the csv upload. We should do another PR that gets the correct number of fragments sent into admin.


tests/conftest.py
```diff
@@ -902,11 +902,19 @@ def _get(service_id, template_id, version=None):
 
 @pytest.fixture(scope="function")
 def mock_get_service_email_template_without_placeholders(mocker):
-    return mock_get_service_email_template(
```
sastels, line 905: You're not allowed to call mocks as they do here (`mock_get_service_email_template()`). We never noticed that `mock_get_service_email_template_without_placeholders` had this problem because we were not using this mock before this PR.

app/translations/csv/fr.csv
```diff
@@ -1650,4 +1651,10 @@
...
 "This exceeds your daily limit. ","Ceci dépasse votre limite pour aujourd’hui. "
-"This might exceed your daily limit. ","Ceci pourrait dépasser votre limite pour aujourd’hui. "
\ No newline at end of file
+"This might exceed your daily limit. ","Ceci pourrait dépasser votre limite pour aujourd’hui. "
+"You’ve reached the daily limit","Vous avez atteint votre limite quotidienne"
+"text message fragments per day","bouts de message texte par jour"
+"Too many text message fragments","Trop de bouts de message texte"
+"These messages exceed your daily limit","Ces messages dépassent votre limite quotidienne"
+"This message exceeds your daily limit","Ce message dépasse votre limite quotidienne"
+"If a message is long, it travels in fragments. Each fragment counts toward your daily limit.","Un message long est transmis en fragments. Chaque fragment compte dans votre limite quotidienne de fragments de message texte."
```
amazingphilippe, line 1660: "Un long message est transmis en fragments. Chaque fragment compte dans votre limite quotidienne de fragments de message texte."

Just the "long" and "message" are inverted.

amazingphilippe (commented):
We should also hide this block behind `not send_exceeds_daily_limit`

https://github.com/cds-snc/notification-admin/blob/800feda32156814d9d34f75888d409a74f4dd1cb/app/templates/views/check/column-errors.html#L173-L182

andrewleith:
When doing a CSV send, the h1 doesn't match what the error is:
image

sastels:
> We should also hide this block behind `not send_exceeds_daily_limit`
> 
> https://github.com/cds-snc/notification-admin/blob/800feda32156814d9d34f75888d409a74f4dd1cb/app/templates/views/check/column-errors.html#L173-L182

Currently if you're over the daily limit (or for other problems with the csv) we show the "Choose file" button - do we want to remove that button for all errors?
image
(Yes, we should have a different header here too)

andrewleith:
> > When doing a CSV send, the h1 doesn't match what the error is
> 
> yeah, I noticed that and fixed it in the new "too many sms" error but not for the existing "over daily limit" error.
> 
> I can change that as well, ie use the new "You've reached the daily limit", etc headers for daily limit errors as well, that's a minor tweak that would improve the page 👍

Looks great 👍 
image

I am still able to send a one-off SMS though when I reach the limit.

andrewleith:
Ok just to wrap this up, the problem I was having was that my local celery workers were dead, so my messages weren't actually sending, but the requested SMS number was still increasing in the UI.  This particular issue is a pretty unlikely edge case in production, but we will have to think on if we want to use requested SMSs to enforce limits or sent.

To note, right now, the redis counter for daily sends only increments when a message is SENT, not while it is queued for sending.

andrewleith (approved):
LGTM

PR #1345, opened 2022-10-03
fix typo in sender address preview

amazingphilippe:
# Summary | Résumé

Just a minor fix that I think we should fix right away. When we are creating a new service, we invert "canada" and "notification" in the sender email preview.

image

Create a new service, try the sender email previewer.


app/templates/partials/add-service/step-create-service.html
```diff
@@ -28,7 +28,7 @@ <h2 class="heading-medium">{{ _("Email address") }}</h2>
 
       <span id="preview" style="display: none">
         {{ _("Your service’s email address will be:") }}
-        <b><span id="fixed-email-address"></span>@canada.notification.ca</span></b>
+        <b><span id="fixed-email-address"></span>@notification.canada.ca</b>
```
andrewleith, line 31: Good catch!

andrewleith (approved):
:shipit:

PR #1346, opened 2022-10-03
fix/parsing api sms error message

sastels:
# Summary | Résumé

In https://github.com/cds-snc/notification-api/pull/1612 we changed api (when the FF is on) to returns an error message "Exceeded sms send limits" to admin if sending the sms would put a service over its limits, as required by #1343.  This causes an admin server error since admin currently doesn't know what to do with this error.

This PR makes admin treat the error as an "over daily limit" error for now. This will be remedied by #1343.

Note that this has no affect when the FF is off in api, ie production.

Set your sms daily limit low and try to send sms messages. no more 500 errors!


sastels:
Just fixed in another PR

PR #1348, opened 2022-10-04
added the appropriate value for GC_ARTICLES_API to test config

smcmurtry:
# Summary | Résumé

I have a staging url set in my .env file for `GC_ARTICLES_API`. This causes the tests to fail. Adding the correct value to the test config makes the tests pass regardless of what you have set in your local .env file.

## Test instructions

Check out branch and try setting a different value for `GC_ARTICLES_API` in your .env file. The tests should still pass.


sastels (approved):
🚀

PR #1349, opened 2022-10-04
Fetch sms fragment count from redis

sastels:
# Summary | Résumé

In the csv upload code, fetch the number of sms segments from redis rather than just using the number of messages.


app/main/views/send.py
```diff
@@ -633,8 +639,8 @@ def _check_messages(service_id, template_id, upload_id, preview_row, letters_as_
 
     statistics = service_api_client.get_service_statistics(service_id, today_only=True)
     remaining_messages = current_service.message_limit - sum(stat["requested"] for stat in statistics.values())
-    # todo: currently stat["requested"] uses sms messages, update this when sms parts becomes available
-    remaining_sms_messages = current_service.sms_daily_limit - sum(stat["requested"] for stat in statistics.values())
+    sms_sent_today = daily_sms_count(service_id)
+    remaining_sms_messages = current_service.sms_daily_limit - sms_sent_today
```
smcmurtry, line 643: What are the implications of using "sms sent today" rather than "sms requested today"? The number we eventually show for "sms remaining" will not be true for very long if a bulk send is in progress. 

I think the "sms requested today" number would be better for usability. Is it not possible to get that number to admin currently, because the billable units aren't calculated until the message is sent?

smcmurtry (changes requested):
I believe you still need to update the logic with this new variable in a couple more places:

https://github.com/cds-snc/notification-admin/blob/main/app/main/views/send.py#L773-L775
and
https://github.com/cds-snc/notification-admin/blob/main/app/main/views/send.py#L1033-L1035

That was an easy mistake to make since I duplicated the logic in a few different places. Sorry!

app/main/views/send.py
sastels: should name it better - this is really `sms_fragments_sent_today` - will update variable and function

agree we should really use "sms fragments requested today" but didn't want to get into the api changes / new endpoint. Was thinking of pushing that to V2.

smcmurtry: Okay that makes sense to me - we can make those changes for the V2 release since it will not impact V1.

smcmurtry (approved):
Looks good!

PR #1351, opened 2022-10-04
New feature flag `FF_SMS_PARTS_UI`

smcmurtry:
# Summary | Résumé

Part of https://github.com/cds-snc/notification-planning/issues/901

I created the new feature flag `FF_SMS_PARTS_UI` and put UI components behind it that are not part of the V1 release of the SMS Limits feature.

I also added a test to check that the relevant section is shown/hidden on the "review before sending" page depending on the feature flag.

## To test

Check out branch, set the value of the new feature flag to `True` and `False`, checking that the relevant sections are visible.


app/config.py
```diff
@@ -182,6 +183,7 @@ class Test(Development):
     TESTING = True
     WTF_CSRF_ENABLED = False
     FF_SPIKE_SMS_DAILY_LIMIT = False
+    FF_SMS_PARTS_UI = False
```
andrewleith, line 186: When is this class used?

tests/app/main/views/test_send.py
```diff
@@ -3601,3 +3601,23 @@ class Object(object):
     multiple_choise_options = [x.text.strip() for x in options]
 
     assert multiple_choise_options == expected_filenames
+
+
+@pytest.mark.parametrize(
+    "feature_flag, expected",
+    [(True, 1), (False, 0)],
+)
```
andrewleith, line 3609: This is a cool way to test the FF on/off 👍

andrewleith (approved):

app/config.py
smcmurtry: This sets the value of the feature flag when the tests are running. The other classes let you set defaults for staging, prod, etc.

PR #1354, opened 2022-10-12
[Do Not Merge] Experiment: Testing out the design system

smcmurtry:
# Summary | Résumé

I followed the [Install with CDN](https://github.com/cds-snc/gcds-components/tree/main/packages/web#install-with-cdn) instructions and created this branch. Just an experiment to see what would happen.

Currently the styling is not getting applied correctly.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -632,12 +632,13 @@ def useful_headers_after_request(response):
         (
             "report-uri https://csp-report-to.security.cdssandbox.xyz/report;"
             "default-src 'self' {asset_domain} 'unsafe-inline';"
-            f"script-src 'self' {asset_domain} *.google-analytics.com *.googletagmanager.com https://tagmanager.google.com https://js-agent.newrelic.com 'nonce-{nonce}' 'unsafe-eval' data:;"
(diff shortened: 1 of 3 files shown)


andrewleith:
👀

app/__init__.py
```diff
@@ -632,12 +632,13 @@ def useful_headers_after_request(response):
         (
             "report-uri https://csp-report-to.security.cdssandbox.xyz/report;"
             "default-src 'self' {asset_domain} 'unsafe-inline';"
-            f"script-src 'self' {asset_domain} *.google-analytics.com *.googletagmanager.com https://tagmanager.google.com https://js-agent.newrelic.com 'nonce-{nonce}' 'unsafe-eval' data:;"
+            f"script-src 'self' {asset_domain} *.google-analytics.com *.googletagmanager.com https://tagmanager.google.com https://js-agent.newrelic.com https://unpkg.com/ 'nonce-{nonce}' 'unsafe-eval' data:;"
```
amazingphilippe, line 635: Adding CSP headers allows the js and styles to load! Should we be more precise than just "unpkg" I tried to add the whole `https://unpkg.com/gcds-components` but I don't think I know what I'm doing.

andrewleith: Nice! I *think* you do this at the domain-level, so you can't be more specific.

smcmurtry: Nice fix @amazingphilippe! I can see the button now, although I think we will need to do some server side rendering to avoid the flash of unstyled content when the page loads:
https://www.youtube.com/watch?v=pLmQ2EzxEWY

PR #1356, opened 2022-10-14
Consistent labels for create and choose template

amazingphilippe:
# Summary | Résumé

(Fixes) https://github.com/cds-snc/notification-planning/issues/566

1. On a service with templates
2. View the dashboard
3. Find the "Create template" and "Choose template" buttons. There should be no "a" as in "Choose a template"
4. When browsing templates. The button should also be "Create template"
5. When moving a template. The final button should be "Create template"


smcmurtry:
From a dev point of view this PR looks fine to me. I will defer to @YedidaZalik on if we want to make this change from a content standpoint. I noticed there are other buttons that use "a" or "an" like:
- Add a new service
- Create an API key

YedidaZalik:
@amazingphilippe @smcmurtry it looks fine to me as well
It would be good to be consistent and change to 
- Add new service
- Create API key
- there may be other examples of this as well

But these changes are not urgent from a content standpoint, unless they impact accessibility in other ways

smcmurtry (approved):

PR #1358, opened 2022-10-17
Fix styling for SMS limit error

smcmurtry:
# Summary | Résumé

Working on [notification-planning/issues/928
](https://github.com/cds-snc/notification-planning/issues/928)

Before:
Screen Shot 2022-10-12 at 3 41 02 PM

After:
Screen Shot 2022-10-17 at 10 09 17 AM

## Testing instructions:

Set your SMS daily limit to 1 and then send some SMS. This will trigger the error message.


amazingphilippe:
> Should we also get the heading inside the red box, though?
We need a h1 out of the box. I think the content inside the box needs to be updated in an upcomming PR. The message in the box will have its own heading, should look less out of place!

amazingphilippe:
@smcmurtry Did you consider using an more global approach for `p` margin in banners? 

In see two options
1. We could add a line in layout.css to target paragraphs only in banners using something like `.banner p:last-child`
2. We could add a line in typography.css to target `p:last-child`. I'm pretty confident we don't want margin on a paragraph if their the last thing in a container. That might be a bold move though...

We have a few more places where we have the same bug. Like:

Capture d’écran 2022-10-17 à 13 28 24

smcmurtry:
My last push is a more global approach. Here's what it looks like for those 2 case:
Screen Shot 2022-10-17 at 4 53 24 PM
Screen Shot 2022-10-17 at 4 54 12 PM

app/assets/stylesheets/tailwind/components/layout.css
```diff
@@ -31,44 +44,53 @@
   }
 
   .column-main .heading-large,
-  .column-main > .heading-medium {
+  .column-main>.heading-medium {
```
amazingphilippe, line 47: Is this auto formatting? I don't know if the space matters.

amazingphilippe (approved):

PR #1361, opened 2022-10-18
Integration of qualtrics survey in staging only

jimleroyer:
# Summary | Résumé

This integrates the survey Javascript widget into the admin UI. The widget will only show in non-production environments and for authenticated users.

**TODO**: 
* Include UI tests within Cypress (won't be done within these changes and require MFA to work with Cypress).

* Load into the review environment and check if the widget works.
* Unit tests around presence or absence of survey widget element depending on the admin base URL (local vs staging vs prod).


smcmurtry (changes requested):
In the review app, the feedback button text is showing up in french for me, when I have Notify displaying in English:
Screen Shot 2022-10-19 at 3 27 17 PM

jimleroyer:
Good find @smcmurtry . I investigated a bit and left comments on our internal Slack thread about this topic. 👍

app/__init__.py
```diff
@@ -633,12 +633,15 @@ def useful_headers_after_request(response):
...
+            f"script-src-elem 'self' *.siteintercept.qualtrics.com https://siteintercept.qualtrics.com 'nonce-{nonce}' 'unsafe-eval' data:;"
+            "connect-src 'self' *.google-analytics.com *.googletagmanager.com *.siteintercept.qualtrics.com https://siteintercept.qualtrics.com;"
             "object-src 'self';"
             "style-src 'self' *.googleapis.com https://tagmanager.google.com https://fonts.googleapis.com 'unsafe-inline';"
             "font-src 'self' {asset_domain} *.googleapis.com *.gstatic.com data:;"
             "img-src 'self' {asset_domain} *.canada.ca *.cdssandbox.xyz *.google-analytics.com *.googletagmanager.com *.notifications.service.gov.uk *.gstatic.com data:;"  # noqa: E501
-            "frame-src 'self' www.googletagmanager.com www.youtube.com;".format(asset_domain=asset_domain)
+            "frame-src 'self' www.googletagmanager.com www.youtube.com https://cdssnc.qualtrics.com/;".format(
+                asset_domain=asset_domain
+            )
```
andrewleith, line 644: These look about right to me, although I had to look each up 

- `script-src-elem` -> allows us to load scripts from qualtrics
- `connect-src` -> allows us it to fetch, do XMLHttpRequest, web sockets, etc. from qualtrics
- `frame-src` -> allows us to load an iframe hosted on qualtrics

---

`script-src-elem`:
> The HTTP [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) (CSP) script-src-elem directive specifies valid sources for JavaScript [<script>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script) elements, but not inline script event handlers like onclick.

`connect-src`:
> The HTTP [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) (CSP) connect-src directive restricts the URLs which can be loaded using script interfaces.

`frame-src`:
> The HTTP [Content-Security-Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy) (CSP) frame-src directive specifies valid sources for nested browsing contexts loading using elements such as [<frame>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/frame) and [<iframe>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe).

tests/app/main/views/test_headers.py
```diff
@@ -134,6 +134,6 @@ def test_headers_non_ascii_characters_are_replaced(
         "style-src 'self' *.googleapis.com https://tagmanager.google.com https://fonts.googleapis.com 'unsafe-inline';"
         "font-src 'self' static.example.com *.googleapis.com *.gstatic.com data:;"
         "img-src "
-        "'self' static.example.com *.canada.ca *.cdssandbox.xyz *.google-analytics.com *.googletagmanager.com *.notifications.service.gov.uk *.gstatic.com data:;"  # noqa: E501
+        "'self' static.example.com *.canada.ca *.cdssandbox.xyz *.google-analytics.com *.googletagmanager.com *.notifications.service.gov.uk *.gstatic.com https://siteintercept.qualtrics.com data:;"  # noqa: E501
```
andrewleith, line 137: Looks good 👍

andrewleith (approved):
Survey still doesn't load in French when the app is in French (although the button correctly translates), but approving this anyway since you said that wasn't critical for this stage.

jimleroyer:
> Survey still doesn't load in French when the app is in French (although the button correctly translates), but approving this anyway since you said that wasn't critical for this stage.

Only the survey button on the side at this point should follow the language. Within the survey, there is a language selector. I don't know yet if we can get rid of that selector to follow the website selected lang or at least, select the right one by default. 

For these changes, the goal is to integrate Qualtrics to Notify in a safe env by deploying to staging so the Growth team can persue Qualtrics configuration. 👍

smcmurtry (approved):
I'm still seeing the button only in french, but maybe that can be fixed by some config on the qualtrics side

PR #1364, opened 2022-10-19
Template history and API logs as lists

amazingphilippe:
# Summary | Résumé

---

---


sastels:
Would the template versions be better as an unordered list or is this important for a11y?
image

amazingphilippe:
> Would the template versions be better as an unordered list or is this important for a11y?

This made me think a bit, but [checking the w3c docs](https://www.w3.org/WAI/WCAG21/Understanding/meaningful-sequence.html), it seems that:

> Tables and ordered lists are meaningful sequences, but unordered lists are not. 

So I think we need to use an `ol` (if we don't want to use a table!)
Regarding the reverse number CSS. This has no impact with voice over. Each point is read as "1 of 3", "2 of 3"

so the reverse order is just for visuals, so that the version number aligns with the list number. Otherwise, reading the list as...

1. Version 3
2. Version 2
3. Version 1

...felt off!


I'll push another commit to hide those markers. but it won't work on safari, possibly not on chrome too. I'll do some tests.

amazingphilippe:
> I'll push another commit to hide those markers. but it won't work on safari, possibly not on chrome too. **I'll do some tests.**

Visually hidden ordered list markers:

Chrome: reads as bullets when markers have `content: ""` 🚫 
Firefox: ✅ Surprisingly works well with voiceover
Safari: 🚫 No effect -> no impact

sastels (approved):
Looks good and sounds good! 🚀

PR #1367, opened 2022-10-24
Changed error message for sms limit reached

amazingphilippe:
# Summary | Résumé

This updates the error message when the sms limit is reached to be a bit more relevant in different situations. 

1. On a trial service
2. Try to send more messages than you're allowed daily
3. Get this error message,

Maybe as a control
1. Try to send less messages
2. Don't get this error message.

### DIFF
--- a/app/templates/components/links.html
+++ b/app/templates/components/links.html
@@ -1,4 +1,4 @@
-{% macro content_link_light(
+{% macro content_link(
     label="TODO: LABEL THIS LINK",
     href="/",
     class=None,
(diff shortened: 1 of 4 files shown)


smcmurtry (changes requested):
Looks good except for one very small thing - the "contact us" link is a slightly larger font than the surrounding text:
 
Screen Shot 2022-10-24 at 3 37 11 PM

amazingphilippe:
Can't believe I missed that! Good catch!

smcmurtry (approved):
Error message looks good now. Your fix did change the font size of a link on the home page from 2.0em to 1.9em, but that is probably fine 😄 

Screen Shot 2022-10-24 at 4 21 02 PM

amazingphilippe:
> Your fix did change the font size of a link on the home page from 2.0em to 1.9em

Yeah! I figured that an inline link component shouldn't have opinions about font size! That way it can adapt to more contexts.

PR #1369, opened 2022-10-25
Introduce a new js var to store qualtrics lang

jimleroyer:
# Summary | Résumé

Set the JS variable that stores Qualtrics survey language to always be upper case. Derive from existing `APP_LANG` and store into `QUALTRICS_LANG`.

* Configure Qualtrics intercept to have language javascript variable set with `QUALTRICS_LANG`.
* Test the review environment to see if the survey is in the appropriate language as set by the web app.


andrewleith (commented):
I'm still seeing the survey open in english when the app is in French.

jimleroyer:
@andrewleith Qualtrics haven't been configured yet on this new js variable.

jzbahrai (approved):

PR #1373, opened 2022-10-27
Fix catching SMS over limit errors from API

andrewleith:
This PR updates the error message caught from API to show a user an error when they go over their limit.  This will now align with the changes [made here](https://github.com/cds-snc/notification-api/pull/1653).


smcmurtry (approved):
Looks good! I was able to trigger the error message during a one off send as before.

PR #1374, opened 2022-10-28
Add some types

smcmurtry:
# Summary | Résumé

Very small PR to add some type annotations. Had to add the `# type: ignore` comments because mypy expects this to be the `LocalProxy` type. But this change allows the correct `Service` and `Organisation` types to come through to all the other files that import from here.


app/__init__.py
```diff
@@ -93,10 +93,10 @@ def _get_current_service():
     return _lookup_req_object("service")
 
 
-current_service = LocalProxy(_get_current_service)
+current_service: Service = LocalProxy(_get_current_service)  # type: ignore
```
andrewleith, line 96: Nice!  I was wrangling this a little while back and it would have been nice to have this typing. 🚀

andrewleith (approved):
LGTM, much better! :)

PR #1380, opened 2022-11-09
Adding routes for SLA and SLO

amazingphilippe:
# Summary | Résumé

Adding routes for SLA and SLO


andrewleith:
Hmm interesting that the test endpoints check fails: it's actually testing against staging, we'll need to think through how this check should work.  cc: @jimleroyer

andrewleith (approved):
Looks good!

PR #1382, opened 2022-11-09
Fixing ol lists for GCA + list-custom counter fix

amazingphilippe:
# Summary | Résumé

Adds support for ordered lists on GCA pages

1. On the review app
2. Visit /service-level-agreement
3. Visit /accord-niveaux-de-service
4. Ordered lists should display. 
5. They should match the numbers as seen in the wysiwyg

Also fixes this bug for custom lists where the numbers would show as 2. 4. 6.
Capture d’écran, le 2022-11-09 à 14 03 33


andrewleith (approved):
Looks good. Lists display properly on desktop and mobile.

PR #1396, opened 2022-11-15
Fix "preview page server error" bug

smcmurtry:
# Summary | Résumé

I ran this with the debugger and discovered that the `template` object is an empty dict if we have not yet visited the preview page by clicking on the link. This is because the key/value that stores this info (see the `get_preview_data` fn) has not yet been set in Redis. If this is the case, we can use the `get_template_with_user_permission_or_403` method, which makes a request to the API.

See the attached issue.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,30 @@
+
+# Summary | Résumé
+
+
+> 1-3 sentence description of the changed you're proposing, including a link to
+> a GitHub Issue # or Trello card if applicable.
+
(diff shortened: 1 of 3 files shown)


sastels (approved):
Tested locally, can go directly to template page, and "Save" and "Edit" buttons work appropriately.

smcmurtry:
@jimleroyer we have a test `test_preview_page_contains_preview`, I'll look into why that didn't catch this bug.

smcmurtry:
That test mocks the redis function that was returning `None`: https://github.com/cds-snc/notification-admin/blob/main/tests/app/main/views/test_templates.py#L1668

I think I can update the test to make sure it still works if that function returns `{}`.

jimleroyer (approved):
:shipit:

PR #1398, opened 2022-11-16
Fix Reply-to bug

smcmurtry:
# Summary | Résumé

This PR adds support for deleting the default reply-to email address, returning the service to the original state where none are set. This also addresses a weird bug that can be seen on the attached card. The changes needed on the API side have been introduced here: https://github.com/cds-snc/notification-api/pull/1678

~After the [API PR](https://github.com/cds-snc/notification-api/pull/1678) is merged~ (This is now merged!) 
Use the review app to try out the following:

- Add only one reply-to address to a service. This will become the default for that service. Make sure it is possible to delete this and return to the original state.
- Add 3 reply-to addresses to a service. Try deleting one of the non-default ones and make sure this works the same as before. 
- There should be 2 remaining reply-to addresses. Now try deleting the default address. You should see a message that the remaining address will become the new default. Clicking to delete should complete the action successfully.


app/utils.py
```diff
@@ -816,3 +817,13 @@ def _constructLoginData(request):
...
     }
+
+
+def get_new_default_reply_to_address(current_service: Service, default_reply_to_email_address):
+    non_default_reply_to_addresses = [
+        reply_to
+        for reply_to in current_service.email_reply_to_addresses
+        if reply_to["email_address"] != default_reply_to_email_address["email_address"]
+    ]
+    new_default_reply_to_address = non_default_reply_to_addresses[0]
```
Pensai, line 827: Just a suggestion / nit-pick. 
Even though we validate API side that two reply_to email addresses that are the same cannot be used. If for some reason, malicious or otherwise, the user is able to use two reply_to email addresses that are the same `non_default_reply_to_address` will have a length of 0 and throw an index out of bounds exception, resulting in a not so graceful 500 error browser side. 
Maybe shuffle the logic around in the list comprehension or add a check before accessing the list?

smcmurtry: I've added a check in there: `https://github.com/cds-snc/notification-admin/pull/1398/files#diff-ac7b8d56e509c921efaadb5a776e1c9037531c4d2af78787f06f67a1f3533ae4R826` 

And I've added some type annotation to that function now as well.

Let me know if that addresses your comment.

amazingphilippe:
Tested a few times and works as expected! though I found a new (probably already existing) bug. 😅

amazingphilippe (approved):

smcmurtry:
Here is the bug @amazingphilippe found - it is an existing bug not related to this PR: https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/cds-snc/notification-planning/1007

app/main/views/service_settings.py
```diff
@@ -687,6 +688,7 @@ def service_edit_email_reply_to(service_id, reply_to_email_id):
         try:
             notification_id = service_api_client.verify_reply_to_email_address(service_id, form.email_address.data)["data"]["id"]
         except HTTPError as e:
+            # TODO: missing translation
```
Pensai, line 691: Will this translation be a part of this PR eventually or just holding off until it is available?

tests/app/test_utils.py
```diff
@@ -691,3 +699,14 @@ def test_set_lang_external_route(client_request):
         }
     )
     assert len(page.findAll(text="/accounts-or-dashboard")) == 1
+
```
Pensai, line 702: A bit nit picky, but maybe we can add one more quick test to pass in two emails that are the same and test that we get `None` back from the `get_new_default_reply_to_address` method?

Pensai (commented):
Looks good to me, just a couple small  questions / suggestions more than anything else.

I like the typing on `EmailReplyTo` it's giving me fuzzy Java vibes on that List type in `get_new_default_reply_to_address`

app/main/views/service_settings.py
smcmurtry: The missing translation is unrelated to this PR so I won't add it here. But thought I would add a note to make it easier to find later.

Pensai: Ah, makes sense!

tests/app/test_utils.py
smcmurtry: I added a test for this plus another one to check `None` is returned when only one address exists.

app/utils.py
whabanks: Looking great!

whabanks (approved):

PR #1402, opened 2022-11-17
Fix screen overflow on team page

amazingphilippe:
# Summary | Résumé

Noticed some overflow on the teams page that caused unwanted horizontal scrolling.  Screenshot below:

Capture d’écran, le 2022-11-17 à 11 41 07

I removed some older float CSS and replaced with some flex magic. 

1. Go to the team member screen
2. Test different viewport widths.


Pensai:
Noticed a couple of issues. The `change details` link seems to have disappeared, and the scroll bar is still present.
image

The element is present in the DOM, just seemingly not visible.
image

I ran main and checked the same service / team member list:
image

smcmurtry:
This fixes the issue for me - tested on Chrome for a number of different screen widths. 

@Pensai I didn't notice any issue with the "Change details" button disappearing. Here is what it looks like on my end: https://youtu.be/cge4YwtJ0es. Did you test with the review app linked above, or locally? If you are testing locally, maybe your CSS needs to be rebuilt with `npm run tailwind`.

Pensai:
@smcmurtry I was testing locally and it looks like `npm run tailwind` was what I was missing. @amazingphilippe apologies for the false alarm, it's looking great both locally and with the above environment link.

Pensai (approved):
Looks great, much smoother and pretty looking!

PR #1405, opened 2022-11-23
Run prettier to fix formatting

andrewleith:
# Summary | Résumé

This just runs prettier on our js/css resources to fix formatting issues.  There are no functional changes.


sastels (approved):
I kinda like it the way it was, but I won't argue with the formatting gods.

andrewleith:
💯 Agreed - not sure how/why the rules changed all of the sudden.  Prettier does not seem to have updated 🤷

PR #1408, opened 2022-11-30
Pin waffles and node alpine to specific versions

smcmurtry:
# Summary | Résumé

Not having waffles and node alpine pinned to specific versions was causing failure in our CI pipeline over in this PR: https://github.com/cds-snc/notification-admin/pull/1406

I thought it would be best to open a separate PR to address this.

CI should pass. Test suite should pass in CI. This should be a safe change since this returns us to versions of `notification-utils` and node alpine we have been using until recently.


patheard (approved):

sastels (approved):
LGTM!

PR #1425, opened 2022-12-05
Revert "Make bulk the default priority for newly created templates (#748)"

jzbahrai:
Reverts cds-snc/notification-admin#1399

PR #1426, opened 2022-12-05
Reinstate set template default priority to bulk - Fix 403 error for template preview and saves

whabanks:
# Summary | Résumé

Re-introduce the template priority default to Bulk change with a fix the 403 error when attempting to edit and save a template where a variable had been added to the body of the template.

- Updated mocked template used in tests to reflect new default template process type.
- Updated a couple tests

1. Open a template
2. Add a `((somename))` variable to the body
3. Save the template
4. Note no longer receiving a 403 on save

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -44,6 +44,7 @@ from app.main.forms import (
     TemplateFolderForm,
 )
 from app.main.views.send import get_example_csv_rows, get_sender_details
+from app.models.enum.template_process_types import TemplateProcessTypes
 from app.models.service import Service
 from app.models.template_list import TemplateList, TemplateLists
 from app.template_previews import TemplatePreview, get_page_count_for_letter
@@ -707,11 +708,11 @@ def add_service_template(service_id, template_type, template_folder_id=None):
 
     template = get_preview_data(service_id)
     if template.get("process_type") is None:
-        template["process_type"] = "normal"
+        template["process_type"] = TemplateProcessTypes.BULK.value
     form = form_objects[template_type](**template)
 
     if form.validate_on_submit():
-        if form.process_type.data != "normal":
+        if form.process_type.data != TemplateProcessTypes.BULK.value:
             abort_403_if_not_admin_user()
         subject = form.subject.data if hasattr(form, "subject") else None
         if request.form.get("button_pressed") == "preview":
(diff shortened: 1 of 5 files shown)


app/main/views/templates.py
```diff
@@ -707,11 +708,11 @@ def add_service_template(service_id, template_type, template_folder_id=None):
 
     template = get_preview_data(service_id)
     if template.get("process_type") is None:
-        template["process_type"] = "normal"
+        template["process_type"] = TemplateProcessTypes.BULK.value
     form = form_objects[template_type](**template)
 
     if form.validate_on_submit():
-        if form.process_type.data != "normal":
+        if form.process_type.data != TemplateProcessTypes.BULK.value:
```
jzbahrai, line 715: Is this correct? I was under the impression that an old template that was made under normal/ priority can't be viewed due to this line of code?

whabanks: Good catch, something must have gone a bit haywire during merge/rebase. I'll fix this.

jzbahrai: this is what we reverted here: https://github.com/cds-snc/notification-admin/pull/1425/files#diff-ab6836e1de1ae35fdbf46e68b40e71ca658c734751950019b2c08aba9c4084a1R714

and from how I read it - we are saying if the the template is not bulk, abort 403 ?

smcmurtry: +1 @jzbahrai  I think we don't want to change this line. This will cause the 403 bug that we saw before: when a user edits a template that was created in the past, that has a normal priority.

smcmurtry: Also, it will be normal for a platform admin to change a template from "bulk" priority to another priority. And then the user could no longer edit the template. 

I'm confused why the main branch is showing: `if form.process_type.data != "normal":` though. 
I think it should be: `if form.process_type.data != template["process_type"]:`
https://github.com/cds-snc/notification-admin/pull/1399/files#diff-ab6836e1de1ae35fdbf46e68b40e71ca658c734751950019b2c08aba9c4084a1L808

smcmurtry: Looks like the revert PR reverted that line to the wrong logic: https://github.com/cds-snc/notification-admin/pull/1425/files#diff-ab6836e1de1ae35fdbf46e68b40e71ca658c734751950019b2c08aba9c4084a1R714

smcmurtry (changes requested):

app/main/views/templates.py
whabanks: Almost: If the template being created is not bulk, and the user is not an admin, abort 403. 
The problematic code giving us the 403 was in `edit_service_template` on L810, not here in `add_service_template`. 

This check should remain the same, we set the priority to Bulk by default and when they actually save the template we check to make sure that the `process_type` does not differ from the default before saving. We 403 if it does and they are not an admin.

The extra check on L810: `request.form.get("button_pressed") != "save"` was unnecessary and what caused the issue.

smcmurtry (approved):
Looks good!

PR #1432, opened 2022-12-28
Update client session timeout from 20 hours to 8 hours

whabanks:
# Summary | Résumé
Lower the session token expiration timeout from 20 hours to 8 hours (may change?). 

Log in, wait until the session expires. You should be sent to a "You were signed out" type of page, asking you to log back in when the session timeout has elapsed.


smcmurtry (approved):
I checked this out locally and changed this variable to 60 and I was signed out after 1 minute of inactivity. So I think your change will work as expected ✅

PR #1435, opened 2023-01-10
add CodeQL workflow

sastels:
# Summary | Résumé

Basically #1385 but not from a forked repo, so our tests run


sastels:
TBH, I think it's best to just turn off the test requirement and merge the original PR.

PR #1465, opened 2023-01-11
Restore go-live setting service limit to 10k. 

whabanks:
# Summary | Résumé
Reverting work done in [this PR](https://github.com/cds-snc/notification-admin/pull/935) to ensure that when a service goes live, their message limit is properly set to 10k. When the service is no longer live, but restricted, the limit is restored to the default of 50.

Create a service and change the daily message limit. Go live with the service. Note that the daily message limit is set to 10k.
Turn off live, and note that the message limit is set to 50.

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -371,6 +371,7 @@ def service_switch_live(service_id):
 
         if live:
             message_limit = current_app.config["DEFAULT_LIVE_SERVICE_LIMIT"]
+            sms_daily_limit = current_app.config["DEFAULT_LIVE_SMS_DAILY_LIMIT"]
 
         flash(_("An email has been sent to service users"), "default_with_tick")
 
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -368,18 +368,11 @@ def service_switch_live(service_id):
         live = form.enabled.data
         message_limit = current_app.config["DEFAULT_SERVICE_LIMIT"]
         sms_daily_limit = current_app.config["DEFAULT_SMS_DAILY_LIMIT"]
+
         if live:
-            if current_service.message_limit != current_app.config["DEFAULT_SERVICE_LIMIT"]:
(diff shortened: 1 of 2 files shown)


app/main/views/service_settings.py
```diff
@@ -368,18 +368,11 @@ def service_switch_live(service_id):
...
 
-            if current_service.sms_daily_limit != current_app.config["DEFAULT_SMS_DAILY_LIMIT"]:
-                sms_daily_limit = current_service.sms_daily_limit
-            else:
-                sms_daily_limit = current_app.config["DEFAULT_LIVE_SMS_DAILY_LIMIT"]
+        if live:
+            message_limit = current_app.config["DEFAULT_LIVE_SERVICE_LIMIT"]
 
-            flash(_("An email has been sent to service users"), "default_with_tick")
+        flash(_("An email has been sent to service users"), "default_with_tick")
```
andrewleith, line 376: Makes much more sense to me! 👍

tests/app/main/views/test_service_settings.py
```diff
@@ -622,10 +622,10 @@ def test_show_restricted_service(
 @pytest.mark.parametrize(
     "current_limit, expected_limit, current_sms_limit, expected_sms_limit",
     [
-        (42, 42, 33, 33),
+        (42, 10_000, 33, 50),
```
andrewleith, line 625: Isn't the default live sms limit supposed to be 1000? Maybe I'm [looking in the wrong place](https://github.com/cds-snc/notification-admin/blob/main/app/config.py#L66).

andrewleith (commented):
A question below on the default number of SMS for a live service.

tests/app/main/views/test_service_settings.py
whabanks: That looks more correct to me, I'll make the adjustment.

andrewleith (approved):
Thats better! Not sure why deploying the pr environment is failing, ill take a look at that now.

PR #1468, opened 2023-01-24
Feat/update why notify route fr

andrewleith:
# Summary | Résumé

- Update the French `/pourquoi-gc-notification` to `/pourquoi-notification-gc` as per our new naming convention.

### DIFF
--- a/tests/app/articles/test_pages.py
+++ b/tests/app/articles/test_pages.py
@@ -102,8 +102,9 @@ def test_bad_slug_doesnt_save_empty_cache_entry(app_, mocker):
             assert not mock_redis_method.set.called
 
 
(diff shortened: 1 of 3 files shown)


tests/app/articles/test_pages.py
```diff
@@ -102,8 +102,9 @@ def test_bad_slug_doesnt_save_empty_cache_entry(app_, mocker):
             assert not mock_redis_method.set.called
 
 
-@pytest.mark.skip(reason="TODO: a11y test")
```
andrewleith, line 105: I'm uncommenting this test because it isn't an a11y test.  It is a valid test to ensure our redirects are working.

whabanks (approved):

PR #1471, opened 2023-01-25
fix: build PR review Docker image with `docker build`

patheard:
## Summary
The Docker images being built by the GitHub Docker build and push action are no longer working with AWS Lambda.  We get the following error when attempting to create a function with them:

```
The image manifest or layer media type for the source image $IMAGE_TAG is not supported.
```

Using a raw `docker build` command works as expected and is consistent with how all the other Notify Docker images are built.

### DIFF
--- a/.github/workflows/test-admin-deploy.yaml
+++ b/.github/workflows/test-admin-deploy.yaml
@@ -40,43 +40,20 @@ jobs:
         id: login-ecr
         uses: aws-actions/amazon-ecr-login@261a7de32bda11ba01f4d75c4ed6caf3739e54be # v1.5.3
 
-      - name: Set up Docker Buildx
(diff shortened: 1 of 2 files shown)


.github/workflows/test-admin-deploy.yaml
```diff
@@ -40,41 +40,20 @@ jobs:
...
+      - name: Build Docker image
+        run: |
+          docker build \
+            --build-arg GIT_SHA=${{ github.sha }} \
+            -t ${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.PR_NUMBER }} \
+            -f ci/Dockerfile.lambda .
+          
+      - name: Push Docker image to ECR
+        run: |
+          docker push ${{ env.REGISTRY }}/${{ env.IMAGE }}:${{ env.PR_NUMBER }}
```
andrewleith, line 56: 🏅

andrewleith (approved):
Nicely done @patheard!

PR #1476, opened 2023-02-15
Limit the api key name to 255 characters to prevent a 500 error

smcmurtry:
# Summary | Résumé

This change will prevent users from entering in api key names longer than the allowed length into the form field.

See attached issue card.


whabanks (approved):
LGTM!

PR #1479, opened 2023-02-16
Restore Renovate functionality

whabanks:
# Summary | Résumé
Alpine docker images use `musl` rather than `glibc` which breaks wheel support.

[PEP 656](https://peps.python.org/pep-0656/) proposes support for musl, the caveat is it's up to package maintainers to support musl based wheels.
For packages that don't support it, C libs are compiled at dependency resolution time increasing docker build times, resulting in Renovate exceeding an internal timeout causing the tool to fail.

This PR swaps out Alpine images for the official Python image `python:3.10-slim`, restoring support for wheels and cutting dependency resolution times from 300+ seconds to ~56 seconds.  In addition the `poetry.lock` file was updated.

- [ ] Deploy to staging
- [ ] Run smoke tests to ensure functionality is intact


jimleroyer:
We fixed the renovate issue in some other manner back then and these changes might be unnecessary.

PR #1480, opened 2023-02-17
chore: remove the Renovate Poetry config

patheard:
# Summary
This config is not required as the default Renovate Python config already manages Poetry manifests by default.

The reason for the lack of dependency updates is because of a related bug with the Poetry dependency resolution causing the Renovate Docker container to fail with a `SIGTERM`.

# Related
- renovatebot/renovate#10769


whabanks (approved):
LGTM!

PR #1482, opened 2023-02-27
test: Renovate ignore Werkzeug dependency

patheard:
# Summary
Update the Renote config to no longer manage the Python Werkzeug dependency.  This is being done because Poetry currently has a bug that causes dependency conflicts to result in an infinite dependency resolution bug.

# Related
- cds-snc/notification-planning#658
- cds-snc/platform-core-services#287


jimleroyer (approved):
We can try this out. 👍

jimleroyer:
@patheard Did you have a link to an issue that describes that infinite loop bug?

PR #1484, opened 2023-02-28
update the new relic python agent

andrewleith:
# Summary | Résumé

This PR updates the new relic agent used in admin.


andrewleith:
Hey @jimleroyer let me know if you have any concerns with this PR.  I was unable to test the agent update locally so I put it into staging.  It looks like it works well, data is still coming into NR, including errors etc. and the warning is now gone from the main screen in NR staging.

PR #1486, opened 2023-03-08
feat: allow PR review env to use Redis

patheard:
# Summary
Update the PR review environment to be deployed in the VPC so that it has access to the Redis cluster.

# Related
- cds-snc/notification-terraform#682
- cds-snc/notification-terraform#681
- cds-snc/notification-planning#650
- cds-snc/notification-planning#1017
- cds-snc/platform-core-services#285


smcmurtry (approved):
Looks like redis is working in the review app here!

PR #1489, opened 2023-03-16
Update Dashboard - bounce rate

whabanks:
# Summary | Résumé

This PR implements the new dashboard  main page features for the bounce rate epic: 
 - [x] Inaccurate addresses status block
 - [x] Enhance failure widget
 - [x] Review email addresses link (adds the link but the link doesnt work yet!)

Notes:
- Data for these new features is taken directly from the available stats data the UI already polls
- When the redis back-end is complete for bounce rate, we may want to consider reading the data from there instead
- These features are hidden behind a feature flag named `FF_BOUNCE_RATE_V1`
- These features are based on a set of configuration values which modify the behaviour:
   - `BR_DISPLAY_VOLUME_MINIMUM` - this flag controls the minimum number of email notifications sent by a service before the warning/critical displays kick in.  If the service has sent less than this amount, the UI will always be green
   - `BR_WARNING_PERCENTAGE` - this value controls when threshold at which the "warning" state is shown
   - `BR_CRITICAL_PERCENTAGE` - this value controls when threshold at which the "critical" state is shown

## Other info
- [Figma link](https://www.figma.com/file/w1rwZuxuh8D6SGskv0KPaK/Bounce-rate?node-id=451-10014&t=GUrFnlfpFDtM1rYv-4)
- [Req doc](https://docs.google.com/document/d/1rc85aF7PGVN5WO6XMnVtKsiQSqiEvnaKabQ5ZWYVdjo/edit#)


app/templates/views/dashboard/_totals.html
```diff
@@ -1,35 +1,60 @@
...
         message_count_label(statistics['email']['requested'], 'email', suffix=sent_single, suffix_plural=sent_plural),
         statistics['email']['failed'],
-        statistics['email']['failed_percentage'],
+        (statistics['email']['failed_percentage'])|int,
         statistics['email']['show_warning'],
         failure_link=url_for(".view_notifications", service_id=service_id, message_type='email', status='failed'),
         link=url_for(".view_notifications", service_id=service_id, message_type='email', status='sending,delivered,failed'),
         smaller=smaller_font_size
       ) }}
     </div>
```
andrewleith, line 30: The UI looks good here, but I think the entire red square should be clickable, like the blue squares are.  The blue squares are marked up as `a` tags with some inner content which makes the entire area clickable.  The red ones contain the `a` tag inside a div instead, making it a little harder to interact with and causing some minor layout issues:
image

andrewleith (commented):
Still looking through the code but here is a comment re: the UI

app/templates/views/dashboard/_totals.html
whabanks: This should be much better now, the entire block is clickable and I've added the appropriate text colour changes when hovering as well as the yellow outline when clicking. There's still 1 small issue. The "Review Email Addresses" text does not seem to be respecting the "no-underline" utility class from Tailwind on hover like the block above it does.

app/extensions.py
```diff
@@ -8,4 +9,5 @@
 statsd_client = StatsdClient()
 zendesk_client = ZendeskClient()
 redis_client = RedisClient()
+bounce_rate_client = RedisBounceRate(redis_client)
```
jzbahrai, line 12: do we need this? given we are just changing the query on the backend?

jzbahrai (commented):
I went through the review app and the code. It looks good to me from both a design perspective and code perspective. I will wait for Phil/ Adrianne to give a design review.

app/templates/views/dashboard/_totals.html
```diff
@@ -13,23 +23,38 @@
...
-        statistics['sms']['failed'],
-        statistics['sms']['failed_percentage'],
-        statistics['sms']['show_warning'],
-        failure_link=url_for(".view_notifications", service_id=service_id, message_type='sms', status='failed'),
-        link=url_for(".view_notifications", service_id=service_id, message_type='sms', status='sending,delivered,failed'),
-        smaller=smaller_font_size
-      ) }}
+
+    <div id="problem-email-addresses" class="{{column_width}}">
+      {% if config["FF_BOUNCE_RATE_V1"] %}
```
jzbahrai, line 33: :+1

andrewleith:
Going to close this PR in favor of https://github.com/cds-snc/notification-admin/pull/1531, which has all of these commits already.

PR #1483, opened 2023-02-27
wip: CRM integeration with trial service creation

patheard:
# Summary
Initial work prototyping how the Notify and Salesforce integration could work.  The goal of this PR is to give stakeholders a chance to see how the following two flows in the Notify admin will be updated:

1. `New trial service creation`: users will be asked to link their trial service to a department/organization.
2. `Go live request`: the trial service's department/organization will be pre-populated.

This PR's review environment will also handle the creation of Salesforce Contact and Opportunities data in the QA environment to give the CRM team a chance to test.

# Related
- cds-snc/platform-core-services#277
- cds-snc/platform-core-services#278
- cds-snc/platform-core-services#282
- cds-snc/platform-core-services#285

PR #1491, opened 2023-03-20
CRM-related changes to Trial Account creation flow

smcmurtry:
# Summary | Résumé

This PR contains all of the UI changes for the CRM integration from https://github.com/cds-snc/notification-admin/pull/1483, but none of the Salesforce API calls, since the plan is to add those to `notification-api`.

This PR adds a "Step 3" to the Trial Service creation flow. This step is only visible if the feature flag `FF_SALESFORCE_CONTACT` is set to `True`. The organisation names are populated from [here](https://github.com/cds-snc/gc-organisations/tree/main/data). The new autocomplete "drop down" is actually a text input box.

The "organisation name" question is now hidden from the "go live" flow if the user has already provided us with an answer. 

~I still need to add tests to this code.~ Done

Design:
1. Log in to the [review app](https://b2xiqan2457ffygjcrxdnljdby0raotr.lambda-url.ca-central-1.on.aws/) using your staging account
2. Create a new trial service 
3. Review the new Step 3 page in English and French
4. Click the "Can't find my organisation" link to review the "other" step 3 page
5. Once the service is created, go to service settings and click "request to go live"
6. click "tell us about how you intend to use GC Notify"
7. verify that you are not asked for your organisation name again

Dev:
- Same steps as above, but run locally and verify that no changes are visible in the admin site if the feature flag is turned off (`FF_SALESFORCE_CONTACT=false` in .env)
- Verify that the organisation name(s) are saved correctly in the new `organisation_notes` column in the `services` table in the db.
- Test the new autocomplete component for accessibility (I have already done this but a second opinion is always good)


adriannelee (commented):
Design - test instructions 1 to 4
- [ ] Not sure if this is how it should be, but when I change languages from Step 3, it takes me back to Step 1 (but keeps the info I input on Step 2)
- [ ] Same as ☝️ from FR to EN
- [ ] Step 3 says "This helps us better meet your needs" and clicking "Can't find your department or organisation?" says something similar "This information helps us understand how GC Notify is being used so we can better meet your needs." Should it say something else or maybe nothing? And actually, I wonder if we should be more explicit that this information won't be shown to their recipients, since we're asking them this after asking them information that **is** shown to recipients.
- [ ] Do we already have error messages for the input fields? When I tried to submit an empty form, it let me go to another screen and says "Sorry, we’re experiencing technical difficulties" and "Try again later."
- [ ] Same as ☝️ when I type an input for "Select your department or organisation" that's not a part of the list
- The error message "This field is required" works great for the "Name of your organization" field.

Design - test instructions 5 to 7
- Looks great!

smcmurtry:
> Not sure if this is how it should be, but when I change languages from Step 3, it takes me back to Step 1 (but keeps the info I input on Step 2)
> Same as ☝️ from FR to EN

@adriannelee Yeah that is an existing bug where changing the language moves you back to Step 1: https://youtu.be/lPpMy2uJDQ8
I'd prefer to address that one in a separate PR. The form info is saved so if a user goes back to a previous step, they don't lose information they've already entered.

smcmurtry:
> When I'm on the dropdown field, I don't get the example read to me. The second field on the page, I get the example read to me when I'm on the field.

@adriannelee nice find! the id I was using for that label wasn't correct. That issue should now be fixed

smcmurtry:
>  the first time something is typed into the dropdown field, the "Select your department or organisation" gets spelled out.

I wasn't able to reproduce this one. This is what I hear when going through the page: https://youtu.be/qlFF3XcWAWU

app/main/views/add_service.py
```diff
@@ -42,9 +47,31 @@
         "header": STEP_LOGO_HEADER,
         "tmpl": "partials/add-service/step-choose-logo.html",
     },
+    STEP_ORGANISATION: {
+        "form_cls": "",
```
patheard, line 51: Now that we're no longer selecting the type of org do we still need this to be dynamic with the `ORGANISIATION_STEP_DICT` below?

app/main/views/add_service.py
```diff
@@ -105,85 +134,126 @@ def _create_service(
...
+
+
+def _renderTemplateStep(form, current_step: str, government_type: Optional[str]) -> Text:
+    WIZARD_ORDER = get_wizard_order()
+
     back_link = None
     step_num = WIZARD_ORDER.index(current_step) + 1
+
+    autocomplete_data = None
+    if current_app.config["FF_SALESFORCE_CONTACT"] and current_step == STEP_ORGANISATION:
```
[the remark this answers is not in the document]
smcmurtry: Hmm good question. In that case we could show the `step-enter-other-organisation.html` page that doesn't have the dropdown, so we still capture some organisation data to help us if the service requests to go live. That could make a good follow up PR.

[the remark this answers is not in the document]
patheard: Nice, that would work!

app/main/views/add_service.py
patheard: Ah gotcha, I'd forgotten about the case where your org isn't in the list.

patheard:
Test locally and in the PR review env and it's working as expected 🎉

adriannelee:
Huh, that's weird - I can't reproduce it either now 😅 @smcmurtry

app/templates/partials/add-service/step-enter-combined-organisation.html
```diff
@@ -0,0 +1,22 @@
+{% from "components/radios.html" import radios %}
+{% from "components/autocomplete.html" import autocomplete %}
+
+<p>{{ _("We use this information to improve GC Notify.") }}</p>
+
+<div data-module="url-typer" class="form-wrap">
+  {% call form_wrapper() %}
+
+    <div class="mb-gutterAndAHalf">
+      {{ autocomplete(form.parent_organisation_name, autocomplete_items_combined[current_lang], hint=_("For example: Treasury Board of Canada Secretariat")) }}
```
andrewleith, line 10: There seem to be many elements with the same id in the markup (`parent_organisation_name__assistiveHint`).  This technically fails HTML validation and the accessibility checker:
image

app/templates/partials/add-service/step-enter-combined-organisation.html
```diff
@@ -0,0 +1,22 @@
...
+      <a href="{{ url_for('.add_service', current_step='choose_organisation', government_type='other') }}">
+        {{ _("Not on list?") }}
```
andrewleith, line 12: Since the "Not on list" link happens before the "other group names" input, this flows so that if you choose "not on list" you likely wont get the chance to enter the "other group name".  Was that intentional?

andrewleith (changes requested):
Looks great overall.  Very cleanly implemented, nice work!

A few questions below and one accessibility issue to be fixed

app/templates/partials/add-service/step-enter-combined-organisation.html
smcmurtry: Yes that is intentional. If you click "Not on list" then we just give one textbox to fill and ask "Enter name of your group".  Anything the user entered on the previous page is ignored. Then it will be a manual process to either create a new organisation for them if the service goes live, or figure out what existing organisation to assign the service to.

app/templates/partials/add-service/step-enter-combined-organisation.html
```diff
@@ -0,0 +1,22 @@
...
+      </a>
+    </div>  
+    
+    {{ textbox(form.child_organisation_name, hint=_("For example: Canadian Digital Service"), maxlength=255) }}
```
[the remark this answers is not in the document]
smcmurtry: Nice find! I added `required=false` to that `textbox` and that has fixed the issue

tests/app/main/views/test_add_service.py
```diff
@@ -168,6 +168,91 @@ def test_wizard_flow_with_step_2_should_call_email_from_is_unique(
...
+        _expected_redirect=url_for(
+            "main.service_dashboard",
+            service_id=101,
+        ),
+    )
+    assert mock_create_service.called is True
+    assert mock_create_or_update_free_sms_fragment_limit.called is True
+
+
+def test_wizard_flow_with_step_3b_should_create_service(
```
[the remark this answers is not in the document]
smcmurtry: Added some more tests around invalid data here: https://github.com/cds-snc/notification-admin/pull/1491/commits/ee52e96ebd7df8836bb552d5f8bfee39f1ac0cd9

app/templates/partials/add-service/step-enter-combined-organisation.html
smcmurtry: Okay it looks like the 2 elements with the same id are:
Hint 1:
```
<span id="parent_organisation_name__assistiveHint" class="form-hint">
  For example: Treasury Board of Canada Secretariat
</span>
```

and Hint 2:
```
<span id="parent_organisation_name__assistiveHint" style="display: none;">
  When autocomplete results are available use up and down arrows to review and enter to select.  
  Touch device users, explore by touch or with swipe gestures.
</span>
```

Hint 1 is from me and Hint 2 is an invisible hint that comes packaged with the autocomplete component. 

The fix for this would be to change the id of Hint 1, and update the `aria-describedby` attribute on the form field to [include the ids of both labels](https://stackoverflow.com/questions/53398842/can-i-reference-multiple-elements-using-aria-describedby). 

But unfortunately I can't update the `aria-describedby` attribute on the form because that is set internally in the accessible autocomplete component 😞 

And if I change the id of Hint 1 then it won't be referenced in the `aria-describedby` attribute on the form field, so we would re-introduce the problem that Adrianne described above:
> When I'm on the dropdown field, I don't get the example read to me. The second field on the page, I get the example read to me when I'm on the field.

smcmurtry: AXE is flagging this as a "critical" issue so I've changed the id to resolve that.

andrewleith: What was the behaviour when the 2 had the same id? If you want this read by a screen reader you could also just move in inside the label and not use `aria-describedby`

app/templates/partials/add-service/step-enter-combined-organisation.html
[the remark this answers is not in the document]
andrewleith: Looks good!

andrewleith (approved):
LGTM.  Approving, just a  question about the behaviour you were going for with the `aria-describedby`.

PR #1513, opened 2023-04-03
style: make the go-live content well padding more consistent

andrewleith:
# Summary | Résumé
This PR makes the padding in the go live content well more consistent.

image


amazingphilippe:
We could flip to a column direction on mobile, or even on all breakpoints. We'd have much more space for the paragraph, and it would be easier to read. 

Changes I made in the browser for the following screenshots:
1. Moved the anchor tag outside of the p tag
2. Moved the h2 inside the flex container
3. Added the `space-y-gutterHalf` spacer utility to the flex container
4. Also added `flex-col` to the container. 

Capture d’écran, le 2023-04-03 à 14 39 33

If we want to keep the row layout, we can do the following

1. Keep the h2 outside the container
2. Add `mt-gutterHalf` and `flex-col lg:flex-row items-baseline` to the container

If we feel really fancy, we could even add some css for `align-items: last-baseline`

Capture d’écran 2023-04-03 à 14 44 54

andrewleith:
@amazingphilippe I've added the changes you asked for, let me know if I understood correctly! :)

app/templates/views/service-settings.html
```diff
@@ -13,20 +13,20 @@
...
             {{ _("We’ll be in touch within 2 business days.") }}
           </p>
         {% else %}
-          <h2 class="heading-small">{{ _('Your service is in trial mode') }}</h2>
+          <h2 class="heading-small p-0 m-0">{{ _('Your service is in trial mode') }}</h2>
           {% if current_user.has_permissions('manage_service') %}
-            <div class="flex mt-5">
-              <p class="flex-1 self-center">
+            <div class="flex mt-5 mt-gutterHalf flex-col lg:flex-row items-baseline">
+              <p class="flex-1 self-center mr-10 mb-0">
```
amazingphilippe, line 26: We should remove `self center` from the classes

amazingphilippe (changes requested):
3 Small changes and then lgtm!

andrewleith:
Removing that P tag somehow causes a test to fail, even though it isn't happening locally.  Not sure what is going on here.

amazingphilippe:
We can fix the test. The test is trying to select the anchor inside of a paragraph tags:

https://github.com/cds-snc/notification-admin/actions/runs/4621007615/jobs/8172370312#step:9:808

andrewleith:
> We can fix the test. The test is trying to select the anchor inside of a paragraph tags:
> 
> https://github.com/cds-snc/notification-admin/actions/runs/4621007615/jobs/8172370312#step:9:808
> 
> Instead of targeting `main p`, we could also add a test-id to the go-live box, to make it easier to select and make sure things don't break.

Yep for sure - except the test was passing locally for some reason even with the same selector so I was trying to do a sanity check here.  I can try the test-id, but after that I think we are spending too much time here.

andrewleith:
Good call on the testid @amazingphilippe, that takes the arbitrary HTML structure out of the test.  Everything is passing now!

amazingphilippe (approved):
Looks really good! 
We should bring this testing methodology up with the other devs

PR #1515, opened 2023-04-04
Fix/daily fragment limit validation

andrewleith:
# Summary | Résumé
This PR adds 2 validations to prevent 2 invalid states from occurring:
1. SMS daily limit being higher than the total daily limit
2. The total daily limit being lower than the SMS daily limit

## QA
- [ ] Ensure setting an SMS daily limit higher than the total daily limit displays a validation error and stops you
- [ ] Ensure setting a total daily limit lower than the SMS daily limit displays a validation error and stops you


andrewleith:
@amazingphilippe could you have another look at this?  The messages should now be correct!

whabanks (approved):
LGTM, both messages are displayed when setting the SMS limit higher than the daily limit, and when setting the daily limit lower than the SMS limit.

PR #1516, opened 2023-04-04
More menu a11y and semantic fixes

amazingphilippe:
# Summary | Résumé

Addresses https://github.com/cds-snc/notification-planning/issues/1149

This PR:

- Moves the more-menu container inside the nav, so that the links don't pop in and out of the main nav region. 
- Unwraps the div containing the aria and calling the js modules. Simply mapping all of it onto the anchor tag itself
- Updated how the icon is loaded
- Updates to the styles and toggle in the js module so that everything works fine. 

Axe dev no longer complains about the misplaced aria attribute! But there are still issues with our navigation setup, notably duplicate ids. 

You can trigger the more menu on smaller screen breakpoints, above mobile. 

Test with the main 3 browsers:
1. Safari
2. Firefox
3. Chrome

Test with Voice Over on Safari. 

Functionality checks
- Clicking on `more` can open and close the menu container
- The icon flips to show what happens when clicking on `more`
- Clicking away does nothing particular

Keyboard functionality
- While focused on `more`, ENTER toggles the menu
- While `more` is expanded, TAB navigates to the first menu item
- While `more` is collapsed, TAB skips the menu items and goes to the next logical interactive target. 
- ARROW keys do nothing particular
- ESCAPE key does nothing particular


app/templates/partials/nav/gc_header_nav.html
```diff
@@ -122,32 +122,21 @@
...
-              class="account-menu-arrow ml-2 inline arrow self-center"
-              src="{{ asset_url('images/arrow-licorice.svg') }}"
-              width="15"
-            />
+            <i aria-hidden="true" class="p-2 fa-solid fa-angle-down"></i>
           </a>
-        </div>
       </div>
       <!-- MD-LG menu and nav bar END -->
+    <div id="more-menu-container" class="relative hidden">
```
andrewleith, line 138: I'm seeing the "More" menu appear and then dissapear when I navigate pages, even on large screens.  On the templates page, it actually stayed on the page in desktop mode even though it has no items in it:
image

andrewleith (commented):
A comment about the menu flashing and appearing in one case where it shouldn't.  Otherwise all the tests are passing!

Tests:
- [ ] Clicking on more can open and close the menu container
- [ ] The icon flips to show what happens when clicking on more
- [ ] Clicking away does nothing particular
- [ ] While focused on more, ENTER toggles the menu
- [ ] While more is expanded, TAB navigates to the first menu item
- [ ] While more is collapsed, TAB skips the menu items and goes to the next logical interactive target.
- [ ] ARROW keys do nothing particular
- [ ] ESCAPE key does nothing particular

andrewleith (approved):
Looks good - no more flashing menus! a11y checks pass (except for the color contrast of the survey button, but I'm pretty sure I saw you raise that to Jeana already)
🚀

PR #1517, opened 2023-04-05
Fix/googly eyes

andrewleith:
# Summary | Résumé

This PR disables the `minifySelectors` optimisation in `cssnano` that was producing css that affected our radio button styles.


andrewleith:
This doesn't actually work.

PR #1520, opened 2023-04-05
Fix the trial service reports

smcmurtry:
# Summary | Résumé

At some point the API started returning JSON data and that messed up the CSV report. But it is easy to convert JSON data to csv, so this is an easy fix.

## Testing

Log onto the review app and try exporting a trial report.


app/main/views/platform_admin.py
```diff
@@ -300,18 +302,14 @@ def performance_platform_xlsx():
...
-        "notification_sum",
-    ]
+
+    output = io.StringIO()
+    writer = csv.DictWriter(output, data[0].keys())
+    writer.writeheader()
+    writer.writerows(data)
 
     return (
-        Spreadsheet.from_rows([headers] + data).as_csv_data,
```
smcmurtry, line 314: In my opinion we should not be putting code like this in `notification-utils`. It's like 4 lines instead of 1 to just use the csv library directly. Putting this in utils makes development changes much slower - multiple PRs and version bumps required to change a small thing.

app/main/views/platform_admin.py
```diff
@@ -300,18 +302,14 @@ def performance_platform_xlsx():
...
-        "service_name",
-        "creation_date",
-        "created_by_name",
-        "created_by_email",
-        "notification_type",
-        "notification_sum",
-    ]
+
+    output = io.StringIO()
+    writer = csv.DictWriter(output, data[0].keys())
```
smcmurtry, line 307: Just use whatever keys are present in the 1st json object in the list as the headers for the CSV. This lets the API decide what to put in the CSV and will make it easier if we want to change the data in the future.

PR #1522, opened 2023-04-06
Fix the UI vs API report

smcmurtry:
# Summary | Résumé

Fix the "Sending method: services sending via API vs. User Interface" report in platform admin. Similar fix to https://github.com/cds-snc/notification-admin/pull/1520 but I created some new functionality in the `Spreadsheet` to support JSON data.

## Test

Log in to the review app and try downloading the report.


jzbahrai (approved):
logged in an downloaded a report. lgtm!

PR #1523, opened 2023-04-06
fix caching issue

jzbahrai:
# Summary | Résumé

Fix caching issue.

# Test:
- change the url of the articles end point locally
- you should still see the page load (if there is old data -fallback) in the cache.


andrewleith (commented):
Nice! This works locally now - no more corrupted cache! 🙌 

Could you add a test for this as well? 🙏

app/articles/menu.py
```diff
@@ -30,8 +30,11 @@ def _get_nav_wp(locale: str) -> Optional[list]:
         nav_response = json.loads(cached)
     else:
         nav_response = get_content(nav_url)
-        current_app.logger.info(f"Saving menu to cache: {cache_key}")
-        redis_client.set(cache_key, json.dumps(nav_response), ex=GC_ARTICLES_NAV_CACHE_TTL)
+        if nav_response is not None:
+            redis_client.set(cache_key, json.dumps(nav_response), ex=GC_ARTICLES_NAV_CACHE_TTL)
+            current_app.logger.info(f"Saving menu to cache: {cache_key}")
+        else:
+            return []
```
jimleroyer, line 37: Should we add a log warning or error? or is this going to be logged already through `get_content`?

jzbahrai: it gets logged through get_content

andrewleith (approved):
Looks good!

PR #1538, opened 2023-04-27
Fix anchor font size within headings. 

amazingphilippe:
# Summary | Résumé

This PR fixes a bug where anchors in headings would overwrite the font size of the heading. 

Before:
Capture d’écran, le 2023-04-27 à 10 45 10

After:
image

I've removed the `text-smaller` apply that was applied on `a {...}`. In `main`, we also apply a base font size of `text-smaller`, which means all anchors in main will already have the correct font size by default. 

This only had an impact in the mobile navigation, so I targeted the font size where needed. 

There were no issues with the footer. 

1. On the terms of use page, there are some links in headings. the font size should not change within the same heading. 
2. If you can open up the production version in a new tab
3. Visually compare the font sizes with the mobile menu closed and opened. Everything should be exactly the same.


andrewleith (approved):
Looks good! Axe tests pass too!

PR #1540, opened 2023-05-01
chore: remove Scan Websites

patheard:
# Summary
Remove the calls to invoke Scan Websites.  This has been replaced by https://github.com/cds-snc/automatic-website-scanning

# Related
- cds-snc/site-reliability-engineering#858

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -93,12 +93,6 @@ jobs:
         sbom_name: "notification-admin"
         token: "${{ secrets.GITHUB_TOKEN }}"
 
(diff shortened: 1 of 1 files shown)


patheard:
Yup, right now the replacement runs on a schedule and sends the data to Sentinel like so:
https://github.com/cds-snc/automatic-website-scanning/blob/c71be03eac522aba92e86c38b9bfc9c37d9d2c4f/.github/workflows/oswasp-zap-default.yml#L33-L39

The idea is to start making this data available as Sentinel alerts in Slack and creating dashboards, but we're not quite there yet.

PR #1543, opened 2023-05-03
Remove big number link

amazingphilippe:
# Summary | Résumé

Related to issue #1 found during the bug bash.

Blue box link was removed from the bounce rate statistic on dashboard. It was linking to the same page as the other blue box. for now, we prefer to make people go through the "problem addresses" screen.

Before:

Capture d’écran, le 2023-05-03 à 14 17 09

After:

Capture d’écran, le 2023-05-03 à 14 16 05

You may have to test this one out locally if the bounce rate feature flag isn't enabled in this review app. 

1. On a service dashboard
2. Notice the blue box has no link
3. The styling should be identical to the box with a link, minus the hover and underline effects.


andrewleith (approved):
Looks great! This is a small change, but it makes the dashboard feel a lot more sane.  Nice work!

PR #1544, opened 2023-05-05
Fix problem address checkbox on CSV notifications report

andrewleith:
# Summary | Résumé

This PR refactors the bulk/CSV notifications report so that the filter problem address checkbox isn't being constantly updated like the rest of the report needs to be.

This PR separates the bulk/CSV notification report into 3 parts:
- header (ajax)
   - Schedule note
   - report download link 
- Checkbox filter (not ajax)
- notifications (ajax)


smcmurtry (changes requested):
I tried doing a bulk send on staging and then again with the review app here, and this fixes the bug for me!

I did notice one small styling regression though - the "Review address" column now moves when you check the box: https://www.youtube.com/watch?v=XkZ7Fly0CJk

andrewleith:
> I tried doing a bulk send on staging and then again with the review app here, and this fixes the bug for me!
> 
> I did notice one small styling regression though - the "Review address" column now moves when you check the box: https://www.youtube.com/watch?v=XkZ7Fly0CJk

Good catch, I'll track down what I did to cause that, thanks!

smcmurtry (approved):
The moving column issue is fixed for me now 🎉

PR #1546, opened 2023-05-10
Feat/add ab testing bounce rate

andrewleith:
# Summary | Résumé

This PR updates the bounce rate feature flag logic to show new features if the FF is on OR if the service being accessed is in a new config value, `FF_ABTEST_SERVICE_ID`.

You'll need to add a new variable to your `.env` file with a service id or 2 you want to test:
```
FF_ABTEST_SERVICE_ID=9cfb3ddd-fddd-4824-8xsx-c7d4447cc5b4
```

### A/B testing service:
image

### Other service
image


app/__init__.py
```diff
@@ -208,6 +208,7 @@ def get_locale():
 
     # allow gca_url_for to be called from any template
     application.jinja_env.globals["gca_url_for"] = gca_url_for
+    application.jinja_env.globals["current_service"] = current_service
```
andrewleith, line 211: This allows all jinja templates (`.html` files) to have access to `current_service` so we can do the A/B testing based on `current_service.id`.

app/config.py
```diff
@@ -141,6 +141,7 @@ class Config(object):
     FF_SPIKE_SMS_DAILY_LIMIT = env.bool("FF_SPIKE_SMS_DAILY_LIMIT", False)
     FF_SMS_PARTS_UI = env.bool("FF_SMS_PARTS_UI", False)
     FF_BOUNCE_RATE_V1 = env.bool("FF_BOUNCE_RATE_V1", False)
+    FF_ABTEST_SERVICE_ID = [x.strip() for x in os.environ.get("FF_ABTEST_SERVICE_ID", "").split(",")]
```
jzbahrai, line 144: can we default this to []

andrewleith: If the env is omitted or empty you mean?

jzbahrai: yes

jzbahrai: I made a mistake it already defaults to []. Sorry about that.

whabanks (approved):
LGTM! Tested a service with and without being in the service list flag.

PR #1547, opened 2023-05-10
upgrade utils to 50.3.3 and rerun poetry lock

sastels:
# Summary | Résumé

admin uses the token generating and checking code from utils to verify user email addresses.

Here we switch admin to use the [new token code from utils](https://github.com/cds-snc/notification-utils/pull/204) that uses "token" for the salt when generating tokens,  but still accepts allows tokens signed with salt DANGEROUS_SALT.

After this has been in production long enough for all old tokens to expire (min1 hour) we will push another PR to admin that stops sending the salt to the utils token code. (after that, we can update the utils token code to not accept a salt as an argument)

- change your email address

# Release Instructions | Instructions pour le déploiement

So it's possible that there could be a race condition where during deployment someone tries to change their email address and their link is generated with a pod running the new code, and when they open the link they are connected to a pod running the old code. In this case the link would fail. The link would work if they tried a few minutes later when the system is fully deployed. Given how infrequent people change their email address I do not think it's worth it to try to prevent this situation.


jimleroyer (approved):
LGTM

PR #1552, opened 2023-05-16
Fix/UI issue low bounce rate

andrewleith:
# Summary | Résumé

This PR fixes the case where bounce rates are less than 0.1%.  In that case, they were being rounded down to 0 and displaying incorrectly.


smcmurtry (approved):
Looks like this fixes the bug and matches the designs. I notice that the other box just says "0.0% failed" rather than "less than 0.1%" so we may want to make these work the same way to avoid confusion.

Screenshot 2023-05-16 at 3 42 49 PM

PR #1553, opened 2023-05-16
Update the data retention limit from "3 to 90" to "3 to 7"

smcmurtry:
# Summary | Résumé

We have a bucket policy that [expires uploaded CSV files after 7 days](https://github.com/cds-snc/notification-terraform/blob/main/aws/common/s3.tf#L22).

And our security page states that we will only keep personal info for 7 days: https://notification.canada.ca/security#h-we-keep-personal-information-for-7-days

So we should change the maximum data retention period that could be set by a platform admin from 90 days to 7, to be consistent.

Before:
Screenshot 2023-05-16 at 1 59 36 PM

After:
Screenshot 2023-05-16 at 2 03 12 PM

CI tests should pass.


jimleroyer (approved):
:shipit:

PR #1554, opened 2023-05-17
Display Redis data from the bounce rate client in the dashboard

smcmurtry:
# Summary | Résumé

This PR does the following:
- adds the `RedisBounceRate` client to notification-admin
- update to the latest version of notification-utils, to make use of the new `get_total_hard_bounces` method
- update the `calculate_bounce_rate` function to get the date from the `RedisBounceRate` client
- updated tests that were mocking the stats data that was previously used, so that they now mock the `RedisBounceRate` client data instead

Tests should pass in CI.

It isn't really possible to manually test this locally or in our review environment because:
- our review environment can't connect to Redis in Staging
- even if we run admin and api locally, we don't get the delivery status back from AWS if we run locally, so we can't populate actual bounce data to test with locally

So we will have to do more of the manual testing of this change in Staging.


app/main/views/dashboard.py
```diff
@@ -295,45 +293,32 @@ def _get_daily_stats(service_id):
...
+    bounce_rate.bounce_total = bounce_rate_client.get_total_hard_bounces(service_id)
+    bounce_status = bounce_rate_client.check_bounce_rate_status(service_id)
+
+    # We only want to show a neutral color on the dashboard if we are less than minimum volume
+    # but still over the warning threshold
+    if (
+        bounce_status == BounceRateStatus.NORMAL.value
+        and bounce_rate.bounce_percentage > current_app.config["BR_WARNING_PERCENTAGE"]
+    ):
+        bounce_rate.below_threshold = True
```
andrewleith, line 317: This is much simpler than before ⭐

tests/app/main/views/test_dashboard.py
```diff
@@ -1632,118 +1632,13 @@ def test_dashboard_page_a11y(
...
-                    },
-                ],
-                1,
-                "0.1% problem addresses",
-            ),
+            (0.05, "warning", 1, "5.0% problem addresses"),
+            (0.10, "critical", 1, "10.0% problem addresses"),
+            (0.0, "normal", 0, "No problem addresses"),
+            (0.0005, "normal", 1, "Less than 0.1% problem addresses"),
+            (0.001, "normal", 1, "0.1% problem addresses"),
```
andrewleith, line 1641: This is so much cleaner, too! :)

andrewleith (dismissed):
This looks good!  You are missing the second acceptance criteria from the ticket though, " Dashboard queries have been updated to a true rolling 24 hour period".

The issue is the way it stands, if someone were to send say 200 problem emails at noon yesterday, and then they come in to their dashboard this morning, it will say 200 problem emails on the right hand side, but on the left hand side it will say "0 emails sent" because the dashboard queries are based on last night at midnight UTC or something, and NOT a true 24 hour period.

Happy to sync with you if this doesn't make sense!

whabanks (approved):
Just one thing I noticed. I had 2387 total notifications and 240 bounces. So 10.05% bounce rate and the status box did not turn from grey to red. The error seems to be in `check_bounce_rate_status` on L130, the bounce rate is a decimal and the threshold is a whole number so the comparison will always result in a status of 'normal'. Other than that it looks great!

image

smcmurtry:
@whabanks - okay I'll test that on staging after this merges and then follow up with a bug fix PR, since this is all behind a feature flag it should be low risk

PR #1557, opened 2023-05-19
Added a new color if we are below threshold but are still critical on…

jzbahrai:
# Summary | Résumé

Change the dashboard to include a neutral colour in the following case:
- We are below the threshold (in most cases it is < 1000)
- We are still over the WARNING amount

# Test in the following order:
- Send 900 mixed messages (you should see the background on the dashboard for `Review email addresses` be GREY
- Send 100 good messages ( you should see the background be green)
- Send 100 bad messages (you should see the background be red)


whabanks (approved):
LGTM, when below the threshold the neutral color is present and the other thresholds still work as intended!

PR #1559, opened 2023-05-23
chore: update renovate config to allow grouping

andrewleith:
# Summary | Résumé

This PR re-enables dep grouping to make it easier for the ops lead to bump deps.  [This config](https://github.com/cds-snc/notification-admin/pull/1481) was added as a troubleshooting step previously but is no longer necessary.


whabanks (approved):
LGTM!

PR #1565, opened 2023-05-29
Fix for bug where the x% problem addresses box is grey instead of red

smcmurtry:
# Summary | Résumé

Fixing the bug described here: https://github.com/cds-snc/notification-admin/pull/1554#pullrequestreview-1449792013

---


whabanks (approved):
LGTM! Tested in the review app and the colour changes are working as expected now.

smcmurtry:
> @smcmurtry Is it possible to get these values from utils, instead of duplicating them in admin at all? 🤔
Thanks for the suggestion @andrewleith! I made that change

PR #1567, opened 2023-05-30
Task/a11y fixes part1

andrewleith:
# Summary | Résumé

This is the first of 3 PRs to fix up a11y and HTML validation issues (see commits):
- [prefer-button](https://html-validate.org/rules/prefer-button.html) use `<button>` over `<input type="button">`
- [no-redundant-role](https://html-validate.org/rules/no-redundant-role.html): remove redundant role="button" from `summary` element
- [wcag/h32](https://html-validate.org/rules/wcag/h32.html): ensure form has a submit button
- [element-permitted-content](https://html-validate.org/rules/element-permitted-content.html): fix span that contained a `div`


app/templates/components/pill.html
```diff
@@ -17,7 +17,7 @@
 
           <div class="pill-content{% if show_count %} pill-count-item{% else %} pill-centered-item{% endif %}">
           {% if show_count %}
-            <span>{{ big_number(count, **big_number_args) }}</span>
+            <div>{{ big_number(count, **big_number_args) }}</div>
```
smcmurtry, line 20: looks good, no styling regressions that I can see

app/assets/javascripts/apiKey.js
smcmurtry, line 1: looks good, don't see any styling regressions

app/assets/javascripts/templateFolderForm.js
```diff
@@ -304,13 +304,13 @@
     this.nothingSelectedButtons = $(`
       <div id="nothing_selected">
         <div class="js-stick-at-bottom-when-scrolling">
-          <button class="button" value="add-new-template">${window.polyglot.t(
+          <button class="button" type="submit" value="add-new-template">${window.polyglot.t(
```
smcmurtry, line 307: So this is to resolve the "form missing a submit button" issue. The form actually has a submit button, it's just hidden from the user until they click a checkbox and then click "Move": 
Screenshot 2023-05-30 at 4 21 11 PM

But I can see why that would cause an a11y issue since it is hidden when the scan takes place.

smcmurtry: this change is probably fine though

smcmurtry (approved):
Looks good!

app/assets/javascripts/templateFolderForm.js
andrewleith: The issue here I believe is that none of the buttons were explicitly marked as submit buttons, this fixes that!

smcmurtry: The "Move" and "Add to new folder" are buttons with `type="submit"`, they just aren't visible until you click a few things on the page:
Screenshot 2023-05-30 at 4 34 20 PM
Screenshot 2023-05-30 at 4 34 01 PM

andrewleith: Interesting!  For some reason, I don't see type="submit" on those buttons.. i wonder why you do?

andrewleith: image

PR #1568, opened 2023-05-30
Task/a11y fixes part2

andrewleith:
# Summary | Résumé

This is the second of 3 PRs to fix up a11y and HTML validation issues (see commits):
- [wcag/h63](https://html-validate.org/rules/wcag/h63.html): use scope attribute to associate header cells and data cells 
- [valid-id](https://html-validate.org/rules/valid-id.html): ensure generated ids are valid
- [wcag/h32](https://html-validate.org/rules/wcag/h32.html): all forms must have submit buttons
- [element-permitted-content](https://html-validate.org/rules/element-permitted-content.html): a element is not allowed as a child of a `p` element


app/templates/components/table.html
```diff
@@ -8,7 +8,7 @@
     <thead class="table-field-headings{% if field_headings_visible %}-visible{% endif %}">
       <tr>
         {% for field_heading in field_headings %}
-          <th scope="col" class="table-field-heading{% if loop.first and not equal_length %}-first{% endif %}{% if not field_headings_visible %} border-0{% endif %}" width="{% if equal_length %}{{ (100 / field_headings|length)|int }}%{% endif %}">
+          <th scope="col" class="table-field-heading{% if loop.first and not equal_length %}-first{% endif %}{% if not field_headings_visible %} border-0{% endif %}" {% if equal_length %}{{ 'style=\"width: {0}%\"'.format((100 / field_headings|length)|int)|safe }}{% endif %}>
```
smcmurtry, line 11: Checked out the table layout on multiple screen widths on Chrome - looks the same as before ✅

smcmurtry (approved):
Looks good!

PR #1574, opened 2023-06-06
Increase size of 'Sign in' button in top right nav

whabanks:
# Summary | Résumé
This PR increases the size of the Sign in link on the home page, and adds an aria-label as well.

Nuance suggests adding the `alt` attribute to anchor tags but this would invalidate our HTML. Instead we use an `aria-label` on the sign in button, to correct the issue with the sign in link not "responding" to voice commands via Dragon Naturally Speaking. This is based off the assumption that the software will treat the aria-label in the same way it would treat `alt` text. 

Reference material:
[Official guide on aria-label](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA8)
[Guidelines for speech accessible html for Dragon Nuance](https://www.yumpu.com/en/document/read/3266376/guidelines-for-speech-accessible-html-for-dragon-nuance)

Testing that the link responds to voice commands is difficult as we do not have access to Dragon Naturally Speaking.
1. Visit the home page, note that the size of the "Sign in" link text is increased
2. Inspect the sign in element and note the presence of two `span` elements denoting the aria-label:

image


app/translations/csv/fr.csv
```diff
@@ -1726,4 +1726,5 @@
 "See all problem email addresses","Voir toutes les adresses e-mail problématiques"
 " problem email address"," adresse e-mail problématique"
 " problem email addresses"," adresses e-mail problématiques"
-".contact",".etat-du-systeme"
\ No newline at end of file
+".contact",".etat-du-systeme"
+"Sign in to GC Notify","Se connecter à GC Notify"
```
amazingphilippe, line 1730: In French, let's have "Se connecter à Notification GC"

app/templates/partials/nav/gc_header.html
```diff
@@ -39,6 +39,8 @@
             {{ nav_button(
               href=url_for('main.sign_in'),
               label=_('Sign in'),
+              aria_label=_('Sign in to GC Notify'),
+              class='text-base',
```
amazingphilippe, line 43: Two things:

1. We can modify the macro directly, as we use it for the language toggle and back to top link on mobile. Both of these would benefit from the size increase. 
2. Let's use the confusingly named `text-smaller` (1.9rem) utility class to match the larger font sizes we use by default (except on buttons for some reason?).

amazingphilippe (changes requested):
A few design things and a translation!

Another thing for design: Could we check this account menu button and make sure we also apply the font size increase?

image

We could also look into the mobile version of the menu, which is in a completely different file. I think it still has a smaller font size. 

Capture d’écran, le 2023-06-07 à 13 22 32

app/templates/partials/nav/gc_header.html
whabanks: Changes made! Thanks for pointing this out it's a much better idea.

app/translations/csv/fr.csv
whabanks: Corrected!

whabanks:
I also made these changes:
- Increased the account menu text size
- Increase the account menu items text size
- Increased text sizes for sign in links, and account menu items on mobile to match the text size used for the items displayed in the `Menu` navigation.

PR #1577, opened 2023-06-09
Default to first pill option if none would be selected

smcmurtry:
# Summary | Résumé

If no options in the pill menu would be selected, select the first option by default.

Before this change: https://youtu.be/FLpEaP-BMD8

After this change: https://youtu.be/_s5smLJqFAI

Visit the notifications page or a job page and try it out!

### DIFF
--- a/app/templates/components/pill.html
+++ b/app/templates/components/pill.html
@@ -7,9 +7,15 @@
   show_count=True,
   label=""
 ) %}
+  {% set ns = namespace(current_value_not_in_items=True) %}
+  {% for label, option, link, count in items %}
+    {% if current_value == option %}
+      {% set ns.current_value_not_in_items = False %}
+    {% endif %}
+  {% endfor %}
   <nav class='pill flex' aria-label="{{ label }}">
     {% for label, option, link, count in items %}
-      {% if current_value == option %}
+      {% if current_value == option or (loop.index == 1 and ns.current_value_not_in_items) %}
         <a href="{{ link }}" class='pill-selected-item' aria-current='true'>
       {% else %}
         <a href="{{ link }}" class='pill-unselected-item'>


app/main/views/dashboard.py
```diff
@@ -546,8 +546,8 @@ def get_tuples_of_financial_years(
     partial_url,
     start=2015,
     end=None,
-):
-    return (
+) -> list:
+    return list(
```
smcmurtry, line 550: This function was returning a generator and passing that to the pill component. That caused a problem because the pill component now loops through these items 2 times, and a generator can only be looped through once. Calling `list(` here fixes the problem.

app/templates/components/pill.html
```diff
@@ -7,9 +7,15 @@
...
+  {% set ns = namespace(current_value_not_in_items=True) %}
+  {% for label, option, link, count in items %}
+    {% if current_value == option %}
+      {% set ns.current_value_not_in_items = False %}
+    {% endif %}
+  {% endfor %}
   <nav class='pill flex' aria-label="{{ label }}">
     {% for label, option, link, count in items %}
-      {% if current_value == option %}
+      {% if current_value == option or (loop.index == 1 and ns.current_value_not_in_items) %}
```
andrewleith, line 18: Where does `loop` come from?

andrewleith (approved):
This behaviour works as a clean default, nice work.  

Perhaps slightly unrelated, did you think about highlighting the failed pill in the case the filter checkbox is selected?  @amazingphilippe were you working on any designs for that?

app/templates/components/pill.html
smcmurtry: It's a special variable that Jinja makes available inside a for loop: https://jinja.palletsprojects.com/en/3.0.x/templates/#for

smcmurtry:
> Perhaps slightly unrelated, did you think about highlighting the failed pill in the case the filter checkbox is selected? @amazingphilippe were you working on any designs for that?

@andrewleith here's this discussion in Figma for context: https://www.figma.com/file/w1rwZuxuh8D6SGskv0KPaK?node-id=1137:12882#465647989

PR #1582, opened 2023-06-21
Fix/update incorrect bounce links

andrewleith:
# Summary | Résumé

This PR fixes two incorrect links on the `/problem-emails` page:
1. The link to "Keep accurate contact information": it should point to `/keep-accurate-contact-information`
2. The link to bulk csv's should include the `pe_filter` param to check the checkbox by default: `pe_filter=true`

### DIFF
--- a/app/templates/views/dashboard/review-email-list.html
+++ b/app/templates/views/dashboard/review-email-list.html
@@ -47,7 +47,7 @@
             {% for job in jobs %}
                 {% if job["bounce_count"] > 0 %}
                     <li>
-                        <a href="{{ url_for('.view_job', service_id=current_service.id, job_id=job.id, status='failed') }}">
+                        <a href="{{ url_for('.view_job', service_id=current_service.id, job_id=job.id, status='failed', pe_filter='true') }}">
                             {{ job["bounce_count"] }}
                             {% if job["bounce_count"] > 1 %}
                                 {{_(' problem email addresses')}}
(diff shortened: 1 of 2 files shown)


app/templates/views/dashboard/review-email-list.html
```diff
@@ -47,7 +47,7 @@ <h2 class="heading-medium mt-8">{{ _('Problem emails over the past 24 hours') }}
             {% for job in jobs %}
                 {% if job["bounce_count"] > 0 %}
                     <li>
-                        <a href="{{ url_for('.view_job', service_id=current_service.id, job_id=job.id, status='failed') }}">
+                        <a href="{{ url_for('.view_job', service_id=current_service.id, job_id=job.id, status='failed', pe_filter='true') }}">
```
smcmurtry, line 50:
```suggestion
                        <a href="{{ url_for('.view_job', service_id=current_service.id, job_id=job.id, status='permanent-failure', pe_filter='true') }}">
```

If we want to just show problem addresses, I think this needs to be "`permanent-failure"`

smcmurtry (changes requested):
The "Keep accurate contact information" link change looks good! I have one change request for the other link.

app/templates/views/dashboard/review-email-list.html
andrewleith: 💯  Great catch, thank you!

smcmurtry (approved):
Looks good!

PR #1596, opened 2023-06-29
A11y/remove nested divs from big number

andrewleith:
# Summary | Résumé

This PR changes the sub-elements of a pill and of the big-number component to spans, in the hopes that it will bring with it better screen reader compatibility.

This PR should be visually tested anywhere the `pill` component and `big-number` components are used (including, but not limited to):
- Dashboard
- Job (bulk) page
- Template statistics page
- Admin panel: live services


smcmurtry (changes requested):
~It looks like this introduces some styling regressions - mostly around padding.~

Before:
Screenshot 2023-06-30 at 10 57 32 AM

After:
Screenshot 2023-06-30 at 10 56 42 AM

amazingphilippe:
Also noticed some changes in how the component reacts to smaller viewports. I can't say if its worse or better, but its different!
image

andrewleith:
> It looks like this introduces some styling regressions - mostly around padding.

I'm not seeing these regressions locally or in the PR review env - did you see this locally?  I'll revisit Monday though!

andrewleith:
> Just found one glitch on the dashboard where the "See all" link overlaps on the big number component.
> 
> Capture d’écran, le 2023-06-30 à 10 50 02
> Otherwise wondering if we might like to consider this work in a larger cleanup of the component states. We end up with a lot of nesting depending on what conditions the component falls into:
> 
> ```
> <a href="/services/8e3e3886-dac2-46ba-b4fa-7723197dfa69/notifications/email?status=sending,delivered,failed" class="pill-selected-item" aria-current="true">
> <span class="pill-content pill-count-item">
>   <span>
>     <span class="big-number">
>       <span class="big-number-number text-lg">
>          0
>        </span>
>     </span>
>   </span>
>   <span class="pill-label">total</span>
> </span>
> </a>
> ```

I've fixed the issue with the "see all" links.  In terms of all the nesting, I didn't address that all I did was change the markup to use spans to hopefully the issue with screen readers not reading the entire text at the same time.

PR #1599, opened 2023-07-04
Improve FR percentage and number formatting on the dashboard

whabanks:
# Summary | Résumé
- Commas are used in place of periods for decimal points
- For percentages there is now a space between the number and the % sign
- A space is used in place of a comma for the thousands place delimiter

1. Ensure the language is set to French
2. Send at least 1k emails (and sms too for completion sake). Make sure some of the sends will bounce
3. Observe that the above formatting changes for the total notifications sent and bounce rate percentages are present.


smcmurtry (changes requested):
Looks good! Just one thing I noticed: the french formatting is getting applied to the percentage in english for problem addresses:
Screenshot 2023-07-04 at 4 18 55 PM


I think this is also wha't causing the test failure in CI.

smcmurtry (approved):
Fixed now. SMS looks good too: 
Screenshot 2023-07-04 at 4 49 44 PM

PR #1600, opened 2023-07-04
Remove `FF_BOUNCE_RATE_V1`

smcmurtry:
# Summary | Résumé

Remove the feature flag for bounce rate V1 - `FF_BOUNCE_RATE_V1` and the A/B testing feature flag: `FF_ABTEST_SERVICE_ID`. This will really simplify our v1.5 release for bounce rate and we will avoid having `if` statements that have to check all three of these feature flags and decide what to do for the different cases. Many of those cases also would make no sense and could create potential bugs - like if `FF_BOUNCE_RATE_V1=False` and `FF_BOUNCE_RATE_V15=True`.

## Reviewing

- Tests should pass in CI
- Searching in the project for the string "FF_BOUNCE_RATE_V1" or "FF_ABTEST_SERVICE_ID" should turn up nothing
- The review app should look and behave the same as what is currently in staging/production, i.e. the bounce rate feature should be visible in the dashboard and elsewhere and work as expected

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -426,97 +426,48 @@ def format_notification_type(notification_type):
 
 
 def format_notification_status(status, template_type, provider_response=None, feedback_subtype=None):
-    if current_app.config["FF_BOUNCE_RATE_V1"] or current_service.id in current_app.config["FF_ABTEST_SERVICE_ID"]:
-        if template_type == "sms" and provider_response:
-            return _(provider_response)
-
-        def _getStatusByBounceSubtype():
-            """Return the status of a notification based on the bounce sub type"""
-            if feedback_subtype:
-                return {
-                    "email": {
-                        "suppressed": _("Blocked"),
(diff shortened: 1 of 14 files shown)


app/__init__.py
```diff
@@ -426,97 +426,48 @@ def format_notification_type(notification_type):
...
-        }[template_type].get(status, status)
-
-    # -----------------
-    # remove this code when FF_BOUNCE_RATE_V1 is removed
-    # -----------------
-    if provider_response:
+    if template_type == "sms" and provider_response:
         return _(provider_response)
 
+    def _getStatusByBounceSubtype():
```
smcmurtry, line 432: github totally messed up the diff here. All I did was remove the if statement, unindent the code, and remove the code below

whabanks (approved):
LGTM! Reviewed the code, checked if there was anything that was missed (there isn't yay!), and tested in the review app that functionality is intact.

PR #1602, opened 2023-07-05
devcontainer needs newer node to build

sastels:
# Summary | Résumé

devcontainer wasn't building, upgrading node to same version as prod fixed it

We do 3 things here:
- [upgrade node in the devcontainer](https://github.com/cds-snc/notification-admin/pull/1602/files#diff-24ad71c8613ddcf6fd23818cb3bb477a1fb6d83af4550b0bad43099813088686R34) 14 -> 16 
- also upgrade node for the [tests](https://github.com/cds-snc/notification-admin/pull/1602/files#diff-245392b692a50c38ecab4381b118862db514035c10983f3bd4f4b7f1f4be4692R17)
- ~prettier failed and insisted on a bunch of formatting changes, mainly adding commas to places~

- rebuild your devcontainer! 😬 
- run the app in the devcontainer


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.0.1",
   "description": "Admin front end for Notification",
   "engines": {
-    "node": "14"
```
sastels, line 6: I don't think this does anything but 16 is probably more correct than 14 🤷

jzbahrai (approved):

PR #1605, opened 2023-07-06
Switch to the new layout if `FF_BOUNCE_RATE_V15=true`

smcmurtry:
# Summary | Résumé

This PR switches to the new layout if `FF_BOUNCE_RATE_V15=True`. If `FF_BOUNCE_RATE_V15=False`, there should be no visible changes.

I believe the new guidance page doesn't exist yet, so that link will need to be updated in the future when the page is created.

`FF_BOUNCE_RATE_V15=False`
Screenshot 2023-07-07 at 9 44 24 AM

`FF_BOUNCE_RATE_V15=True`
Screenshot 2023-07-07 at 12 01 44 PM

*note that there are no problem emails shows in this screenshot because this data comes from Redis in v1.5 and I am running locally and connecting to the staging API, which doesn't work with Redis. 

1. check out the branch locally
2. set `FF_BOUNCE_RATE_V15=False` in your .env file
3. verify that there is no change in layout and functionality
4. set `FF_BOUNCE_RATE_V15=True` in your .env file
5. verify that the new layout matches figma


tests/app/main/views/test_dashboard.py
```diff
@@ -660,17 +696,117 @@ def test_dashboard_single_and_plural(
...
+        page = client_request.get("main.service_dashboard", service_id=service_one["id"], lang=lang)
+
+        assert (
+            normalize_spaces(page.select(".big-number-with-status")[0].text),
+            normalize_spaces(page.select(".big-number-with-status")[1].text),
+            normalize_spaces(page.select(".big-number-with-status")[2].text),
+        ) == expected_big_numbers_single_plural
+
+
+@pytest.mark.parametrize(
```
whabanks, line 713: Nice coverage with this one!

whabanks (approved):
LGTM! Tested locally with both the bounce rate v1 and v15 flags. Layouts are as expected!

amazingphilippe:
The component looks really good! A few things:

The "see all" cta should be under the sent last week section. I also noticed that when the help card was taller, the cta would overlap and hide the border. Moving the CTA might solve this issue, but it might reappear differently!
Capture d’écran, le 2023-07-07 à 14 53 24

Then I also noticed an issue in the admin panel. I remember fixing a similar problem a while back by removing as much divs as I could and applying the padding and gap on the divs that are always or almost always shown. 
Capture d’écran, le 2023-07-07 à 14 52 27

jzbahrai (approved):

smcmurtry:
@amazingphilippe it looks like the admin panel styling problem is a pre-existing bug, so let's fix that one in a separate PR. Staging:
Screenshot 2023-07-10 at 1 19 21 PM

smcmurtry:
I have moved the CTA for v1.5 and that also solved the overlapping issue you pointed out @amazingphilippe:

Screenshot 2023-07-10 at 1 46 08 PM

PR #1613, opened 2023-07-11
add a route that will throw an exception

sastels:
# Summary | Résumé

Add a new `/_debug?key=<DEBUG_KEY>` route that will throw an exception when the correct key is supplied (404 otherwise).

Note that we currently block this route on production but allow on staging / scratch.

- go to `/_debug?key=<DEBUG_KEY>` locally or otherwise and see the exception
- use no key or the wrong key and see a 404


app/status/views/debugging.py
```diff
@@ -0,0 +1,16 @@
...
+
+
+@status.route("/_debug", methods=["GET"])
+def debug():
+    """A route we can hit to help us debug in AWS. Currently just raises an exception to test logging.
+
+    Raises:
+        Exception: if you pass the correct key as a query param
+    """
+    if current_app.config["DEBUG_KEY"] and current_app.config["DEBUG_KEY"] == request.args.get("key", None):
```
jimleroyer, line 13: I like the idea of preventing this to execute when it is undefined and the assumption is that it will be in production. Can we make it explicit in the condition? to make sure the environment is not production?

ben851: I asked Steve to expand on this a little bit and instead of making it dependent on an environment name, make it a feature flag that can be enabled/disabled. This way we don't have to worry about environment naming conventions in future environments. We can simply set the feature flag to disabled in production.

ben851 (approved):
So spiffy! LGTM!

PR #1615, opened 2023-07-11
Fix the AJAX auto update of dashboard stats

smcmurtry:
# Summary | Résumé

Adding the `h2` back to the `ajax-block-container` somehow fixes the auto updating.

Log into the review app, bulk send through the api, observe that the total email sent auto-updates within a few seconds.


jzbahrai:
Tested this on the review app!

jzbahrai (approved):

PR #1621, opened 2023-07-21
Frontend changes for email daily limit

jzbahrai:
# Summary | Résumé

Frontend changes for email daily limit code.
You need FF_DAILY_EMAIL_LIMIT=True to test this fully.

When you switch on the feature:
1. You see the following changes (sending 1 email more than the limit)
Screenshot 2023-07-24 at 9 41 13 AM


app/main/forms.py
```diff
@@ -705,15 +705,15 @@ def __init__(self, *args, **kwargs):
 
 class MessageLimit(StripWhitespaceForm):
     def validate_message_limit(self, field):
-        if field.data < current_service.sms_daily_limit:
+        if field.data < current_service.sms_daily_limit and current_app.config["FF_EMAIL_DAILY_LIMIT"] is False:
```
andrewleith, line 708: Looks good.  We can remove this entire method when the FF is removed.

andrewleith (approved):
This looks good to me.  Tested sending in en/fr for one-off's and bulk;s and both cases were stopped once I reached the limit. 

I did notice an issue that is out of scope of this PR:

In, `too-many-messages.py`, for the 0 case, the message doesn't really make sense:
image

PR #1626, opened 2023-07-26
Fix the less than 0.1% translation

smcmurtry:
# Summary | Résumé

Bug:
Screenshot 2023-07-26 at 2 40 40 PM

This is fixed in the problem addresses screen so we can grab the code from there:
Screenshot 2023-07-26 at 2 40 55 PM

Visit this service and verify that the translation is correct: https://jecdvlhsyux3ksknhrnwxmia6m0abbyb.lambda-url.ca-central-1.on.aws/services/a2ab2d38-3edb-4316-987f-33792ecc7b7f

Screenshot 2023-07-26 at 3 05 48 PM

### DIFF
--- a/app/templates/components/big-number.html
+++ b/app/templates/components/big-number.html
@@ -93,7 +93,7 @@
             <div class="review-email-label p-6 flex justify-between items-end border-0">
               <span>
                 {% if problem_percentage < 0.1 %}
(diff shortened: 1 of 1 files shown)


amazingphilippe (approved):
Good catch

PR #1629, opened 2023-07-27
updates routes to updated slugs

amazingphilippe:
# Summary | Résumé

This PR updates the routes to GCA content. We updated the titles for these pages, so updating the slugs is also a good idea.


andrewleith (approved):
Can confirm all new routes work! ✅

PR #1633, opened 2023-08-02
Refactor multi-step contact us form

whabanks:
# Summary | Résumé
This PR refactors our contact us form to resolve issues when navigating between the various steps via modification of query string parameters. While not my cleanest work, it does resolve the issue of state not being properly maintained between the different steps, which was the crux of multiple 500 errors related to this form.

## Changelist
- Re-purposed `.../views/contact/form.html`, turning it into a template: `contact_template.html`
- Added four new forms which extend `contact_template.html` each represent a step in the contact flow
    - `contact-us.html`
    - `message.html`
    - `demo-org-details.html`
    - `demo-primary-purpose.html`
- Removed use of query parameters to maintain the state of the `current_step` and related data. We now use plain old URL slugs for this
- Fixed a bug where the `back_link` was not properly set by `_labels()` when at the `primary-purpose` step
- Now using plain Flask routes to hop between steps, and encapsulate the state required in each step.

# Motivation
After I squashed the bug outlined in the connected card it became a game of whack a mole. Bop one bug and two more are revealed, bop one more and...well you know the story.
Inspired by the multi-step form we use for creating a new template, and fully embracing the spirit of K.I.S.S, we trade some of the elegance of the previous implementation for simplicity and maintainability, with the drawback of some code duplication. There is still some room for improvement and iteration but it resolves the issues we had.

Between each test case, clear your session.
## Test 1
1. Navigate to `/contact`. 
2. Don't fill out any fields or submit 
3. Change the url slug to each of the following, repeating steps 1 and 2 observe that you are redirected back to `/contact` in each case.
    - `/contact/message`
    - `/contact/organization-details`
    - `/contact/primary-purpose`

## Test 2
1. Navigate to `/contact` 
2. Fill out the form 
    - `How can we help?` -> Any option except `Set up a demo to learn more about GC Notify`
3. Click continue
4. Navigate to both `/contact/organization-details` and `/contact/primary-purpose` from the `/contact/message` page
5. Note that you are returned to the `/contact` page

## Test 3
1. Navigate to `/contact` 
2. Fill out the form 
    - `How can we help?` -> `Set up a demo to learn more about GC Notify`
2. Without filling out the form, or submitting, navigate to `/contact/primary-purpose`
4. Note that you were redirected to `/contact`

## Test 4
1. Navigate to `/contact` 
2. Fill out the form 
    - `How can we help?` -> `Set up a demo to learn more about GC Notify`
2. Fill out the form but do not submit
3. Navigate to `/contact/message`, note that you have been redirected back to `/contact/primary-purpose`

## Test 5
1. Navigate to `/contact` 
2. Fill out the form and click continue
    - `How can we help?` -> `Set up a demo to learn more about GC Notify`
3. Navigate to `/contact/message`, note that you have been redirected back to `/contact`

## Test 6
State for the forms that do not have a button labeled `send` should remember the choices you have made once clicking continue, and back populate the fields when navigating around steps. This includes both navigation by using the back link and by editing the URL directly:
- `/contact`
- `/contact/organization-details`


smcmurtry (approved):
This is great @whabanks! I think this is a much better approach to use separate pages and redirection instead of all the confusing spaghetti logic with the query parameters.

I tried the tests you suggested and they all worked fine. I tried out the forms in both the logged-in and logged-out states. I also checked out this branch and ran it locally along with `notification-api`. I added in the Freshdesk credentials locally so that actual tickets would be created. I was able to verify that the tickets were created successfully and all the expected data was there, both logged in and logged out. 

Nice work!

PR #1634, opened 2023-08-02
fix: add confirmation message when changing service settings

andrewleith:
# Summary | Résumé

This PR adds a message to confirm service setting changes for the following settings:
- Service name
- Sending email address name
- Send emails
- Email branding
- Send text messages


whabanks (approved):
Tested on the review app, confirmations present for each setting they were added to. LGTM!

PR #1642, opened 2023-08-09
Pass PR review URL consistently to cypress tests job

andrewleith:
# Summary | Résumé

This PR takes a more consistent approach to get the PR review URL from AWS whether the lambda was created initially or updated by a subsequent commit.

### DIFF
--- a/.github/workflows/test-admin-deploy.yaml
+++ b/.github/workflows/test-admin-deploy.yaml
@@ -90,8 +90,6 @@ jobs:
             aws lambda update-function-code \
               --function-name $FUNCTION_NAME-$PR_NUMBER \
               --image-uri $REGISTRY/$IMAGE:$PR_NUMBER > /dev/null 2>&1
-            URL=$(aws lambda get-function-url-config --function-name $FUNCTION_NAME-$PR_NUMBER --query 'FunctionUrl' --output text)
-            echo "pr_url=$URL" >> $GITHUB_OUTPUT
 
           else
             aws lambda create-function \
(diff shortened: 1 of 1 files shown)


.github/workflows/test-admin-deploy.yaml
```diff
@@ -112,8 +110,7 @@ jobs:
               --principal "*" \
               --function-url-auth-type NONE > /dev/null 2>&1
 
-            URL=$(aws lambda get-function-url-config --function-name $FUNCTION_NAME-$PR_NUMBER --query 'FunctionUrl' --output text)
-            echo "URL=$URL" >> $GITHUB_ENV
+            URL="$(aws lambda create-function-url-config --function-name $FUNCTION_NAME-$PR_NUMBER --auth-type NONE | jq .FunctionUrl)"
```
patheard, line 113: Since you'll capture this URL again lower down, what if you just created the URL config at this point like so?
```suggestion
            aws lambda create-function-url-config --function-name $FUNCTION_NAME-$PR_NUMBER --auth-type NONE | jq .FunctionUrl
```
And then you could remove the `echo "pr_url=$URL" >> $GITHUB_OUTPUT` just below as well.

andrewleith: Hmm good point.  Now that I look more closely I didn't mean to get rid of the part that sets `URL` though - since that is what's used to have the bot comment on the PR.  I'll fix up both.

andrewleith: Ah I had meant to remove line 117 but keep line 116, so I'll need to continue setting `URL` so that the workflow only posts the PR review env comment once on creation.

patheard (approved):

PR #1644, opened 2023-08-10
Make cypress-tests job optional

andrewleith:
# Summary | Résumé

This PR updates the `cypress-tests` job to be optional so that if they fail the PR can still merge.  Once we are comfortable with the `cypress-tests` job we can turn this back off.


smcmurtry (approved):
Makes sense to me - we don't want this to block PRs

PR #1648, opened 2023-08-14
Fix/guidance panel preview ol css

amazingphilippe:
# Summary | Résumé

We added ordered lists to the guidance, and we just need to fix some of the css so that the preview matches what a Notify template will look like. 

Essentially, removing the padding on each `li + li` will do the trick. This change is scoped to only the `panel-preview` container, so that our other non-preview lists still show up the same. 

- Check the formatting guide pages. FR and EN
- Checkout a random GCA page that contains a list. Like `/service-level-objectives`


sastels (approved):
LGTM Phil!

There a merge conflict with index.css but after fixing that :shipit:

PR #1649, opened 2023-08-14
Fix font-size on activity page

andrewleith:
# Summary | Résumé

This PR reduces the font size a bit on the activity page to allow the display of the large number of notifications we have sent :)


app/templates/views/activity.html
```diff
@@ -16,14 +16,14 @@ <h1 class="heading-large">{{ _('Activity on GC Notify') }}</h1>
   <div class="flex flex-col lg:flex-row mb-20">
     <div class="w-full lg:w-1/2 bg-gray py-10 px-6 mb-6 lg:mb-0 lg:mr-3">
       <div class="font-bold text-title extra-tight font-sans h-20">{{ _('Total notifications sent') }}</div>
```
amazingphilippe, line 18: The French wraps to 2 lines and the spacing is a bit close. Should we increase, or remove `h-20`? Or with the line height of the big number?

Capture d’écran, le 2023-08-14 à 10 00 34

andrewleith: Fixed! Let me know if you think thats enough spacing.  The next one up is `h-30` so its a bit more...

andrewleith: image

amazingphilippe: Looks good!

amazingphilippe (approved):
Looks good!

PR #1650, opened 2023-08-17
Fix/remove fragment content

andrewleith:
# Summary | Résumé

This PR removes the word "fragment" from the website content and updates the error messages when sending one-off+bulk emails and sms.


app/main/forms.py
```diff
@@ -708,7 +708,7 @@ def validate_message_limit(self, field):
         if field.data < current_service.sms_daily_limit and current_app.config["FF_EMAIL_DAILY_LIMIT"] is False:
             raise ValidationError(
                 _l(
-                    "Your daily limit for text fragments is {sms_daily_limit}. Enter a total number of daily messages that is {sms_daily_limit} or higher."
+                    "Your daily limit for text messages is {sms_daily_limit}. Enter a total number of daily messages that is {sms_daily_limit} or higher."
```
jzbahrai, line 711: is this right? with the current_app.config["FF_EMAIL_DAILY_LIMIT"] on, we won't validate message limit?

andrewleith: I think you're right. This is for admin settings though.

andrewleith: Oh! This is the validation that stopped them from putting a higher SMS limit than the total daily limit.  Now that the total daily limit is the email daily limit this is not required.

jzbahrai (approved):

PR #1656, opened 2023-09-01
Show message limits for trial services

whabanks:
# Summary | Résumé
This PR updates the service settings page so that daily / yearly message limits for email and sms are displayed for trial services. 
It also improves a couple of tests that relied on hard coded indexes to find specific rows in the service settings page. These tests should no longer fail and need attention whenever a new row is added to the service settings.

1. Create a trial service
2. Navigate to the service settings page
3. Note that daily and yearly limits are now displayed under the `Emails` and `Text Messages` sections

### DIFF
--- a/app/templates/views/service-settings.html
+++ b/app/templates/views/service-settings.html
@@ -181,7 +181,7 @@
           )}}
         {% endcall %}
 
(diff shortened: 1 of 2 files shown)


smcmurtry (approved):
Looks good! 
Screenshot 2023-09-01 at 4 24 03 PM

PR #1658, opened 2023-09-07
Feat/add daily usage widget

andrewleith:
# Summary | Résumé

This PR adds a new component for daily limits and uses it on the dashboard to display both email and SMS limits:

image

Note: a bug was found where the daily stats used in the dashboard are not based off of midnight UTC like other places in the app.  This will be fixed in a follow-up PR.  **_EDIT:_** This bug has been fixed in https://github.com/cds-snc/notification-api/pull/1976


app/templates/views/dashboard/dashboard.html
```diff
@@ -25,6 +25,9 @@ <h2 class="heading-medium mt-8">{{ _("Scheduled sends") }}</h2>
     {% endif %}
 
     {{ ajax_block(partials, updates_url, 'weekly_totals', interval=5) }}
+    {{ ajax_block(partials, updates_url, 'daily_totals', interval=5) }}
```
smcmurtry, line 28: Since we haven't fixed the bug with the data yet - could you revert this one change? That way we could merge this PR but the new widgets won't be visible on the dashboard. Then we can make them visible once we fix the data.

app/templates/components/remaining-messages.html
```diff
@@ -0,0 +1,18 @@
+{% macro remaining_messages(header, total, used) %}
+  {% set near_limit = ((used/total) >= 0.8) %}
+  <div class="remaining-messages">
+    <div class="rm-header mb-2">{{ header }}</div>
```
smcmurtry, line 4: I did some testing of this component with a screen reader: https://youtu.be/y-hsXyCGuck
Would it be possible to make this header get read out along with the numbers below? Could be accomplished by making it an aria-label? I'm thinking it would be nice to have the screenreader read it like: "2 of 5 emails" instead of "emails"... "2 of 5".

smcmurtry: I also tested what happens when you are near the limit: https://youtu.be/_LeUcFtlgps

The hidden text is getting read out correctly - very nice! 

One suggestion - it might be better for it to say "near daily email limit" instead of "near daily limit"

smcmurtry (changes requested):
Overall this looks great! The component resizes nicely on different screen sizes. I have a couple requests/suggestions before merging.

app/templates/views/dashboard/dashboard.html
andrewleith: Yes, can do.  I will leave that until last so @amazingphilippe can take a look at the UI implementation first (if he wants to!)

app/templates/components/remaining-messages.html
andrewleith: Great ideas, I'll incorporate those now.

andrewleith: Ok I went a slightly different route for implementation but should arrive at the same result.  Instead of trying `aria-label` or `aria-describedby` I just added the contextual piece of info (i.e `emails`/`text messages`) into the message itself. After testing in voice over, I felt that a more direct message would stand out better, so I changed it to:
"1900 of 2000 emails - you are nearing the daily emails limit"

amazingphilippe (approved):
Looks good to me!

PR #1664, opened 2023-09-12
Calculate job percent complete only when the job has rows in the notification table

whabanks:
# Summary | Résumé
This PR adds a defensive check to `get_job_partials` to prevent a `DivisionByZeroError` that occurred when attempting to view a scheduled job. This issue was [uncovered when investigating an alarm](https://gcdigital.slack.com/archives/CNWA63606/p1694545734519119)

1. Schedule a bulk send via the api
2. Navigate to your service's dashboard and click the link to the job under the `Scheduled Jobs` section
3. Note that the page loads without any errors

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -430,7 +430,7 @@ def get_job_partials(job, template):
             "partials/jobs/notifications_header.html",
             notifications=list(add_preview_of_content_to_notifications(notifications["notifications"])),
             percentage_complete=0
-            if job["job_status"] == "scheduled"
+            if job["notification_count"] == 0
             else (job["notifications_requested"] / job["notification_count"] * 100),
             download_link=url_for(
                 ".view_job_csv",
@@ -451,7 +451,7 @@ def get_job_partials(job, template):
             notifications=list(add_preview_of_content_to_notifications(notifications["notifications"])),
             more_than_one_page=bool(notifications.get("links", {}).get("next")),
             percentage_complete=0
-            if job["job_status"] == "scheduled"
+            if job["notification_count"] == 0
             else (job["notifications_requested"] / job["notification_count"] * 100),
             download_link=url_for(
                 ".view_job_csv",
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -429,7 +429,9 @@ def get_job_partials(job, template):
         "notifications_header": render_template(
             "partials/jobs/notifications_header.html",
             notifications=list(add_preview_of_content_to_notifications(notifications["notifications"])),
-            percentage_complete=(job["notifications_requested"] / job["notification_count"] * 100),
+            percentage_complete=0
+            if job["job_status"] == "scheduled"
+            else (job["notifications_requested"] / job["notification_count"] * 100),
             download_link=url_for(
                 ".view_job_csv",
                 service_id=current_service.id,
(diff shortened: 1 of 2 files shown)


app/main/views/jobs.py
```diff
@@ -429,7 +429,9 @@ def get_job_partials(job, template):
         "notifications_header": render_template(
             "partials/jobs/notifications_header.html",
             notifications=list(add_preview_of_content_to_notifications(notifications["notifications"])),
-            percentage_complete=(job["notifications_requested"] / job["notification_count"] * 100),
```
sastels, line 432: So if a job is "scheduled" then the `job["notification_count"]` is zero? I would have thought that would be the (eventual) total # of notifications?

andrewleith: I tend to agree that the eventual # of notifications should be displayed on this page, but it looks like the job stats [come from here in API](https://github.com/cds-snc/notification-api/blob/main/app/dao/jobs_dao.py#L34) and at a glance they don't seem to include scheduled notifications.  I think that change might be a separate PR for a related bug (i.e. that scheduled notifications aren't part of the stats)

In terms of a guard against this page crashing, this change seems to fit the bill - I'm not sure we could do much more on the admin side if we're given a 0 for the `notification_count`.  Although maybe the check should be if `job["notification_count"]` is `0` rather than if the job is scheduled?

Out of scope for this PR, but for consistency sake, do scheduled notifications similarly count against the daily limit I wonder?

whabanks: Good point regarding the bug where scheduled notifications aren't part of the stats. When I tested locally and in staging the notifications for scheduled jobs only get created once the `scheduled_for` time has elapsed. So until the send occurs it seems we won't be able to include those notifications in the stats. It looks like this is the expected behaviour - I came across a test that indicates [no stats should be returned if a job has no associated rows in the notifications table](https://github.com/cds-snc/notification-api/blob/9f06faeeb7cf300e3ca815c20e9ce1c4af1b6bf3/tests/app/job/test_rest.py#L677).

That said, checking the `notification_count` instead of if the job is `scheduled` makes more sense as eventually there will be notifications available in the DB for the `percent complete` calculation.

To answer your question about scheduled notifications counting against the daily limit, I think the answer is `yes`. When we create a job we [increment the daily count in Redis](https://github.com/cds-snc/notification-api/blob/9f06faeeb7cf300e3ca815c20e9ce1c4af1b6bf3/app/job/rest.py#L188).

smcmurtry:
@whabanks do you know how to reproduce the original bug? I have not been able to reproduce it so I'm not sure how to test if this PR fixes the bug.

whabanks:
> @whabanks do you know how to reproduce the original bug? I have not been able to reproduce it so I'm not sure how to test if this PR fixes the bug.

I tested and debugged this with the following steps:
1. Via the API, schedule a bulk job for at least 1 day in the future
2. Query the notifications in the notification table by the job id just sent/scheduled
3. You should note that there are no notifications in the DB yet
4. Navigate to the Dashboard and click on the pending scheduled job under the `Scheduled Jobs` section
5. Should result in the 500 error.

smcmurtry:
I followed those steps on Staging and I still wasn't able to reproduce the bug:

1. via the API, I scheduled a bulk send for 48 hours in the future, for one of my services on Staging
2. Querying the notifications table for the job id returns 0 rows (as expected, since the notifications aren't added until they are about to be sent)
3. clicked on the job on my service dashboard
4. page looks fine:
Screenshot 2023-09-21 at 4 29 37 PM

smcmurtry:
I clicked the wrong button and closed this by accident! Sorry.

app/main/views/jobs.py
smcmurtry: Looking in the `jobs` table for a job I just created on staging and scheduled for 48 hours in the future:
Screenshot 2023-09-21 at 4 36 11 PM

`notification_count` is the number of notifications that will be eventually sent. @whabanks if you are able to schedule a job and you see a `notification_count` of 0 that would be a bug. I think we still need to figure out how to reproduce that before we make any code changes.

whabanks:
We were unable to reproduce this bug.

PR #1672, opened 2023-09-18
Remove a public endpoint that is unused and is causing 500 errors

smcmurtry:
# Summary | Résumé

We are no longer using this endpoint. It is available to the public, and clicking the link on the page causes 500 errors.

## Reproducing the bug
Navigate to https://staging.notification.cdssandbox.xyz/agreement/crown and click the "Download the agreement" link. You should see a 500 error page.

## Testing the fix
Navigate to the `/agreement/crown` page on the review app. You should now get a "Page not found" error.


andrewleith:
We should remove the corresponding WAF rule for this page also!

PR #1676, opened 2023-09-22
Let a user upload a recipient spreadsheet in either EN or FR

jzbahrai:
# Summary | Résumé

We need to pass a language filter to decide what error message we will show on a CSV file.
After we merge this code users will be able to upload a file with either "email address" or "adresses courriel" in either FR/ EN context. Same for phone numbers.

The best way to test this code is the following:

1. Check out this branch
2. Copy paste these files into your ADMIN AND API dev container:
 https://github.com/cds-snc/notification-utils/blob/task/fix-upload/notifications_utils/recipients.py
 https://github.com/cds-snc/notification-utils/blob/task/fix-upload/notifications_utils/columns.py

Here are the use cases I could think of:

I was using success@simulator ..you need to USE an actual email address and ensure the email reaches your inbox.

- [x]  User preference is EN and they send an EMAIL spreadsheet with an EN heading
- [x]  User preference is EN and they send an EMAIL spreadsheet with an FR heading
- [x]  User preference is EN and they send a SMS spreadsheet with an EN heading
- [x]  User preference is EN and they send a SMS spreadsheet with an FR heading
- [x]  User preference is EN , send an email without the correct heading - error message should be the EN heading
- [x]  User preference is EN, send an SMS without the correct heading - error message should be the EN heading
- [x]  User preference is FR and they send an EMAIL spreadsheet with an EN heading
- [x]  User preference is FR and they send an EMAIL spreadsheet with an FR heading
- [x]  User preference is FR and they send a SMS spreadsheet with an EN heading
- [x]  User preference is FR and they send a SMS spreadsheet with an FR heading
- [x]  User preference is FR, send an email without the correct heading - error message should be the FR heading
- [x]  User preference is FR, send an SMS without the correct heading - error message should be the FR heading
- [x] Send an empty CSV file, you should get an error in the language of your preference
- [x] Test with placeholders in the template but the file fails - error in lang of your preference
- [x] Test with placeholders in the template but file doesn't fail
- [x] Test the sample spreadsheet is generated with the correct headings depending on the context you are in.
Eg: Screenshot 2023-09-25 at 5 07 00 PM

This PR will fail until we merge: https://github.com/cds-snc/notification-utils/pull/245


andrewleith (approved):
Works perfectly, all scenarios pass, everything working as expected.

PR #1680, opened 2023-09-27
First attempt at improving a11y on the template page

smcmurtry:
# Summary | Résumé

Just an experiment for now, it's not working 100%.

TBD

### DIFF
--- a/app/templates/views/templates/_move_to.html
+++ b/app/templates/views/templates/_move_to.html
@@ -1,7 +1,7 @@
 {% from "components/radios.html" import radios, radios_nested %}
 {% from "components/page-footer.html" import page_footer %}
 
-<div id="sticky_template_forms">
+<div id="sticky_template_forms" aria-live="assertive">
   <button type="submit" name="operation" value="unknown" hidden></button>
   <input type="hidden" name="csrf_token" value="{{ csrf_token() }}" />
   {% if templates_and_folders_form.move_to.choices and template_list.templates_to_show %}
@@ -35,7 +35,7 @@
       </a>
     </div>
   </div>
-  <div class="selection-counter visuallyhidden" role="status" aria-live="polite">
+  <div class="selection-counter visuallyhidden" role="status">
     {{ _('Nothing selected') }}
   </div>
 </div>
--- a/app/templates/components/checkbox.html
+++ b/app/templates/components/checkbox.html
@@ -31,7 +31,7 @@
 
 {% macro checkbox_input(id, name, data=None, value="y") %}
   <input
-    id="{{ id }}" name="{{ name }}" type="checkbox" value="{{ value }}"
+    id="{{ id }}" name="{{ name }}" type="checkbox" value="{{ value }}" tabindex="-1"
     {% if data %}
       checked
     {% endif %}
--- a/app/templates/views/templates/_template_list.html
+++ b/app/templates/views/templates/_template_list.html
@@ -23,14 +23,11 @@
 {% else %}
   {% set display_checkboxes = current_user.has_permissions('manage_templates') and not sending_view %}
 
-  <nav id="template-list" class="{{ 'top-gutter-5px contain-floats' if (not show_template_nav and not show_search_box) else '' }}"  aria-label="{{ _('Templates and folders') }}">
-    {% if display_checkboxes %}
-      <fieldset>
-        <legend class="sr-only">{{ _("Select templates and folders")}}</legend>
-    {% endif %}
+  <table id="template-list" role="grid" class="{{ 'top-gutter-5px contain-floats' if (not show_template_nav and not show_search_box) else '' }}" aria-label="{{ _('Templates and folders') }}">
     {% for item in template_list %}
-      <div class="template-list-item {% if display_checkboxes %}template-list-item-with-checkbox{% endif %} {% if item.ancestors %}template-list-item-hidden-by-default{% endif %} {% if not item.ancestors %}template-list-item-without-ancestors{% endif %}">
+      <tr class="template-list-item {% if display_checkboxes %}template-list-item-with-checkbox{% endif %} {% if item.ancestors %}template-list-item-hidden-by-default{% endif %} {% if not item.ancestors %}template-list-item-without-ancestors{% endif %}">
         {% if display_checkboxes %}
+          <td tabindex="-1">
           {{ unlabelled_checkbox(
             id='templates-or-folder-{}'.format(item.id),
             name='templates_and_folders',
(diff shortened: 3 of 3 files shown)


whabanks:
Tested locally and in the review app LGTM!

https://github.com/cds-snc/notification-admin/assets/7444334/ca6a9763-2eb9-4968-aab0-c37efc524b6f

whabanks (approved):

whabanks (dismissed):
Looks like we'll need to change the location of the `legend` element in `_move_to.html` as it isn't permitted to have a `legend` element in a `div`.

app/templates/views/templates/_move_to.html
```diff
@@ -35,7 +34,14 @@
...
-  <div class="selection-counter visuallyhidden" role="status" aria-live="polite">
-    {{ _('Nothing selected') }}
-  </div>
+  <div class="aria-live-region visuallyhidden" aria-live="polite" role="status">
+      <div class="selection-counter">
+        {{ _('Nothing selected') }}
+      </div>
+      <legend id="aria-live-legend">
+        {{ _('Create and copy templates or create a new folder') }}
+      </legend>
```
andrewleith, line 43: This `<legend>` is causing the test failures.. I'm not sure why this would be in a `legend` here at all.  All of the content in the parent div is hidden visually, and the `aria-live` attribute makes it so that whatever content is added to this div is read out loud.  This should likely just be a `<div>`.

amazingphilippe: The `legend` was used because we grouped the buttons in a fieldset. I think a fieldset needs a legend? But the placement of it seems unusual. We could also revert to no `fieldset` and no `legend`. This isn't the typical use case for this kind of markup.

whabanks: @amazingphilippe +1 to reverting from using fieldsets. The `legend` element is only able to be placed as the first descendant of the `fieldset` element. I think a `role='group'` with an `aria-labelledby` or `aria-label` might be a better solution as long as we can dynamically set their values when the buttons change.

app/translations/csv/fr.csv
```diff
@@ -1760,4 +1760,7 @@
 "Message limits reset each night at 7pm Eastern Time","Les limites d’envoi sont réinitialisées chaque soir à 19 h, heure de l’Est"
 "Maximum 612 characters. Some messages may be too long due to custom content.","612 caractères au maximum. Certains messages peuvent être trop longs en raison de leur contenu personnalisé."
 "Too many characters","Limite de caractère atteinte"
-"New features","Nouvelles fonctionnalités"
\ No newline at end of file
+"New features","Nouvelles fonctionnalités"
+"Create and copy templates or create a new folder","Créer et copier un gabarit ou créer un nouveau dossier."
```
amazingphilippe, line 1764: Some new content for this. We don't want it to only sound like an action, we want to explain what is happening. Let's try:

`You are browsing templates. Create and copy template or add new folder. ` <> `You are selecting templates. Move templates into a new or existing folder.`

and in FR

`Vous explorez les gabarits. Créer et copier un gabarit ou créer un nouveau dossier.` <> `Vous sélectionnez les gabarits. Déplacer les gabarits dans un dossier.`

whabanks: For consistencies sake should we also change the singular version: `Move template to a new or existing folder` to something like `You are selecting a template. Move template to a new or existing folder`, or `You selected a template. Move template to a new or existing folder`?

amazingphilippe: Yeah I was wondering about this. My only concern is that if we do a message for 1 and many, we'll end up announcing two messages. 

So my train of thought was to use a generic plural form, considering also we follow this message with the number of selection. In my mind, after selecting 1 template, it reads `You are selecting templates. Move templates into a new or existing folder. 1 selection.`. Then it continues just reading `2 selections`, `3 selections`...

whabanks: Good point about two message announcements when selecting one, then a second. Going back over it with a SR it sounds redundant and pretty distracting. I'll remove the singular version and adjust the JS so we're not announcing twice for a singular selection vs multiple.

whabanks:
Quick look at how things sound after removing the singular content and using generic pluralized content for 1 selection and > 1 selection:

https://github.com/cds-snc/notification-admin/assets/7444334/ce63c532-cbf8-4f94-9b4c-31dbb3632d1d

Before when no templates exist:

https://github.com/cds-snc/notification-admin/assets/7444334/ffad9f85-ec4b-401f-9121-b4747a310ad7

After when no templates exist:

https://github.com/cds-snc/notification-admin/assets/7444334/b10efb39-d2dc-4418-81f4-775336b070da

app/translations/csv/fr.csv
```diff
@@ -1761,8 +1761,11 @@
 "Maximum 612 characters. Some messages may be too long due to custom content.","612 caractères au maximum. Certains messages peuvent être trop longs en raison de leur contenu personnalisé."
 "Too many characters","Limite de caractère atteinte"
 "New features","Nouvelles fonctionnalités"
+"You are browsing templates. Create and copy template or add new folder.","Vous explorez les gabarits. Créer et copier un gabarit ou créer un nouveau dossier."
+"Move templates to a new or existing folder","Déplacer le gabarits dans un dossier"
```
amazingphilippe, line 1765: Small typo in the FR: 
`Déplacer les gabarits dans un dossier`

amazingphilippe (changes requested):
In FR, the legend repeats itself. I think it may be because of the typo... Although I have to admit I'm not 100% sure how the javascript works with the count logic.

app/assets/javascripts/templateFolderForm.js
```diff
@@ -236,6 +237,24 @@
       },
     };
 
+    this.legendText = {
+      default: window.polyglot.t("create_template_legend"),
+      selected: (numSelected) => {
+        if (numSelected === 1) {
+          return window.polyglot.t("selecting_templates_legend");
+        }
+        return window.polyglot.t("move_template_legend");
```
amazingphilippe, line 246: Should this be `"create_template_legend"`? So that we have the full sentence? I'm not sure I understand the default state vs the browsing state.

amazingphilippe: I think I get it, but I'm not sure it works how we want it to work. I got this screen capture that shows a bug I think:

1. I go over content by title
2. When I reach the end of the buttons
3. I hear the empty selection (This could be removed when templates = 0)
4. I hear the "create_template_legend" But this string has the part about "browsing template"which I think we'd like to remove when template = 0?

whabanks: > Should this be `"create_template_legend"`? So that we have the full sentence? I'm not sure I understand the default state vs the browsing state.
By full sentence are you referring to what is spoken when there is more than 1 selected template? Here are the states as they currently exist for selections.

States:
`Default` - No selections
You are browsing templates. Create and copy template or add new folder.

`First selection`
 You are selecting templates. Move templates into a new or existing folder.

`Subsequent selections`
 x selections, Move templates into a new or existing folder.

Nice catch on the labels in the `aria-live-region` div. I can wrap the div with a check that will remove the region entirely if there are no templates or folders available so that they cannot be navigated to without "empty text" being spoken.

amazingphilippe (approved):
Sounds good to me!

PR #1683, opened 2023-09-29
Feat/skip ga for cds users

andrewleith:
# Summary | Résumé

This PR changes the updates the GA script loading to:
- only load GA scripts in production
- skip loading GA scripts if the user is a cds-snc.ca user


sastels:
basically just wrapped stuff in an `{% if %} ... {% endif %}` and added a test?

andrewleith:
> basically just wrapped stuff in an `{% if %} ... {% endif %}` and added a test?

Yeah exactly. I just wrapped all the GA code you added in a check that says if your domain ends in cds-snc.ca, then don't load the scripts so therefore we shouldn't track any GA stuff for those users.  At least that's the theory!

sastels (approved):
sounds good!

andrewleith:
I just added a check to make sure the app was running in the prod env as well so that we can exclude staging, localhost, pr review envs, etc. from sending stats into GA.

jzbahrai (approved):

PR #1684, opened 2023-10-03
A11y/fix your profile layout

andrewleith:
# Summary | Résumé

This PR updates the user profile page to remove the table layout and use flexbox instead. 

It also adds 2 a11y features to make navigating this page a little easier:
1. Adds tab stops on the informational part to make tabbing through the page an option
2. Adds hidden text on the link to inform the user which setting they would change (rather than just saying "change" for every setting) 

## Before
image

## After
image


.github/workflows/test-admin-deploy.yaml
```diff
@@ -175,6 +175,7 @@ jobs:
           CYPRESS_BASE_URL: ${{needs.deploy-test-admin.outputs.LAMBDA_URL}}
         with:
           record: false
+          config: "video=false,screenshotOnRunFailure=false"
```
andrewleith, line 178: This has nothing to do with this PR but I slipped it in anyway.  It makes the cypress tests run much faster by disabling videos and screenshots, which weren't actually working anyway.  In a future PR we could make the screenshots and recordings available.

whabanks (approved):
Looks great to me! 
One suggestion that might be out of scope for this PR is changing the way we display the phone numbers on this page. Without dashes between the phone number parts screen readers will read `6131234567` as `six billion one hundred thirty one million....`. With dashes it is ready more naturally.

amazingphilippe (changes requested):
A few changes to "wrap" this up:

1. Lets allow the "Change" link to wrap on the smallest breakpoints. I'd say let's use something like `flex-wrap sm:flex-no-wrap`. This will give a chance for more content to fit on the smallest screen, and follows a similar pattern we have on the team management screen. 
2. Lets also keep the base font size. The small grey and blue text can be hard to read. 

Some design doc for the final layout [in Figma](https://www.figma.com/file/r3GzRBjBDWvL2AwCWq1e1W/Navigating-the-site-%7C-Naviguer-le-site?type=design&node-id=8771-1801&mode=design&t=PQoY4LqAkxSbH9p1-4)

andrewleith:
> A few changes to "wrap" this up:
> 
> 1. Lets allow the "Change" link to wrap on the smallest breakpoints. I'd say let's use something like `flex-wrap sm:flex-no-wrap`. This will give a chance for more content to fit on the smallest screen, and follows a similar pattern we have on the team management screen.
> 2. Lets also keep the base font size. The small grey and blue text can be hard to read.
> 
> Some design doc for the final layout [in Figma](https://www.figma.com/file/r3GzRBjBDWvL2AwCWq1e1W/Navigating-the-site-%7C-Naviguer-le-site?type=design&node-id=8771-1801&mode=design&t=PQoY4LqAkxSbH9p1-4)

All set! Please re-review.

amazingphilippe (approved):
Looks good!

PR #1686, opened 2023-10-10
Show current folder hint when inside the root tempalate folder

whabanks:
# Summary | Résumé
Show the current folder hint for the root `Templates` folder in addition to sub-folders when selecting a template to move to a different folder.

image

1. Create a new templates folder
2. While in the root (default) template folder, select an existing template and click `Move`
3. Note that the `Current Folder` hint is displayed for the default `Templates` folder.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -262,7 +262,8 @@ def choose_template(service_id, template_type="all", template_folder_id=None):
         allow_adding_letter_template=current_service.has_permission("letter"),
         allow_adding_copy_of_template=(current_service.all_templates or len(current_user.service_ids) > 1),
     )
-    option_hints = {template_folder_id: "current folder"}
+
+    option_hints = {template_folder_id if template_folder_id is not None else "__NONE__": "current folder"}
 
     if request.method == "POST" and templates_and_folders_form.validate_on_submit():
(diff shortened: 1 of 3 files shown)


app/main/views/templates.py
```diff
@@ -262,7 +262,8 @@ def choose_template(service_id, template_type="all", template_folder_id=None):
         allow_adding_letter_template=current_service.has_permission("letter"),
         allow_adding_copy_of_template=(current_service.all_templates or len(current_user.service_ids) > 1),
     )
-    option_hints = {template_folder_id: "current folder"}
+
+    option_hints = {template_folder_id if template_folder_id is not None else "__NONE__": "current folder"}
```
whabanks, line 266: The seemingly arbitrary `__NONE__` comes from: https://github.com/cds-snc/notification-admin/blob/a11y/indicate-current-item/app/main/forms.py#L255

amazingphilippe, line 266: We're missing a translation! `_("current folder")`

We'll need to add a translation for it: 
`"current folder","dossier actuel"`

whabanks: Nice catch, fixed!

amazingphilippe (approved):
Looks good to me!

PR #1688, opened 2023-10-11
Add SR label to identify which API key is being revoked

whabanks:
# Summary | Résumé
This PR adds a `sr-only` `span` tag along side the revoke link found on the API keys page. This will assist screen reader users identify which API key will be revoked.

1. Create several API keys or use existing keys if you have them.
2. Turn on a screen reader (cmd + F5 to activate VoiceOver)
3. Put focus on the `Revoke` links
4. Note that the screen reader reads back `Revoke API key <API_Key_Name>` / `Révoquer Clé API <API_Key_Name>` for each Revoke link put in focus / selected

### DIFF
--- a/app/templates/views/api/keys.html
+++ b/app/templates/views/api/keys.html
@@ -84,7 +84,7 @@
         {% endcall %}
       {% else %}
         {% call field(align='right', status='error') %}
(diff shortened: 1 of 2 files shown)


andrewleith (approved):
Looks great! 🚀

PR #1690, opened 2023-10-11
A11y/fix img tags

andrewleith:
# Summary | Résumé

This PR fixes various issues with `img` tags throughout the front end.


whabanks (approved):
Checked it out in the review app, looks great to me!

PR #1691, opened 2023-10-11
Remove unused gulp-css-url-adjuster dependency

whabanks:
# Summary | Résumé
This PR is part 2 of 2 ([part 1](https://github.com/cds-snc/notification-admin/pull/1667)) to resolve a [critical security vulnerability](https://github.com/cds-snc/notification-admin/security/dependabot/41) with versions of `lodash` that are < 4.5.0.

## Why?
The dependency tree:
```
gulp-css-url-adjuster
|__ gulp-util
|____ lodash (< v4.5.0)
```
`gulp-css-url-adjuster` is dependent on [`gulp-util`](https://www.npmjs.com/package/gulp-util) which has been deprecated. The authors [recommend moving away from using it](https://medium.com/gulpjs/gulp-util-ca3b1f9f9ac5). Additionally the authors ask that if your project has a dependency which uses `gulp-util` to create an issue in the dep's repo recommending it's removal and offering replacements for certain API calls.

Initially this was my plan, however, there has not been a commit in the repo for [over 8 years](https://github.com/trentearl/gulp-css-url-adjuster) and there are issues that have been open for 6+ years. I expect that if we opened an issue it would receive the same treatment as maintenance of the dep seems to have halted.

Lastly, while we require the dependency in our `gulp.js` we do not appear to call or utilize it, and removing it has had no apparent affect on functionality.

# Testing
- Build the dev container, there should be no errors.
- Run the app and poke around the site for anything CSS related that seems out of place or incorrect.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,7 +12,6 @@ plugins.addSrc = require("gulp-add-src");
 plugins.babel = require("gulp-babel");
 plugins.base64 = require("gulp-base64-inline");
 plugins.concat = require("gulp-concat");
-plugins.cssUrlAdjuster = require("gulp-css-url-adjuster");
 plugins.faMinify = require("gulp-fa-minify");
 plugins.jshint = require("gulp-jshint");
 plugins.prettyerror = require("gulp-prettyerror");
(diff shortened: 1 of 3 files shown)


andrewleith (approved):
Nice!  Good idea on letting GDS know, we have a shared slack channel with them as well.  Could ask @jimleroyer his thoughts on the best way to share this.

As for this change, I compared the contents of the generated assets in this branch (all the files in `app/static`) with those in the main branch and they are identical, so I believe you are correct in that we are not using this anywhere and removing it is innocuous! 🎉

PR #1694, opened 2023-10-31
Creating temporary place to pull cloudfront assets

ben851:
# Summary | Résumé

Grabbing all "static" assets from the AWS cloudfront s3 and putting them in their own folder for now. We will need to revisit this to sort out the asset folder organization.

N/A


andrewleith:
Should the folder for the images be `app/assets/cloudfront/static/images` ?

If we want to add a new static image, would we add it to both `/app/assets/images` as well as this new folder?

ben851:
> Should the folder for the images be `app/assets/cloudfront/static/images` ?
> 
> If we want to add a new static image, would we add it to both `/app/assets/images` as well as this new folder?

The folder structure is taken directly from the S3... It seems as though the "images" part gets removed through some magic in admin  (gulp maybe?). This is part of the reason why I'm creating this interim solution vs fixing the assets directly in images... I'm not too sure what's going on!

Does it mean you need it in two spots? If you point your image reference in source to the assets/cloudfront/static part, then no 😆  If you keep it with the "old" files then yeah it would need to be in both. 

We could/should get a bug bash going for this to sort it out in the near future..

jimleroyer (approved):
LGTM even though it does not have the magic wand sparkles.

PR #1696, opened 2023-11-02
A11y/fix titles add emptystate

andrewleith:
# Summary | Résumé

This PR fixes the titles on the notifications reports pages and adds some empty states to various pages.  Changes:
1. Update main notification report title; add empty state (`/services/{sid}/notifications/email?status=sending,delivered,failed`)
2. Update bulk notification report title (`/services/{sid}/jobs/{jid}`) 
3. Add empty state for sent messages on bulk screen (`/services/{sid}/jobs`)
4. Add empty state for scheduled messages on bulk screen (`/services/{sid}/jobs`)

### Screenshots
1. image
2. image
3. image
4. image


app/assets/images/empty-bird.svg
jimleroyer, line 1: Love the bird 🐦

amazingphilippe (changes requested):
3 areas where I think we need changes or better design docs.

1. The jobs page `{sid}/jobs`
If there are no scheduled jobs, don’t show “sent messages” section at all, unless some have been sent in the past. Ideally we only need one empty state per screen
image

2. The delivery report for a scheduled job that hasn't been sent yet
This screen doesn’t work as well for scheduled jobs. We can design something else that would work better. 
Delivery report seems out of place: nothing has been sent yet.
I don’t know what is being sent, maybe we could get a list, like on the review page?
We can think of a better h1
We can remove the pill nav
We can architect the content a bit more clearly. 
We can create familiarity with the delivery report by putting some info near the heading: scheduled by, starting, origin
image

3. The delivery report after the retention period
Not quite the right content.
List: sent by, started, origin
Tweak screen for after the 7 days. This can communicate the same broad strokes, but the pill nav serves no purpose
image

andrewleith:
> 3 areas where I think we need changes or better design docs.
> 
> 1. The jobs page `{sid}/jobs`
>    If there are no scheduled jobs, don’t show “sent messages” section at all, unless some have been sent in the past. Ideally we only need one empty state per screen

Can the user get to this page when its in this state? When there are no jobs scheduled and none have been sent, this section doesnt appear on the dashboard at all as far as I'm aware.  If that is the case, I don't think we need to spend time on this.

> 2. The delivery report for a scheduled job that hasn't been sent yet
>    This screen doesn’t work as well for scheduled jobs. We can design something else that would work better.
>    Delivery report seems out of place: nothing has been sent yet.
>    I don’t know what is being sent, maybe we could get a list, like on the review page?
>    We can think of a better h1
>    We can remove the pill nav
>    We can architect the content a bit more clearly.
>    We can create familiarity with the delivery report by putting some info near the heading: scheduled by, starting, origin

Yeah this doesn't quite make sense.  The page title should be "Scheduled send" when it hasn't been sent yet to be consistent with the main dashboard and I agree the pill nav should be axed.  How do you want to lay this out?  There are 5 piece of info we have to display here:
1. who uploaded it
2. the file name
3. date it was uploaded
4. date its scheduled for
5. template used

> 3. The delivery report after the retention period
>    Not quite the right content.
>    List: sent by, started, origin
>    Tweak screen for after the 7 days. This can communicate the same broad strokes, but the pill nav serves no purpose
>    image

So we have all the totals but can't click through because we don't have the details?

amazingphilippe (changes requested):
A few observations
- The delivery report info at the top should be expanded so that we have the file name, or API request name somewhere. Similar to the scheduled messages screen.
Capture d’écran, le 2023-11-21 à 08 46 38
- If we use a table layout, we can drop the colons
Capture d’écran, le 2023-11-21 à 08 44 48
- And we should have the missing transaltions in the coming days. 

Otherwise looks good!

whabanks (approved):
LGTM overall! Just a couple minor comments / questions.

app/main/views/jobs.py
```diff
@@ -106,11 +106,18 @@ def view_job(service_id, job_id):
     partials = get_job_partials(job, template)
     can_cancel_letter_job = partials["can_letter_job_be_cancelled"]
 
+    # get service retention
+    svc_retention_days = [
+        dr["days_of_retention"] for dr in current_service.data_retention if dr["notification_type"] == template["template_type"]
+    ]
+    svc_retention_days = 7 if len(svc_retention_days) == 0 else svc_retention_days[0]
```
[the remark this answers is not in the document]
andrewleith: For the hard coded 7?  Yeah, you are right, this should be the default, `ACTIVITY_STATS_LIMIT_DAYS`.  I'll fix this!

app/templates/views/jobs/job.html
```diff
@@ -3,26 +3,95 @@
 {% from "components/ajax-block.html" import ajax_block %}
 {% from "components/page-footer.html" import page_footer %}
 {% from "components/problem-email-checkbox.html" import problem_email_checkbox %}
+{% from "components/empty-list.html" import empty_list  %}
 
 {% block service_page_title %}
-  {{ uploaded_file_name }}
+  {% if job.job_status == 'scheduled' %}
+    {{ _('Scheduled messages') }}
```
amazingphilippe, line 10: Current: title will be "Envois programmés" (plural form)
Designs: title should be "Envoi programmé" (singular)

I'm not sure if this will conflict with other instances of this string. If this is the only instance, then we can make the change in translation csv.

amazingphilippe (commented):
Just one note about the title of the single scheduled job screen. And what we discussed about the delivery report details logic. After this we're good!

app/templates/views/jobs/job.html
amazingphilippe: Not sure! French focuses more on a job, English focuses more on the messages. Both work!

amazingphilippe (approved):
Review app isn't showing the change, but looking at the fr csv file I can see the correct translation, so I'll just approve this!

PR #1698, opened 2023-11-02
Add route for archived terms of use

amazingphilippe:
# Summary | Résumé

Added a new route for another archived version of the terms of use.

Yes. I think the WAF rules already cover this pattern. 

1. Go to review app
2. Navigate to /terms-202304
3. Check that the h1 is correct. Should be the archived version from April 2023
4. Click the language toggle
5. Check that you get the same archived date in the title


amazingphilippe:
We don't actually need to do this. closing the PR.

PR #1699, opened 2023-11-03
Remove non-printable characters from URL before redirecting

andrewleith:
# Summary | Résumé
Remove non-printable characters from URL before redirecting inside of `/set-lang`


tests/app/test_utils.py
```diff
@@ -694,6 +695,23 @@ def test_set_lang_bad_route(
     assert page.select_one("h1").text.strip() == "Page could not be found"
 
 
+def test_set_lang_bad_route_with_control_characters(
```
jimleroyer, line 698: Do we have regression tests that cover valid routes that should not be blocked?

app/main/views/set_lang.py
```diff
@@ -20,4 +20,7 @@ def set_lang():
         # redirect to main page in case of an invalid redirect
         url = url_for("main.show_accounts_or_dashboard")
 
+    # remove non-printable characters from url
+    url = "".join(ch for ch in url if ch.isprintable())
```
jimleroyer, line 24: This feels like a sanitization library exists for this, but that should be enough.

jimleroyer (approved):
I left a few comments but it seems good to avoid the bug immediately.

app/main/views/set_lang.py
andrewleith: Yeah I thought the same thing.  I looked briefly and found Bleach and Schema but neither seemed to work for what we need here.  I prioritized doing it quickly vs. looking in depth though, so there likely is something out there.

tests/app/test_utils.py
andrewleith: No - but if we have an existing route with non-printable characters, that would throw an error anyway.  I can add a todo to make a more thorough test here for existing routes though.

PR #1707, opened 2023-11-07
Owasp/add permissions policy header

andrewleith:
# Summary | Résumé

This PR add the `Permissions-Policy` http header to all requests to disable certain technologies/features that we don't use.


app/__init__.py
```diff
@@ -641,6 +641,10 @@ def save_service_or_org_after_request(response):
 def useful_headers_after_request(response):
     response.headers.add("Referrer-Policy", "strict-origin-when-cross-origin")
     response.headers.add("Strict-Transport-Security", "max-age=63072000; includeSubDomains; preload")
+    response.headers.add(
+        "Permissions-Policy",
+        "accelerometer=(), ambient-light-sensor=(), autoplay=(), battery=(), camera=(), display-capture=(), encrypted-media=(), geolocation=(), gyroscope=(), keyboard-map=(), magnetometer=(), microphone=(), midi=(), navigation-override=(), payment=(), picture-in-picture=(), publickey-credentials-get=(), screen-wake-lock=(), sync-xhr=(), usb=(), web-share=(), xr-spatial-tracking=()",
```
jimleroyer, line 646: @andrewleith Are empty parenthesis mean we won't need these?

andrewleith: Yeah the features are all opt-in, so if it isn't listed, you allow it.

We are saying here that we don't use any of these particular features, so explicitly disable them.

jimleroyer (approved):

PR #1709, opened 2023-11-07
Owasp/csp wildcard directive

whabanks:
# Summary | Résumé
- Add qualtrics to `form-action` header
 - Added `Upgrade-Insecure-Requests` header (wildcard fix maybe?)
 - Added a nonce to the `style-src` header
 - Added nonces to each `<style>` tag


whabanks:
I screwed it up!

PR #1710, opened 2023-11-07
Add qualtrics to form-action header

whabanks:
# Summary | Résumé
- Add Qualtrics to form-action header
- Added Upgrade-Insecure-Requests header (wildcard fix maybe?)
- Added a nonce to the style-src header
- Added nonces to each <style> tag

### DIFF
--- a/app/templates/main_template.html
+++ b/app/templates/main_template.html
@@ -26,10 +26,14 @@
   <link
     href="https://fonts.googleapis.com/css?family=Lato:400,700,900&display=swap"
     rel="stylesheet"
+    integrity="sha384-kVxP/aXwg+PkN511FuqryM/THiiQlE94Vnv1Su+pNFJY+4yvw+4vncZFNZbgY7UL"
+    crossorigin="anonymous"
   />
   <link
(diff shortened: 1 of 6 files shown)


app/templates/main_template.html
```diff
@@ -26,10 +26,14 @@
   <link
     href="https://fonts.googleapis.com/css?family=Lato:400,700,900&display=swap"
     rel="stylesheet"
+    integrity="sha384-kVxP/aXwg+PkN511FuqryM/THiiQlE94Vnv1Su+pNFJY+4yvw+4vncZFNZbgY7UL"
```
jimleroyer, line 29: oh interesting, that's one of the sha attribute you referred to?

jimleroyer (approved):
LGTM :shipit:

app/templates/main_template.html
whabanks: Yes, I went with a hash here because [according to W3C](https://w3c.github.io/webappsec-csp/#is-element-nonceable) only the `<script>` and `<style>` tags are valid targets for a `nonce` attribute.

PR #1712, opened 2023-11-08
Added report-to header to CSP

jimleroyer:
# Summary | Résumé

Added the `report-to` header to the CSP headers. This is a more modern version of `report-uri` which we do support already, except the latter is being deprecated. The policy is meant to define an endpoint to report the CSP violations to.

As the `report-to` policy is not yet supported in all browsers, we can use the `report-uri` policy in conjunction as instructed [in the Mozilla documentation on that topic](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/report-uri). The browsers which do not support `report-to` yet will hence use `report-uri` and those that do will ignore `report-uri`.

* Ask the security team if they still receive report of GCNotify's CSP violations.

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -647,10 +647,10 @@ def useful_headers_after_request(response):
     response.headers.add("Upgrade-Insecure-Requests", "1")
     nonce = safe_get_request_nonce()
     asset_domain = current_app.config["ASSET_DOMAIN"]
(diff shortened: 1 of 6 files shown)


whabanks (changes requested):
Looks good! Just need to run the formatter and possibly update the two tests in [test_headers.py](https://github.com/cds-snc/notification-admin/blob/owasp/csp-directives/tests/app/main/views/test_headers.py)

whabanks (approved):

PR #1713, opened 2023-11-08
Refactor fontawesome

andrewleith:
# Summary | Résumé

This PR refactors the way we get and use font awesome in order to avoid it inserting inline styles into our app and causing us to fail some OWASP tests.

Once this PR is merged, the new way to load an icon is to insert it in `app/assets/javascripts/fontawesome.js` (instructions included there)


app/assets/javascripts/fontawesome.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * This file loads in font awesome for use in Notify
+ * We only load the icons we use to keep the bundle size down. To add an icon:
+ * - import if from @fortawesome/free-solid-svg-icons
+ * - include it in the library.add call
+ */
```
jimleroyer, line 6: Great! 👏

app/templates/main_template.html
```diff
@@ -165,8 +165,6 @@
 {% block page_script %}
 <script nonce="{{ request_nonce }}" type="text/javascript" src="{{ asset_url('javascripts/main.min.js') }}"></script>
 <script nonce="{{ request_nonce }}" type="text/javascript" src="{{ asset_url('javascripts/all.min.js') }}"></script>
-<script defer nonce="{{ request_nonce }}" type="text/javascript" src="{{ asset_url('javascripts/solid.min.js') }}"></script>
-<script defer nonce="{{ request_nonce }}" type="text/javascript" src="{{ asset_url('javascripts/fontawesome.min.js') }}"></script>
```
jimleroyer, line 169: These libraries were basically the ones loading/prepping up our icons? which we now handle?

gulpfile.js
```diff
@@ -111,7 +89,7 @@ const javascripts = () => {
       plugins.addSrc.prepend([
         paths.src + "javascripts/main.min.js",
         paths.src + "javascripts/scheduler.min.js",
-        paths.src + "fontawesome/js/fontawesome.min.js",
+        // paths.src + "fontawesome/js/fontawesome.min.js",
```
jimleroyer, line 92: Do we want to keep this commented?

andrewleith: Nope I'll axe that, thanks.

app/templates/main_template.html
andrewleith: Yes exactly.  `solid.js` contained the icons themselves and some logic, and `fontawesome.js` was FA itself.  Both are now brought in via npm.

package.json
```diff
@@ -22,6 +22,8 @@
   "homepage": "https://github.com/cds-snc/notification-admin#readme",
   "dependencies": {
     "@babel/plugin-proposal-class-properties": "^7.14.5",
+    "@fortawesome/fontawesome-svg-core": "^6.4.2",
+    "@fortawesome/free-solid-svg-icons": "^6.4.2",
```
jimleroyer, line 26: fortawesome? I thought this might be a typo but checking, it's really fortawesome!

jimleroyer (approved):
LGTM! Thank you for including the comments in the new js file. I wonder if we should add a wiki and a link to the wiki along with this, or if it's sufficient as-is? Not sure, not blocking the PR, all good to do when you're ready.

whabanks (approved):
Looks good to me! Just a couple of conflicts to resolve (that was me >_>)

package.json
andrewleith: I thought that was weird as well!

PR #1716, opened 2023-11-08
Removing inline style violating recent CSP inline-style policy

jimleroyer:
# Summary | Résumé

A custom inline style attribute violated recent CSP changes that makes inline styling more restrictive. After review, it seems that custom inline style does not contribute to any style change, as it would be overridden by other classes applying on the same element.

That style affected the ruler on top of the mobile menu, making it 4 pixel. But in reality, it was always 1px by default (it seems -- @amazingphilippe to confirm).

Tested locally through the Chrome debugger. 

* Bring up the mobile responsive view.
* Notice any change: there should not be any.


app/templates/partials/nav/gc_header_nav.html
```diff
@@ -8,7 +8,7 @@
   {% set long_lang_switch = 'Français' %}
 {% endif %}
 
-<hr class="m-0 md:hidden bg-blue" style="height: 4px"/>
+<hr class="m-0 md:hidden bg-blue"/>
```
amazingphilippe, line 11: The height had an effect on the thickness of the hr. We can remove the style attribute and add the `h-2` class to make it `0.5rem` which in our case will be 5px. We can't easily make 4px with our current tokens. 

Don't forget to recompile tailwind!

jimleroyer: Did you test via the chrome debugger to apply/re-apply the style? We did that and while we saw a visual difference when we re-applied, on page load, the height difference isn't applied. We thought that Chrome might do some magic to show it to us but that's not how it would work. It would be better to sync about it (we're on gathertown).

whabanks: Phil and I took a closer look at this and it seems the `4px` does indeed apply. The theory is that we had found this in the midst of tinkering with the `style-src` header, so the inline style here may have been blocked by the policy. We can probably add that `4px` to [style.css](https://github.com/cds-snc/notification-admin/blob/main/app/assets/stylesheets/tailwind/style.css) and remove the inline style.

app/templates/partials/nav/gc_header_nav.html
```diff
@@ -8,7 +8,7 @@
   {% set long_lang_switch = 'Français' %}
 {% endif %}
 
-<hr class="m-0 md:hidden bg-blue" style="height: 4px"/>
+<hr class="m-0 md:hidden bg-blue h-2"/>
```
jimleroyer, line 11: 🕺

PR #1720, opened 2023-11-14
A11y fable/lexical prototype

andrewleith:
# Summary | Résumé

PR to create a prototype for lexical.


app/main/views/styleguide.py
```diff
@@ -27,3 +27,54 @@ class FormExamples(Form):
...
+
+
+Step 7
+
+Add a divider
+
+
+Step 8
+
+Link
```
andrewleith, line 76:
```suggestion
Insert a hyperlink to Canada.ca
```

PR #1722, opened 2023-11-14
Set the auto RUM insertion at global config level instead

jimleroyer:
# Summary | Résumé

Because we want to disabled browser's instrumentation for the admin and the previous way of doing it via the agent Python configuration seemed to fail:

* Disable the RUM auto-insertion via the global `newrelic.ini` configuration file.
* Remove the previously Python agent RUM configuration. Most likely this setting didn't work because it needs to be applied at the web request level, and hence it wouldn't apply on a global configuration level.

We need to test in an env where NewRelic is present and enabled.

### DIFF
--- a/app/newrelic.ini
+++ b/app/newrelic.ini
@@ -1,0 +1,209 @@
+# ---------------------------------------------------------------------------
+
+#
+# This file configures the New Relic Python Agent.
+#
+# The path to the configuration file should be supplied to the function
+# newrelic.agent.initialize() when the agent is being initialized.
+#
+# The configuration file follows a structure similar to what you would
+# find for Microsoft Windows INI files. For further information on the
+# configuration file format see the Python ConfigParser documentation at:
+#
+#    http://docs.python.org/library/configparser.html
+#
+# For further discussion on the behaviour of the Python agent that can
+# be configured via this configuration file see:
+#
+#    http://newrelic.com/docs/python/python-agent-configuration
+#
+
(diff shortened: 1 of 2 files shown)


app/newrelic.ini
```diff
@@ -0,0 +1,209 @@
...
+# To stop specific errors from reporting to the UI, set this to
+# a space separated list of the Python exception type names to
+# ignore. The exception name should be of the form 'module:class'.
+error_collector.ignore_errors = app.v2.errors:BadRequestError jsonschema.exceptions:ValidationError
+
+# Browser monitoring is the Real User Monitoring feature of the UI.
+# For those Python web frameworks that are supported, this
+# setting enables the auto-insertion of the browser monitoring
+# JavaScript fragments.
+browser_monitoring.auto_instrument = false
```
jimleroyer, line 174: This is the setting to change to disable auto instrumentation on the browser side.

app/newrelic.ini
andrewleith, line 1: How is this loaded?  Does new relic automatically look for it here? I see in your manifests PR you create an ENV variable for this path, but I don't see anywhere its used?

andrewleith (approved):
Tested locally and everything works without errors.  Need to get it to staging to give it a good test.

app/newrelic.ini
jimleroyer: Yes the NewRelic Python agent should look for it. We do have this setup and working for the lambda IIRC.

PR #1725, opened 2023-11-16
Moved NewRelic config file to project root folder

jimleroyer:
# Summary | Résumé

The environment variable `NEW_RELIC_CONFIG_FILE` target the `newrelic.ini` configuration file to be in `/app/` of the Docker image yet it is currently sitting in `/app/app/` (first app is the Docker app folder and second is the Python app root folder).

Hence moving the configuration file at the top of the project's root will move it where the environment variable expects it to be.


sastels (approved):
too many "app"s 😅

PR #1726, opened 2023-11-16
Added bam.nr-data.net to connect-src exception for NewRelic

jimleroyer:
# Summary | Résumé

NewRelic needs to be listed in the list of allowed domain of [`connect-src`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/connect-src) to push through following API its monitoring metrics:

[<a>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a) [ping](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#ping),
[fetch()](https://developer.mozilla.org/en-US/docs/Web/API/fetch),
[XMLHttpRequest](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest),
[WebSocket](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket),
[EventSource](https://developer.mozilla.org/en-US/docs/Web/API/EventSource), and
[Navigator.sendBeacon()](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon).

## [NewRelic explanation](https://forum.newrelic.com/s/hubtopic/aAX8W0000008YtrWAE/relic-solution-what-is-bamnrdatanet-new-relic-browser-monitoring) on bam.nr-data.net

So what is [bam.nr-data.net](https://forum.newrelic.com/s/hubtopic/aAX8W0000008YtrWAE/void(0))?

Here’s the short explanation: The Browser application monitoring agent transmits data to New Relic’s data collection servers via the domain [bam.nr-data.net](https://forum.newrelic.com/s/hubtopic/aAX8W0000008YtrWAE/void(0)). Of course, there’s more to the story than that. If you’re the kind of person that needs to know more, then read on!

How does New Relic Browser monitoring work?

Depending on your subscription level, [New Relic Browser](https://forum.newrelic.com/s/hubtopic/aAX8W0000008YtrWAE/void(0)) can monitor a variety of details about everything from page load timing to time spent in the front end, time spent in the back end, geography, and browser type.

To collect the data, New Relic Browser uses JavaScript elements pasted or injected into your webpages, typically as part of the HEAD of the page, containing configuration details and essential browser environment instrumentation. Once the page finishes loading, an additional script is downloaded from a CDN server. This additional script processes collected data and reports it back to New Relic via [bam.nr-data.net](https://forum.newrelic.com/s/hubtopic/aAX8W0000008YtrWAE/void(0)) so that you can see the data in your New Relic account. You can find more details about how Browser instrumentation works in [the documentation](https://forum.newrelic.com/s/hubtopic/aAX8W0000008YtrWAE/void(0)).

You should also know that New Relic Browser only collects performance data, as explained in [this document](https://forum.newrelic.com/s/hubtopic/aAX8W0000008YtrWAE/void(0)). It does not collect any data used or stored by the monitored application unless you explicitly configure it to do so.

We need to push to staging environment to test it (where NewRelic is enabled).


whabanks (approved):
🚀

PR #1735, opened 2023-12-21
A11y. Table structure. Spreadsheet example

amazingphilippe:
# Summary | Résumé

- Add an empty `<td>` instead of nothing on the top-left header. This cell represents the corner between the numbered index, and the headers. Like in an excel document. The idea is that you can't have an empty TH, unless you place an empty TD inside of it. 

1. In any service
2. Navigate to any template
3. Select send to many recipients
4. Expand the spreadsheet example
5. Inspect the example table.
6. The top left header should have an empty TD inside of it.


amazingphilippe:
The failing test is something about the switch to TH for the row indexes. I think TH with a scope of row is what we need all the time. I know there is another component for TH rows.... so I guess we should figure that out before fixing the tests.

jzbahrai (approved):

PR #1745, opened 2024-01-26
Prevent naming variations of .env files to be submited to git

jimleroyer:
# Summary | Résumé

Prevent naming variations of .env files to be submited to git.

Create a file named `.env.staging` locally and check if it can get submitted with git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -74,6 +74,7 @@ npm-debug.log
 environment.sh
 .env
 .envrc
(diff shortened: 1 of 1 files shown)


whabanks (approved):
LGTM, tested locally with `.env.staging` and other variations. I like the whitelist approach here!

sastels (approved):
💯

PR #1746, opened 2024-01-29
Clarify in email hint text that shared inbox is not allowed

whabanks:
# Summary | Résumé
- Changed the password field's validation message to clarify what a hard to guess password is
- Fixed the aria-invalid attribute on text boxes
- Add cypress tests for registration page
- Use jsrsasign to generate JWTs for admin api access in cypress tests


tests_cypress/cypress/Notify/Admin/Pages/CreateAccountPage.js
```diff
@@ -0,0 +1,67 @@
+const { recurse } = require('cypress-recurse')
+const ADMIN_COOKIE = "notify_admin_session";
```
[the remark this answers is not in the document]
whabanks: Fixed

tests_cypress/package.json
```diff
@@ -9,12 +9,12 @@
     "cypress-recurse": "^1.23.0",
     "html-validate": "^8.7.1",
     "imap-simple": "^5.1.0",
-    "jsonwebtoken": "^9.0.0",
+    "jsrsasign": "^11.0.0",
```
whabanks, line 12: This was tricky to get working. `jsonwebtoken` depends on the `crypto` node package which is server side, not browser facing. Suggestions online were to use polyfills to inject `cyrpto` into Cypress via `cypress.config.js` which turned out to not work anyway 😬.

tests_cypress/cypress/e2e/admin/register.cy.js
andrewleith, line 1: `Succeeds with a valid form` only seems to pass the first time then it stops working.  I'm not sure if this is because of archiving the account at the end of the test or not.

Quick idea (which pollutes the db) to fix this would be to generate the account email address randomly (I did this for an api key [over here](https://github.com/cds-snc/notification-api/blob/main/tests_cypress/cypress/e2e/api/sre_tools.cy.js#L8))

tests_cypress/cypress/Notify/Admin/Pages/CreateAccountPage.js
```diff
@@ -0,0 +1,66 @@
...
+
+    Submit: (name, email, mobile, password) => {
+        Components.FullName().type(name);
+        Components.EmailAddress().type(email);
+        Components.MobileNumber().type(mobile);
+        Components.Password().type(password);
+        Components.SubmitButton().click();
+    },
+
+    CreateAccount: (name, email, mobile, password) => {
```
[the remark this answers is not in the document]
whabanks: I love this idea. We should aim to keep the actions succinct and tied to actions you can take **within** that page. The wait for email is definitely something that should be part of the test itself, not the list of actions a user would take on a page.

tests_cypress/cypress/e2e/admin/register.cy.js
whabanks: +1 a random uuid after the `+` in the email is a good idea. I went with archiving the user after the test completed to avoid pollution but looking at the `users` table the entry still remains. It'd be great if we could expand the Admin API to support setup/teardown needs for these tests. ie: delete the user all together instead of just archiving.

tests_cypress/package.json
whabanks: In the [`sign`](https://github.com/auth0/node-jsonwebtoken/blob/bc28861f1fa981ed9c009e29c044a19760a0b128/verify.js#L22) class in `jsonwebtoken` it requires `KeyObject, createSecretKey, createPublicKey` from `crypto` which it is unable to resolve.  The devs claim `crypto` is unresolvable because it's not [intended to be accessed/executed from a browser environment](https://github.com/auth0/node-jsonwebtoken/issues/885#issuecomment-1423983985) but serverside where node.js is running.

It's odd to me though because we're using node 16 which includes the imports that `jsonwebtoken` requests from crypto. There seems to be some separation between what Cypress serves up for the browser gui / running tests, and the Cypress backend.

That said we can do some pretty neat stuff with polyfills by [customizing Cypress's internal webpack config](https://github.com/cypress-io/cypress/discussions/24358).

whabanks: Hmm maybe a package update did break it. Whenever I called `CreateJWT` it failed on the call to `jwt.sign()` with `KeyObject, createSecretKey, createPublicKey` all being undefined. Happy to dig a bit, it's probably best to continue using `jsonwebtoken` instead.

PR #1760, opened 2024-02-20
feat: add new cta to homepage + translations

andrewleith:
# Summary | Résumé

This PR updates the calls to action on the homepage by adding a new one to register for demos.

### Before
image

### After
image

### DIFF
--- a/app/templates/partials/page-content/cta.html
+++ b/app/templates/partials/page-content/cta.html
@@ -7,7 +7,10 @@
           <h2 class="heading-large">
             {{ _("Start sending test messages right away!") }}
           </h2>
-            {{ primary_button_blue_lighter(label=_("Try it now"), href=url_for('.register' )) }}
+            <div class="flex flex-col space-y-4 sm:flex-row sm:space-y-0 sm:space-x-4">
+              {{ primary_button_blue_lighter(label=_("Try it now"), href=url_for('.register' )) }}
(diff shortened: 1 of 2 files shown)


app/templates/partials/page-content/cta.html
```diff
@@ -7,7 +7,10 @@
           <h2 class="heading-large">
             {{ _("Start sending test messages right away!") }}
           </h2>
-            {{ primary_button_blue_lighter(label=_("Try it now"), href=url_for('.register' )) }}
+            <div class="flex flex-col space-y-4 sm:flex-row sm:space-y-0 sm:space-x-4">
+              {{ primary_button_blue_lighter(label=_("Try it now"), href=url_for('.register' )) }}
+              {{ primary_button_blue_lighter(label=_("Register for a demo"), href=gca_url_for('register_for_demo')) }}
+            </div>
           <p class="text-white">
```
amazingphilippe, line 14: We need to remove the "want to know more, contact us" since we're not funelling demo requests through our contact form.

amazingphilippe (changes requested):
One more thing to remove and then I think it should be good!

app/templates/partials/page-content/cta.html
andrewleith: Good catch, thanks.  Just pushed a chage.

amazingphilippe (approved):

PR #1763, opened 2024-02-21
Revert "Filtering heartbeats"

whabanks:
Reverts cds-snc/notification-admin#1755

The language toggle on the index page was causing major slow downs and in some cases full timeouts.


jimleroyer (approved):
LGTM

PR #1765, opened 2024-02-26
Add validation for SMS sends via CSV that exceed 612

whabanks:
# Summary | Résumé
- Added new error message to inform users of the row(s) containing variables that would exceed the 612 character sms limit
- Added inline errors on placeholder content cells when

Depends on: https://github.com/cds-snc/notification-utils/pull/278

# Testing
## Set up
- SMS template with 1 variable and some content. 
- SMS template with 2 (or more) variables and some content
- Create CSVs with some rows where the variables do not exceed 612 characters, and some rows that do.

## Test 1
1. Try to send using the CSV's that go over the limit for both templates
2. Note that an error message is displayed at the top of the `Edit your spreadsheet` screen
3. Note that inline error messages are displayed on the cells with variable content indicating the length limit. 

## Test 2
1. Change the CSV's so there are no more errors
2. Re-upload and verify that normal sending still functions as expected.


jzbahrai:
Will we have a performance problem if we upload a 50000 row csv with custom content?

whabanks:
Performance stats when processing a 5.83Mb CSV with ~13k rows. 

## With new changes:
This was run with some added optimizations to [rows_with_combined_variable_content_too_long](https://github.com/cds-snc/notification-utils/blob/main/notifications_utils/recipients.py#L280) that are not yet merged.
```
notifications_utils.recipients.RecipientCSV.rows:                4.20s
notifications_utils.recipients.RecipientCSV.rows_with_errors:    4.21s
notifications_utils.recipients.RecipientCSV.displayed_rows:      4.26s
notifications_utils.recipients.RecipientCSV.sms_fragment_count:  4.38s
notifications_utils.recipients.RecipientCSV.sms_fragment_count:  66.89s
# The new code is called in here 
Execution time of app.utils.get_errors_for_csv:    16.60s
```
Total time: `100.45s`


## Same CSV without this PR's changes:
```
Execution time of notifications_utils.recipients.RecipientCSV.rows:                 72.29s
Execution time of notifications_utils.recipients.RecipientCSV.rows_with_errors:     4.21s
Execution time of notifications_utils.recipients.RecipientCSV.displayed_rows:       4.26s
Execution time of notifications_utils.recipients.RecipientCSV.sms_fragment_count:   125.12s
Execution time of notifications_utils.recipients.RecipientCSV.sms_fragment_count:   76.37s
Execution time of app.utils.get_errors_for_csv:    1.17s
```

Total time: `282.25s`

The executions times here are wildly inconsistent. Without the PR changes, the CSV should have no errors so I though this should be faster. It seems like the main culprit, in both cases, is `sms_fragment_count`. It gets called twice during the process and is fairly expensive.

### Possible solution
Since we only display the first 50 rows with errors to the user, we could add a processing threshold for the variable content length check.
Once we hit `x` number of errors, where `x` is either 50 or some threshold that doesn't severely affect performance, we stop processing the CSV and return the current set of errors for the user to resolve.

jzbahrai (approved):

PR #1767, opened 2024-03-01
Feat/add scanfiles api

andrewleith:
# Summary | Résumé

This PR adds support for the scan-files API by exposing one method, `is_unsafe()`, which will assess the file and return true if the api responds with 'malicious' or 'suspicious' before the timeout.  In all other cases, including timing out, this method will return `True`

## Reviewer checklist | Liste de vérification du réviseur
- [x] This PR does not break existing functionality.
- [x] This PR does not violate GCNotify's privacy policies.
- [x] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [x] This PR does not significantly alter performance.
- [x] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).
> ⚠ If boxes cannot be checked off before merging the PR, they should be moved to the "Release Instructions" section with appropriate steps required to verify before release. For example, changes to celery code may require tests on staging to verify that performance has not been affected.


jzbahrai (commented):
just one q - should we put this in utils instead of admin? the main scanfiles class?

andrewleith:
> just one q - should we put this in utils instead of admin? the main scanfiles class?

I certainly thought about it.  I didn't for 2 (possibly not-so-good) reasons:
1. Its always a pain to work with utils (extra releases, lining up versions, etc.)
2. I modelled this off of how we integrated salesforce/zendesk, which was done directly in ADMIN

If we do plan to use scanfiles in API though we certainly should.  Thoughts? 

@jimleroyer do you want to weigh in?

jzbahrai:
@andrewleith I guess we don't plan on making branding an api call, this can work

app/scanfiles/scanfiles_api_client.py
```diff
@@ -0,0 +1,67 @@
...
+            bool: True if the file is considered unsafe, False otherwise.
+        """
+        data = {"file": file}
+        headers = {
+            "Authorization": self.auth_token,
+        }
+
+        try:
+            response = requests.post(self.scanfiles_url, files=data, headers=headers, timeout=15)
+        except Exception as e:
```
jzbahrai, line 41: are there specific exceptions this might raise that we can catch and model behaviour for?

andrewleith: There are, but I was just thinking along the lines of if something goes wrong - timeout, its down, anything - we just ignore it and let the person submit their request.

app/scanfiles/scanfiles_api_client.py
```diff
@@ -0,0 +1,67 @@
...
+    IN_PROGRESS = "in_progress"
+    CLEAN = "clean"
+    SUSPICIOUS = "suspicious"
+    MALICIOUS = "malicious"
+    ERROR = "error"
+    UNKNOWN = "unknown"
+    UNABLE_TO_SCAN = "unable_to_scan"
+
+
+class ScanFilesApiClient:
```
whabanks, line 19: Is this client different than the [`antivirus_client`](https://github.com/cds-snc/notification-utils/blob/fix/enforce-sms-csv-limits/notifications_utils/clients/antivirus/antivirus_client.py#L22) in utils?

andrewleith: Good find.  Yes, I do think that is something else although unsure what.  Scan-files didnt exist back in 2018 as far as I'm aware.

jzbahrai (approved):

PR #1764, opened 2024-02-22
Feat/self serve branding

andrewleith:
# Summary | Résumé

This PR adds new pages to simplify and extend the branding process for the end user.  It adds the following screens:

- View branding settings
- Choose GOC branding
- Choose from pool of org branding
- Preview branding
- Request new branding

## Feature workflow
image

## How to access branding settings
- From your **service dashboard**, click **Settings**
- In the **Email** section, click **Change** on the **Email branding** row

## Artefacts
- [Miro board](https://miro.com/app/board/uXjVNqNAmeE=/)
- [Figma screens](https://www.figma.com/file/a1TQZWGPRijQkAfa7DHWXF/Service-branding-%7C-Image-de-service?type=design&node-id=205-403&mode=design&t=BC9cIUNLAnRPeMhb-0)

## Related Issues | Cartes liées
- https://github.com/cds-snc/notification-planning/issues/1157

# Test instructions | Instructions pour tester la modification

## Feature flag
- [x] Feature flag **ON** allows access to new branding pages
- [x] Feature flag **OFF** shows old branding pages

## Other
- [x] Branding can be seen previewed from settings
- [x] GOC branding can be selected
- [x] Branding from org pool can be selected
- [x] New branding can be requested
- [x] Admins can apply new branding to existing services

# Release Instructions | Instructions pour le déploiement

None.

# Reviewer checklist | Liste de vérification du réviseur

- [x] This PR does not break existing functionality.
- [x] This PR does not violate GCNotify's privacy policies.
- [ ] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [x] This PR does not significantly alter performance.
- [x] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).

> ⚠ If boxes cannot be checked off before merging the PR, they should be moved to the "Release Instructions" section with appropriate steps required to verify before release. For example, changes to celery code may require tests on staging to verify that performance has not been affected.

PR #1774, opened 2024-03-07
Add organisation id filter when getting email brands

jzbahrai:
# Summary | Résumé

Add an organisation id filter so that you can fetch email_brands associated with a single organisation.
Added tests.

## Testing
you can test through the preview app after you this PR is merged: https://github.com/cds-snc/notification-api/pull/2137

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1129,10 +1129,10 @@ def set_free_sms_allowance(service_id):
 )
 @user_is_platform_admin
 def service_set_email_branding(service_id):
-    email_branding = email_branding_client.get_all_email_branding()
+    organisation_id = current_service.organisation_id
+    email_branding = email_branding_client.get_all_email_branding(organisation_id=organisation_id)
 
     current_branding = current_service.email_branding_id
-
     if current_branding is None:
         current_branding = (
             FieldWithLanguageOptions.FRENCH_OPTION_VALUE
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -1131,6 +1131,10 @@ def set_free_sms_allowance(service_id):
 def service_set_email_branding(service_id):
     organisation_id = current_service.organisation_id
     email_branding = email_branding_client.get_all_email_branding(organisation_id=organisation_id)
+    # As the user is a platform admin, we want the user to be able to get the no branding option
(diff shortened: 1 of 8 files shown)


app/main/views/service_settings.py
```diff
@@ -1128,10 +1128,10 @@ def set_free_sms_allowance(service_id):
 )
 @user_is_platform_admin
 def service_set_email_branding(service_id):
-    email_branding = email_branding_client.get_all_email_branding()
+    organisation_id = current_service.organisation_id
+    email_branding = email_branding_client.get_all_email_branding(organisation_id=organisation_id)
```
andrewleith, line 1132: This works, but I wonder if we should be adding this now and potentially affecting the way support works right now, before the entire feature is finalized?

andrewleith (commented):
Code looks good but not sure if we should be implementing the filter on the admin panel just yet.

andrewleith:
There seems to be a problem with the caching, when I change orgs, I dont get an updated list of logos, still the same list from the previous org.

app/notify_client/email_branding_client.py
```diff
@@ -7,8 +7,8 @@ def get_email_branding(self, branding_id):
         return self.get(url="/email-branding/{}".format(branding_id))
 
     @cache.set("email_branding")
```
andrewleith, line 9: Not totally clear on how this caching works, but if we change this to the following it seems to work properly:
```suggestion
    @cache.set("email_branding-{organisation_id}")
```

app/main/views/service_settings.py
jzbahrai: hmm fair, I can wait on merging this change in

jzbahrai:
DO NOT MERGE. We dont want to change current functionality

app/main/views/service_settings.py
andrewleith: I added a feature flag in [the first PR](https://github.com/cds-snc/notification-admin/pull/1764) named `FF_NEW_BRANDING`, so we could use that here as well to get this in sooner than later.

andrewleith: Going back through this now, I don't think we actually want to do this here at all.  This will limit support to seeing only the brands for the current org, which will end up excluding the no branding option.

Until we improve the support experience here, I think we should jus leave this one as-is.  Thoughts?

andrewleith (approved):
LGTM!

PR #1776, opened 2024-03-08
experiment: watch and build css

andrewleith:
# Summary | Résumé

This PR adds a new npm script you can use while developing to have the CSS auto-regenerate when changes are made to HTML files.

To run it:
```
npm run watch-tailwind
```

Right now the FA icons might be being loaded twice but we can look into that later, it should have no affect on the app besides bloating the file.


andrewleith:
We implemented this when we upgraded tailwind, closing...

PR #1778, opened 2024-03-14
adding pat environment variable

ben851:
# Summary | Résumé

Forgot to add the definition for the github_workflow_pat on the docker workflow. Adding it in now.

Docker build push deploy action works, and triggers a deploy on manifests.


P0NDER0SA (approved):
that is indeed an addtional variable.  ✅

PR #1781, opened 2024-03-21
Feat/add org id to create branding

andrewleith:
# Summary | Résumé

This PR adds the org list to the create and edit branding pages (platform admin-only pages) so that they can be set when creating a new brand or updated when changing an existing one.

### Branding admin
image
### Editing existing brand
image
## Creating new brand
image


whabanks (commented):
Everything looks good, the association between the new branding and org function as expected.

New branding created this way does not seem to show up in the list on the `email-branding` admin page for me and clearing the cache doesn't help. It does however show up on the change branding page.

Change branding page:
image

Admin Email branding page:
image

andrewleith:
> ...
> New branding created this way does not seem to show up in the list on the `email-branding` admin page for me and clearing the cache doesn't help. It does however show up on the change branding page.

Great catch, I am not totally sure how the cache mechanism works, but it was modified over here in a [different PR related](https://github.com/cds-snc/notification-admin/pull/1774/files) to branding. I will review this.

andrewleith:
A few issues it seems:
1. Clearing the cache does not clear the "all branding" cache (i.e. when no org_id is provided)
2. An extra cache is being created when no org_id is provided (`email_branding-None`)

When we pass no org id to `get_all_email_branding()` (i.e. we want all email branding) it creates a cache key called `email-branding-None` and also one called `email_branding` (this was the original cache key).  I'll update this to get rid of the `email_branding` one and also update the cache clearing mechanism to clear `email_branding-None`).

whabanks (approved):
Looks good. Tested locally and after adding a new branding logo I was able to see it in the branding list.

Not sure what this is about but in the review app when I try to add a new branding logo it returns the default error page.

andrewleith:
> Looks good. Tested locally and after adding a new branding logo I was able to see it in the branding list.
> 
> Not sure what this is about but in the review app when I try to add a new branding logo it returns the default error page.

Yes I think file uploads in general dont work on the PR review env - something we'll need to look into eventually.

PR #1783, opened 2024-03-22
Fix brand preview in admin branding settings page

andrewleith:
# Summary | Résumé
This PR updates the admin branding settings page to make it so the branding can be seen.  It uses the new branding preview component to stay consistent with what the user sees.

## Before
image

## After
image


jzbahrai (approved):
tested this on the preview url! works as expected

jzbahrai (approved):

PR #1784, opened 2024-03-22
fix: validation bug when requesting new logo

andrewleith:
# Summary | Résumé
When a user browses and selects a file, its preview is displayed.  If they then browse for a file, but click cancel, the browser removes in the image from the file input so it cannot be uploaded.

This PR fixes the behaviour so that when the user cancels, the preview is also reset, so the user knows they need to select their file again and cannot submit until they do so.


whabanks (approved):
Tested locally, looks good. The preview resets as expected when you browse for a subsequent image but cancel.

PR #1791, opened 2024-04-03
a11y(mobile menu): make close button text more descriptive

andrewleith:
# Summary | Résumé

- Update the mobile menus (main menu and the account menu) close button text to be more descriptive
- Fix invalid HTML inside of buttons (i.e change `p` and `div` tags to `span` tags)

## Related cards
- https://github.com/cds-snc/notification-planning/issues/1452

### DIFF
--- a/app/templates/partials/nav/account_menu_mobile.html
+++ b/app/templates/partials/nav/account_menu_mobile.html
@@ -42,9 +42,10 @@
               src="{{ asset_url('images/close.svg') }}"
               class="object-none object-center w-8 h-8 m-auto"
               alt="" />
-            <div id="account.menu.close">
+            <span id="account.menu.close">
               {{ _('close') }}
-            </div>
+            </span>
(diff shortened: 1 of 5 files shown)


andrewleith:
> LGTM! One thing I noticed is the SR will announce that the close button is part of a group whereas other buttons in the system do not. The behaviour was there before as well. Could this cause people to think there's more than one button?

Nice catch.  Any idea what is causing it to announce its part of a group? 🤔

jzbahrai (approved):

amazingphilippe:
> Any idea what is causing it to announce its part of a group? 🤔

The button is announced as a group is a side effect of putting spans and stuff within it. It shouldn't be a problem, and probably helps SR users understand the elements inside the button. The labeling of the button after this PR sounds much better than before! I wouldn't make it a blocker to merge this PR 👍

PR #1794, opened 2024-04-03
Add aria-describedby to template links

whabanks:
# Summary | Résumé
Screen readers should now read the template name, as well as the template type when selecting either the checkbox for the template, or the link to the template

1. Navigate to the templates page
2. Turn on your screen reader
3. Put the focus on the checkbox for a template and note that the template name and type is read
4. Put the focus on the link to the template and note that the template name and type is read
5. Put the focus on a folder link, note that the folder name, and number of children in the folder is read

image
image

### DIFF
--- a/app/templates/views/templates/_template_list.html
+++ b/app/templates/views/templates/_template_list.html
@@ -45,17 +45,18 @@
               {{- format_item_name(ancestor.name) -}}
             </a> <span class="message-name-separator"></span>
           {% endfor %}
+          {# TODO: Add Cypress test to ensure describedby is present and properly associated with the link #}
           {% if item.is_folder %}
-            <a href="{{ url_for('.choose_template', service_id=current_service.id, template_type=template_type, template_folder_id=item.id, view='sending' if sending_view else None) }}" class="template-list-folder">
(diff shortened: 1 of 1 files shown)


jzbahrai (approved):

app/templates/views/templates/_template_list.html
```diff
@@ -45,17 +45,18 @@
...
             </a>
           {% else %}
-            <a href="{{ url_for('.view_template', service_id=current_service.id, template_id=item.id) }}" class="template-list-template">
+            <a href="{{ url_for('.view_template', service_id=current_service.id, template_id=item.id) }}" aria-describedby="{{ item.id }}" class="template-list-template">
               <span class="live-search-relevant">{{ format_item_name(item.name) }}</span>
             </a>
           {% endif %}
         </p>
-        <p class="message-type">
+        <p id="{{ item.id }}" class="message-type">
```
amazingphilippe, line 59: We might want to make this a span, and make sure things visually don't break. Normally, you wouldn't have nested <p> tags

amazingphilippe: Oops, my bad. This is not nested

PR #1795, opened 2024-04-03
Fixing autocomplete Text box

jzbahrai:
# Summary | Résumé
https://github.com/orgs/cds-snc/projects/40/views/5?pane=issue&itemId=45222423
The autocomplete text was overlapping after we selected it. We removed the hint text, also added bilingual No results found.

Screenshot 2024-04-04 at 2 45 10 PM
Screenshot 2024-04-04 at 2 44 55 PM
Screenshot 2024-04-04 at 2 34 46 PM


amazingphilippe:
The long input box creates a different issue:
- The proximity of the control helps understand that this is a drop down. If the box is too long, it can be difficult when zoomed in very close. 
- It also makes the layout super un balanced
- It doesn't fix text overlap on small screens.


I suggest that we don't show the hint (autocomplete__hint) when the text input has a value. This may be done via javascript, or CSS.

amazingphilippe:
> This may be done via javascript, or CSS.

May be easier to change the CSS since the whole autocomplete is a plugin that we did not write ourselves... I think

andrewleith:
You'll need to run `npm run tailwind` to get it to recompile the CSS and include your changes!

amazingphilippe:
Here's what I'm thinking:
```css
.autocomplete__wrapper:has(.autocomplete__input:valid) .autocomplete__hint {
    display: none:
}
```
Here is a breakdown of the logic here:

- Select `autocomplete__wrapper` if it **has** a **valid** (not empty) `autocomplete__input`
- Select `autocomplate__hint`
- Then don't display hint. 

I'm not sure why, but we have some other autocomplete CSS in `/components/placeholder.css` We can add this there and refactor later, or maybe create a `autocomplete.css` file within the same folder to refactor now and keep things easy to find.

amazingphilippe (approved):
lgtm

app/assets/javascripts/autocomplete.js
```diff
@@ -10,6 +10,10 @@
           '<svg class="autocomplete__dropdown-arrow-down" style="top: 8px; z-index: 1;" viewBox="0 0 512 512"><path d="M256,298.3L256,298.3L256,298.3l174.2-167.2c4.3-4.2,11.4-4.1,15.8,0.2l30.6,29.9c4.4,4.3,4.5,11.3,0.2,15.5L264.1,380.9  c-2.2,2.2-5.2,3.2-8.1,3c-3,0.1-5.9-0.9-8.1-3L35.2,176.7c-4.3-4.2-4.2-11.2,0.2-15.5L66,131.3c4.4-4.3,11.5-4.4,15.8-0.2L256,298.3  z"></path></svg>',
         selectElement: component[0],
         required: true,
+        tNoResults: () =>
+          document.documentElement.getAttribute("lang") === "en"
```
amazingphilippe, line 14: Usually, we handle translatiuons in JS with something like this:
```javascript
window.polyglot.t("translation_id")
```

PR #1796, opened 2024-04-04
Refactor scheduled message page to use a table

whabanks:
# Summary | Résumé
This PR migrates the data table in scheduled messages status page to actually use a table instead of just a series of divs formatted to look like a table.

WIP


app/templates/views/jobs/job.html
```diff
@@ -15,59 +16,70 @@
...
+
+        {% endcall %}
+      </div>
       {{ ajax_block(partials, updates_url, 'notifications_header', finished=finished) }}
 
     {% else %}
       <h1 class="heading-large">
         {{ _('Delivery report') }}
       </h1>
-
```
whabanks, line 70: Non-changing formatting changes below :)

app/templates/views/jobs/job.html
```diff
@@ -15,59 +16,70 @@
...
-          {{ _('{} recipient(s)').format(job.notification_count) }}
-        </div>
-        <div class="w-full sm:w-1/4 border-solid font-bold sm:font-normal sm:border-b-1 border-gray-400-1 pt-4 pb-4">
-          {{ _('Template') }}
-        </div>
-        <div class="w-full sm:w-3/4 border-dashed border-b-1 border-gray-400-1 pt-4 pb-4">
-          <a href="{{ url_for('.view_template_version', service_id=current_service.id, template_id=template_id, version=job['template_version']) }}">{{ template_name }}</a>
-        </div>
-      </div>
+        {% set caption = _("Scheduled message") %}
```
amazingphilippe, line 25: Missing an "s"!

app/templates/views/jobs/job.html
```diff
@@ -15,59 +16,70 @@
...
+        {% set h_scheduled_by = _("Scheduled by") %}
+        {% set h_api_request_name = _("API request name") %}
+        {% set h_scheduled_for = _("Scheduled for") %}
+        {% set h_spreadsheet = _("Spreadsheet") %}
+        {% set h_sending_to = _("Sending to") %}
```
amazingphilippe, line 30: There should be 2 TH in this table. These seem to create a row of th above, but they don'T match the table structure. 

I don't know if there is a way to flip the axis with this table component. If we can't flip the axis, we should come up with labels for the 2 columns. Something as simple as "Label" and  "Value", which we already have translations for

app/templates/views/jobs/job.html
```diff
@@ -15,59 +16,70 @@
...
+        {% call mapping_table(
+          caption=caption,
+          field_headings=[h_scheduled_by, h_scheduled_for, h_spreadsheet, h_sending_to, h_template],
+          field_headings_visible=False,
+          caption_visible=False,
+          font_size='table-font-small',
+        ) %}
+
+          {% call row() %}
+            {{ text_field(h_scheduled_by) }}
```
amazingphilippe, line 42: I think you could also call `row_heading()` instead of `text_field()` to create headings scoped to the row (this comment applies to the rows below too). In this case, I think we should also remove the header row in my comment above

app/templates/components/table.html
```diff
@@ -48,8 +48,8 @@
 
 {%- endmacro %}
 
-{% macro row(id=None) -%}
-  <tr class="table-row" {% if id and id is string %}id="cds{{id}}"{% endif %}>
+{% macro row(id=None, classes=None) -%}
+  <tr class="table-row {{ classes if classes }}" {% if id and id is string %}id="cds{{id}}"{% endif %}>
```
amazingphilippe, line 52: Neat!

app/templates/components/table.html
```diff
@@ -69,12 +69,12 @@
   {% endif %}
 {%- endmacro %}
 
-{% macro field(align='left', status='', border=True) -%}
+{% macro field(align='left', status='', cell_width=None, border=True) -%}
```
amazingphilippe, line 72: Cool addition to be able to choose the with of a col! We should test on all the breakpoints

amazingphilippe (changes requested):
Some changes need to have proper data headers, but good start!

app/templates/views/jobs/job.html
whabanks: I was able to flip the axis of the table using `row_heading` instead of `text_field` in the first column. I also removed the headings that did not map correctly with the table instead of using `Label` and `Value`. 

One caveat though, the current macro will still render a `thead` and an empty `tr` on the table.
image

If this is an issue, we could either:
1. Edit the macro to not render a `thead` if we do not pass in any header items
2. Use your suggestion of setting `label` and `value` headers.

app/templates/views/jobs/job.html
whabanks: Done!

app/templates/views/jobs/job.html
whabanks: Fixed!

app/templates/views/jobs/job.html
amazingphilippe: Lets label with "Label" and "Value" as this is what we seem to do on other tables!

app/templates/components/table.html
```diff
@@ -1,16 +1,16 @@
 {% from "components/big-number.html" import big_number %}
 
-{% macro mapping_table(caption='', field_headings=[], field_headings_visible=True, caption_visible=True, equal_length=False) -%}
-  <table class="table table-font-xsmall">
+{% macro mapping_table(caption='', field_headings=[], field_headings_visible=True, caption_visible=True, equal_length=False, font_size="table-font-xsmall") -%}
+  <table class="table {{fontsize}}">
     <caption class="heading-medium table-heading{{ ' visuallyhidden' if not caption_visible}}">
       {{ caption }}
     </caption>
     <thead class="table-field-headings{% if field_headings_visible %}-visible{% endif %}">
```
amazingphilippe, line 8: I read into it a bit more, and if we'd like to have tables with headers in the first column, we need to get rid of the `<thead>`
https://www.w3.org/WAI/tutorials/tables/one-header/#table-with-header-cells-in-the-first-column-only

I'm thinking we could wrap the `<thead>` in an if statement similar to `{% if field_headings | length > 0 %}`. That way we can (intentionally) remove the `thead` when using the `row_heading` macro.

whabanks: I updated the `mapping_table` macro as suggested. We should no longer be rendering a `thead` if `field_headings` has no elements in it.

amazingphilippe (approved):
Looks good to me!

PR #1800, opened 2024-04-11
A11y/dashboard no failure info

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/1453

This adds some hidden redundant text when "No failures" is showing under either text messages or emails statistics.

1. On the dashboard, with a screen reader like voicieover on chrome.
2. On a service with a few sends and no failures
3. Read the big_number sections with the screen readers. It should say something like: `X emails sent` then `emails sent: No failures`


andrewleith (approved):
LGTM

PR #1801, opened 2024-04-11
A11y/announce placeholders

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/1456

This slightly overhauls how placeholders are rendered in the admin interface. 

- replaces `SPAN` elements with `MARK`, which is [apparently more suited](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark) for this type of thing
- Adds pseudo elements before and after the `MARK` elements to announce the placeholder.
- Refactors the placeholder CSS so its less repetititititve

1. Have at least 2 templates ready:
    2. SMS template with a conditional and custom content
    3. Email template with a conditional and custom content
2. With voiceover on chrome
3. Start reading the template previews
4. Each placeholder should announce a "begins" and "ends"
5. Conditional placeholders should announce "if" and "then"
6. Visually, there should be no difference from prod.


andrewleith (commented):
This doesn't appear to work, `<span class="placeholder">...</span>` is still in the template preview.  I think you'd need to modify utils to change the way template previews are rendered.

That said, this technique buries language strings inside of CSS, which may make them hard to find for future developers.  I wonder if there is a way to keep the language strings outside of CSS? 

Could an alternative be to output these strings into the markup using the `sr-only` class?

app/assets/stylesheets/tailwind/components/placeholder.css
```diff
@@ -57,18 +41,84 @@
...
+    height: 1px;
+    overflow: hidden;
+    position: absolute;
+    white-space: nowrap;
+    width: 1px;
+  }
+
+  /* PLACEHOLDER TYPES */
+  [lang*="en"] mark.placeholder {
+    --name: "custom content";
```
andrewleith, line 73: The phrase "custom content" confused me here.  Why not call this "variable" or "variable content" instead?

app/templates/views/templates/_template.html
```diff
@@ -29,4 +29,5 @@ <h2 class="heading-medium">{{ _('Ready to send?') }}</h2>
   {{ template|string|translate_preview_template }}
 </div>
 
+<!-- <div id="placeholder_tag" role="tooltip" aria-hidden="true">Placeholder variable</div> -->
```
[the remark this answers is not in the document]
amazingphilippe: Yes! Thanks

app/assets/stylesheets/tailwind/components/placeholder.css
```diff
@@ -57,18 +41,84 @@
...
+  [lang="fr"] mark.placeholder-conditional {
+    --name: "contenu filtré";
+  }
+  [lang="fr"] mark.placeholder-redacted {
+    --name: "contenu rédigé";
+  }
+
+  /* BEGINS or ENDS */
+  [lang*="en"] mark[class*="placeholder"]::before {
+    content: " [" var(--name) " begins] ";
```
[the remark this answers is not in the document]
amazingphilippe: I believe they ensure a short pause. Looking at this table, only Voiceover would pause. NVDA and JAWS would announce "brackets" twice. 

JAWS always announces everything. Only pauses at commas, semicolons and ellipses

NVDA would pause with regular parenthesis ()

I don't mind using semicolons, nobody will see them!

app/assets/stylesheets/tailwind/components/placeholder.css
amazingphilippe: Whenever we do update the guidance, we'll have other things to clean up within admin. I suggest we be consistent for now, and decide if we want to re-consider this term. We also have similar work planned in https://github.com/cds-snc/notification-planning/issues/1350

andrewleith: Yeah makes sense to me!

YedidaZalik: +1 "custom content" was chosen by Yael a year or so ago

andrewleith (approved):

PR #1803, opened 2024-04-15
Add current page indicators to main menu and platform admin menu

andrewleith:
# Summary | Résumé
This PR adds current page indications using the `aria-current` attribute.  It updates the main menu for all users as well as the menu on the platform admin page.

- Inspect the source of the main menu on the site and ensure `aria-current="page"` is set on the active menu item
- Inspect the source of the platform admin menu and ensure `aria-current="page"` is set on the active menu item

### DIFF
--- a/app/templates/components/nav_menu_item.html
+++ b/app/templates/components/nav_menu_item.html
@@ -14,6 +14,7 @@
           href="{{ url }}"
           class="text-smaller text-blue bg-gray py-5 font-bold no-underline self-center hover:underline line-under border-b-4 border-transparent visited:text-blue link:text-blue focus:border-b-4 focus:bg-yellow focus:border-blue focus:outline-none {{ css_classes }}"
           {% if is_external_link %}target="_blank"{% endif %}
+          {% if css_classes and "header--active" in css_classes %}aria-current="page"{% endif %}
         >{{ localised_txt }}
           {% if is_external_link %}<i aria-hidden="true" class="pl-2 fa-solid fa-arrow-up-right-from-square"></i>{% endif %}
         </a>
--- a/app/templates/views/platform-admin/_base_template.html
+++ b/app/templates/views/platform-admin/_base_template.html
@@ -26,8 +26,11 @@
             <a
               aria-label="{{ _(link_text) }}" 
               href="{{ url_for('main.'+view) }}"
-              class="no-underline text-blue visited:text-blue link:text-blue text-smaller">
+              class="no-underline text-blue visited:text-blue link:text-blue text-smaller"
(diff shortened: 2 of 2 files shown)


amazingphilippe (changes requested):
I wish we would rely on aria to apply style, rather than the opposite. It might be more repetitive, but I think we should:
- change the macros to take in a "current" parameter. 
- With CSS, apply "current" styles with `[aria-current="page"] {...}`

Also I think we should look at the `nav_menu_item_mobile` macro too: https://github.com/cds-snc/notification-admin/blob/main/app/templates/components/nav_menu_item_mobile.html

app/templates/partials/nav/gc_header_nav.html
```diff
@@ -72,51 +72,51 @@
...
+                {{ nav_menu_item(i.url, i.title, 'header--active' if i.active else '', is_external_link=i.target, is_active=i.active) }}
               {% endif %}
             {% endfor %}
           {% else %}
-            {{ nav_menu_item('/', _('Home'), 'pl-0 header--'+header_navigation.is_selected('home'), 'nav-head-home')}}
-            {{ nav_menu_item(gca_url_for('whynotify'), _('Why GC Notify'), 'header--'+header_navigation.is_selected('why-notify'), 'nav-head-why-notify') }}
-            {{ nav_menu_item(gca_url_for('features'), _('Features'), 'header--'+header_navigation.is_selected('features'), 'nav-head-features') }}
-            {{ nav_menu_item(documentation_url(), _('API documentation'), 'header--'+header_navigation.is_selected('documentation'), 'nav-head-doc', True) }}
-            {{ nav_menu_item(url_for('main.contact'),_('Contact us'),'header--'+header_navigation.is_selected('contact'), 'nav-head-contact') }}
+            {{ nav_menu_item('/', _('Home'), 'pl-0 header--'+header_navigation.is_selected('home'), 'nav-head-home', is_active=header_navigation.is_selected('home')) }}
```
amazingphilippe, line 79: Do we want to remove the `header--active` class if we don't need it anymore?

amazingphilippe (commented):
Very cool! I know you've probably had enough with this one, but do you think we should cleanup the header--active class we apply to the nav links?

whabanks (approved):
Code looks good, the `aria-current="page"` is present on desktop and mobile for the main nav and admin nav.

LGTM pending Phil's suggestion to remove the `header--active` classes.

app/templates/partials/nav/gc_header_nav.html
andrewleith: Refactored the macros a bit more to get rid of this!

amazingphilippe (approved):
LGTM 🎉

PR #1804, opened 2024-04-15
Tweak 2fa form def in forms.py. Fix tests

amazingphilippe:
# Summary | Résumé

https://github.com/cds-snc/notification-planning/issues/1457

This fixes the issue and updates the 2fa field to a more modern pattern:

- type is text
- inputmode is numeric
- autocomplete is one-time-code
- pattern stays [0-9]*

I focused my changes in forms.py. I think its cleaner than in the (sms and email) templates since all 2fa inputs should behave the same. 

Edited the tests. 

1. Sign in
2. Inspect 2fa code input. It should have all the props listed above.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -183,7 +183,7 @@ class TwoFactorCode(StringField):
     ]
 
     def __call__(self, **kwargs):
-        return super().__call__(type="tel", pattern="[0-9]*", **kwargs)
+        return super().__call__(type="text", inputmode="numeric", autocomplete="one-time-code", pattern="[0-9]*", **kwargs)
 
 
 class ForgivingIntegerField(StringField):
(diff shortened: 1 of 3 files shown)


andrewleith (approved):
This change works well, tested on my Pixel 7. 

Just a note that the `inputmode` property [isn't supported by the Safari](https://caniuse.com/?search=inputmode)

One small issue that we may want to address here is that on small screens the textbox ends up going to full-width, which could confuse users since it can be a much larger box that would be necessary to enter a 5 digit code:

image

amazingphilippe:
> Just a note that the `inputmode` property [isn't supported by Safari](https://caniuse.com/?search=inputmode)

But only on desktop! I believe this prop relates to the touch screen virtual keyboards on phones and such.

PR #1805, opened 2024-04-16
Make checklist stack on md and below

andrewleith:
# Summary | Résumé

This PR updates the go-live checklist at `/services/{service_id}/service-settings/request-to-go-live` to allow it to stack when the viewport gets to `md` or smaller.

## Before
image

## After
image

# Related cards
- https://github.com/cds-snc/notification-planning/issues/1459

- Create a trial service
- Go to the go-live page
- Shrink your window down to less than 768px
- Ensure content stacks instead of overlaps


app/templates/components/task-list.html
```diff
@@ -1,17 +1,17 @@
 {% macro task_list_item(status, label, link) %}
-  <li class="task-list-item">
+  <li class="task-list-item pb-3 md:pb-0">
```
amazingphilippe, line 2: The tailwind classes in this file could be placed in the css file to make it easier to read and reuse

amazingphilippe (changes requested):
I'd like to see all your changes in the css file, I think it will make things easier to maintain. 

Also, I wonder if you considered a flexbox approach that wraps? We could neatly align text to their baseline and not worry about absolute positioning and pixel nudging!

app/templates/components/task-list.html
andrewleith: Yeah I was wondering about that.  We don't seem to be consistent across components. Personally I prefer reducing the mental overhead and not having to switch between css and markup files, but this is a personal preference.

For this particular component, since it already has a css file, I'll go ahead and do it.  How do you do stuff like `md:pb-0` in the css file?

andrewleith:
> I'd like to see all your changes in the css file, I think it will make things easier to maintain.
> 
> Also, I wonder if you considered a flexbox approach that wraps? We could neatly align text to their baseline and not worry about absolute positioning and pixel nudging!

I tried a flexbox approach initially but it broke the custom highlight styles that are part of this component.  The CSS/markup here is fairly complex to pull that off and I didn't think it was worth the time to rebuild all of that in a different way, so I just followed with what they were doing by switching from `position: absolute` to `position: static` on small viewports.

app/templates/components/task-list.html
amazingphilippe: > How do you do stuff like `md:pb-0` in the css file?

You can add those in the `@apply` too! Just needs a recompiling.

andrewleith: Nevermind, I [did it over here](https://github.com/cds-snc/notification-admin/blob/a11y/make-go-live-checklist-stackable/app/assets/stylesheets/tailwind/components/empty-list.css#L17-L18), I'll follow this pattern.

amazingphilippe: Might be easier to just `@apply md:pb-0` here.

amazingphilippe:
> broke the custom highlight styles that are part of this component

Oh! I see. Though would this be broken too?
image

app/templates/components/task-list.html
andrewleith: I ran into this last time I tried this, it wont work:
`([object Object]:[object Object]) /workspaces/notification-admin/app/assets/stylesheets/tailwind/components/task-list.css `@apply` cannot be used with .md\:pb-0 because .md\:pb-0 is nested inside of an at-rule (@media).`

amazingphilippe: Oh no! You are right. This is disappointing... What you have is the recommended way to do this it seems. So lets do it!

andrewleith:
> > broke the custom highlight styles that are part of this component
> 

> Perhaps the highlight shouldn't extend to the status. So I wouldn't sweat it too much. Happy to go with this once the changes are all in the css :)

Hmm yeah missed that one.  Considering this is a use-case that will probably rarely (or never) really be used, I think we should limit how much time we invest in this!

andrewleith:
@amazingphilippe ok all set, the classes have been moved from the markup to the css file!

amazingphilippe (approved):
🚢 (once the tests pass! just a `make format`)

PR #1806, opened 2024-04-16
Make tables responsive/scrollable on smaller screens

andrewleith:
# Summary | Résumé
This PR updates the `mapping_table()` macro to make tables automatically scrollable on small screens.

This technique was taken from: https://adrianroselli.com/2020/11/under-engineered-responsive-tables.html

## Before
image

## After
image

- Navigate to a page with a table on it (dashboard: bulk sends, or notification report table)
- Reduce your screen size to 320 or lower
- Ensure tables scroll horizontally instead of overlap


amazingphilippe (commented):
This is awesome! You made it work!

I think we may need a few design tweaks. I've noticed that:
- On taller tables, we lose the shadow effect. I can come up with some ideas. 
image

- There is a focus style applied when focusing a table container. It doesn't really fit with our visual language, I can also come up with some ideas. 
image

andrewleith:
> * On taller tables, we lose the shadow effect. I can come up with some ideas.

Good catch!

> * There is a focus style applied when focusing a table container. It doesn't really fit with our visual language, I can also come up with some ideas.

Yeah I wasn't too sure about the focus style - grateful if you can come up with something!

whabanks (approved):
LGTM very nice work on the CSS for this! I tested out the shadow effect on a large table (900 rows) and it appears to be functioning as expected:
image

app/templates/components/table.html
```diff
@@ -1,36 +1,38 @@
...
-            </th>
-          {% endif %}
-        {% endfor %}
-      </tr>
-    </thead>
-    <tbody>
-      {{ caller() }}
-    </tbody>
-  </table>
+  <section class="scrollable-table" tabindex="0">
```
whabanks, line 4: Just as a heads up on a potential merge conflict. I also made changes to the `mapping_table` in [this PR](https://github.com/cds-snc/notification-admin/pull/1796/files#diff-6f3b6bacd0cef3451a94304ec31ddf290ee9029f88c9175008c69dfaa16fc3c6R3).

amazingphilippe:
@andrewleith I just pushed some style tweaks. Should look like this:
Capture d’écran, le 2024-04-19 à 09 31 28
Capture d’écran, le 2024-04-19 à 09 31 32
Capture d’écran, le 2024-04-19 à 09 31 36

andrewleith:
LGTM!

PR #1808, opened 2024-04-16
Autocomplete tel fields

amazingphilippe:
# Summary | Résumé

For https://github.com/cds-snc/notification-planning/issues/1458 

Adding autocomplete=tel attributes where I think we need them. Whenever a user is asked to provide their phone number. We don't want this when we're asking for a recipient's phone number. 

1. Create an account
2. Check that autocomplete=tel attribute is on the field.


app/templates/views/register-from-org-invite.html
```diff
@@ -14,10 +14,10 @@
     <h1 class="heading-large">{{ _('Create an account') }}</h1>
     <p>{{ _('Your account will be created with this email:') }}{{invited_org_user.email_address}}</p>
     {% call form_wrapper() %}
-      {{ textbox(form.name, width='w-full md:w-3/4') }}
+      {{ textbox(form.name, width='w-full md:w-3/4', autocomplete='name') }}
```
amazingphilippe, line 17: This one was also missing!

PR #1809, opened 2024-04-16
Make copy template links clearer for all and for SR

amazingphilippe:
# Summary | Résumé

For https://github.com/cds-snc/notification-planning/issues/1462

To fix this issue, I did:

1. Add details to the title: "Copy existing templates into {service name}" Hopefully this makes the destination clear for everyone.
2. Add a "Copy {template name}" prefix to make the action clear for all.
3. Add a hidden suffix "into {service name}" so that each link is extra clear for SR users. Maybe too verbose, so Id like to test if the h1 was sufficient for SR users too... We can safely merge this is and test later. 

1. Get into the copy template flow
2. Navigate through services and folders
3. It helps to have many services and folders
4. Notice the added context 
5. Inspect the hidden context
6. Test with voice over on chrome.

### DIFF
--- a/app/templates/views/templates/copy.html
+++ b/app/templates/views/templates/copy.html
@@ -4,13 +4,13 @@
 {% extends "admin_template.html" %}
 
 {% block service_page_title %}
-  {{ _('Copy an existing template') }}
+  {{ _('Copy an existing template') }} {{ _("into") }} {{ current_service.name }}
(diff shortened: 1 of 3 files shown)


whabanks (approved):
Tested on the review app and locally with voice over. Changes are as expected and code looks good!

PR #1812, opened 2024-04-22
Increase cypress login timeout

andrewleith:
# Summary | Résumé

This PR sets an explicit `10s` timeout when trying to login to notify via the cypress tests, which will hopefully avoid the timeouts we see in staging when the environment is sluggish.


whabanks (approved):
LGTM!

PR #1819, opened 2024-04-24
Button links accessibilty

amazingphilippe:
# Summary | Résumé

Sometimes, links are buttons. From the design system:

> Choose buttons for actions that modify data, change states, or initiate actions, like submitting a form, playing a video, or saving.

This is related to https://github.com/cds-snc/notification-planning/issues/565

Inconsistent behaviour and presentation of controls:
1. Links behaved like links -> We want them to behave like buttons. 
2. Links looked like buttons -> Thats the desired presentation.

This PR suggests that we add a few things to our link-that-look-like-buttons so that they behave like buttons. 

1. They have a `button` role
2. They aren't draggable or selectable
3. They show the correct cursor on hover
4. They can be activated with the space bar. 

There are quite a few pages containing link-that-look-like-buttons. 

- [x] Add new service on Choose account
- [x] Add recipients and Send a test on Template
- [x] Continue on Login events
- [x] Invite someone on Team members
- [x] Request to go live on Settings
- [x] Create API key on API keys
- [x] New branding on Email branding
- [x] Add reply-to on Reply-to
- [x] Continue on Confirm your mobile number
- [x] Re-send on Re-send security code
- [x] New and Test on Security keys


app/templates/views/api/keys.html
```diff
@@ -20,7 +20,7 @@
   </p>
 
   <div class="js-stick-at-bottom-when-scrolling">
-    <a href="{{ url_for('.create_api_key', service_id=current_service.id) }}" class="button button-secondary">{{ _('Create API key') }}</a>
+    <a href="{{ url_for('.create_api_key', service_id=current_service.id) }}" role="button" draggable="false" data-module="button" class="button button-secondary">{{ _('Create API key') }}</a>
```
whabanks, line 23: Perhaps we should add a new macro to `buttons.html` called `secondary_button_link` or similar to encapsulate the attribute additions to cut down on duplicate changes / the cognitive load of needing to remember that these attributes are now required when making a "link button".

app/assets/javascripts/button.js
```diff
@@ -0,0 +1,26 @@
...
+       * - role="button" is added to the link so that it’s recognised and announced as button by voice dictation and screen reader software.
+       * - draggable="false" is added to the link so that people with dexterity impairments don’t accidentally drag the link instead of cliking it.
+       */
+
+      // Listen to keydown on component
+      $(component).on("keydown", (event) => {
+        let target = event.target;
+
+        // If role is a button, and SPACE is pressed...
+        if (target.getAttribute("role") === "button" && event.which === 32) {
```
whabanks, line 18: I'm a little torn between `event.which` and `event.key` here. It's not particularly crucial to this PR since we're just handling the `spacebar` but I think it's an interesting discussion.

On one hand jQuery's [`event.which`](https://api.jquery.com/event.which/) will normalize keycode differences across browsers automagically and hence is the safest in terms of compatibility with older browsers. It's also slightly more cumbersome to work with, most of us probably don't have `32 = spacebar` in the back of our heads.

On the other hand [`event.key`](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) achieves that same level of normalization while also handling differences in keyboard layout / locale and handles modifier keys. It's a little simpler to look at code wise as you can use the literal character: `event.key === " "`.

andrewleith, line 18: Why check for the role here?  This module is exclusively used for this case isn't it?

andrewleith: If we do have it, this check should be outside the event handler closure.  There is no need to add the keydown handler if this condition isn't satisfied.

app/templates/views/api/keys.html
andrewleith: Actually I have another idea. Since we have to use JS to make this happen, what if we keep everything co-located in the JS? 

Instead of marking up our links-as-buttons like this (and/or adding a macro that does this):
```
<a href="#" role="button" draggable="false" data-module="button" class="button button-secondary">Go</a>
```

What if we just did this:
```
<a href="#" data-module="button">Go</a>
```

And in our JS module, we can add the attributes we need and also add the functionality? That way the enhancement from link to button is all together, and there are less files to jump around in to understand how this works.

What do you think, @whabanks @amazingphilippe?

app/templates/components/buttons.html
```diff
@@ -39,6 +39,8 @@
 <a
     {% if id_key %}id="{{ id_key }}"{% endif %}
     href="{{ href }}"
+    role="button"
```
andrewleith, line 42: I think you may have missed `data-module="button"` here?

app/templates/views/api/keys.html
andrewleith: Pushing this further: we could move away from the GOVUK module syntax of `data-module='xxx'` and just target anchor elements that have `class="button"`.  

This would target all of our existing link-as-button instances without the need for changing any markup at all in this PR.

So our link-as-buttons would look like this (exactly how they were before this PR):
```
<a class="button" href="...">{{ _('Create template') }}</a>
```

app/assets/javascripts/button.js
andrewleith: Good question btw @whabanks.  +1 to using `event.key` and not relying on jQuery.

Also, I tracked [down where GDS does this](https://github.com/alphagov/govuk-frontend/blob/main/packages/govuk-frontend/src/govuk/components/button/button.mjs#L74C4-L74C29) and they are using `event.key` and `' '`.

amazingphilippe: Yes! Lets do that, I didn't think of using an actual space char. I was trying `"SPACE"`

amazingphilippe: Checking for the role here might be a bit extra defensive. So that we don't apply this module accidentally to something else.

app/templates/components/buttons.html
amazingphilippe: Very likely!

app/templates/views/api/keys.html
andrewleith: Yeah I agree in general with doing as much before JS as possible. I think we need to balance maintainability as well, and spreading this out across JS, HTML and possibly macros might bite us later.  @whabanks, what do you think?

amazingphilippe: alright, so should I go through the code and try to replace as many `<a buttons>` with our existing macro? It would be just as many changes, but better maintainability in the future.

app/templates/components/buttons.html
```diff
@@ -59,3 +62,14 @@
...
+{% macro link_button(label, href, class, id=None, testid=None) %}
+    <a 
+        {% if id %} id="{{ id }}" {% endif %} 
+        {% if testid %} data-testid="{{ testid }}" {% endif %} 
+        href="{{ href }}" 
+        class="button{{ class and ' ' + class }}" 
+        role="button" draggable="false" data-module="button">
+            {{ label }}
+    </a>
+{% endmacro %}
```
amazingphilippe, line 75: Created a macro. To change the button style, we can pass in a class string.

app/templates/views/api/keys.html
whabanks: I think it makes sense, from a maintainability perspective, to house these enhancements in JS. My only concern is that we're introducing yet another way to template a component. Would this be a one off just for button links or something we could explore in future refactors?

That said, in this case since we already have some JS associated with this change, I think it makes sense to experiment with simplifying our markup in this way. Jinja components turn into spaghetti really quickly the more conditionals and complexity are added.

andrewleith: Thanks @whabanks. What do you think @amazingphilippe?  If we do what we're describing here, we can get rid of the macro and just change the JS a bit to handle all of the enhancement in one place.  If you like I can commit what we're describing (or show you before doing so).

amazingphilippe: Ok! Might be easier to just rip out the js out of this PR and make a new one instead of removing all the changes in this PR. Feel free to commit or start over!

andrewleith: Done!

app/assets/stylesheets/tailwind/components/buttons.css
```diff
@@ -101,5 +101,14 @@
...
     @apply bg-yellow text-blue outline-none border-white no-underline;
   }
+
+  /* Dealing properly with links that are buttons. sometimes, links are buttons... */
+  a[role="button"] {
+    user-select: none;
+  }
+  a[role="button"]:hover {
+    cursor: default;
+  }
```
amazingphilippe, line 111: I think we can safely remove those. If we really want to keep them, we can add them under `a.button {...}` But this might be confusing to read if we never set this role on anchors.

andrewleith: Can remove them for sure, but why are they no longer applicable?

whabanks (approved):
Pretty clean and all encompassing solution. Tested each page / button link listed in the test instructions against the expected behaviour, activated cursor, non-draggable, non-selectable text, etc.
LGTM! 🚀

PR #1839, opened 2024-05-09
Automatic redirect on session timeout

andrewleith:
# Summary | Résumé

This PR: 
- automatically redirects a user at session time out so they are aware it happened
- adds a banner when a user explicitly logs out via the menu items
- adds a note to the login screen to let users know what the session time out is (satisfying [WCAG SC 2.2.6](https://www.w3.org/WAI/WCAG21/Understanding/timeouts.html))

## Related cards
* https://github.com/cds-snc/notification-planning/issues/417

### Session time out
image

### Sign out
image

### Sign in
image

## Session timeout redirect
- Log in to Notify
- Wait 8 hours
   - [x] You should be automatically redirected to the sign in page with a message indicating you were logged out
## Sign out
- Log in to Notify
- Log out via the manu
   - [x] You should return to the sign in form and see a banner indicating you have logged out
## Sign-in
- Go to sign in page
   - [x] You should see a message indicating session length on the login form

### DIFF
--- a/app/translations/csv/fr.csv
+++ b/app/translations/csv/fr.csv
@@ -1869,3 +1869,4 @@
 "Enter the alternative text in English","Entrez le texte alternatif en anglais"
 "Enter the alternative text in French","Entrez le texte alternatif en français"
 "You have been signed out.","FR: You have been signed out."
+"Note: user sessions expire after 8 hours of inactivity.","FR: Note: user sessions expire after 8 hours of inactivity."
--- a/app/translations/csv/fr.csv
+++ b/app/translations/csv/fr.csv
@@ -1868,3 +1868,4 @@
 "Alternative text in French","Texte alternatif en français"
 "Enter the alternative text in English","Entrez le texte alternatif en anglais"
 "Enter the alternative text in French","Entrez le texte alternatif en français"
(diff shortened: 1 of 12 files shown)


app/translations/csv/fr.csv
```diff
@@ -1868,3 +1868,5 @@
 "Alternative text in French","Texte alternatif en français"
 "Enter the alternative text in English","Entrez le texte alternatif en anglais"
 "Enter the alternative text in French","Entrez le texte alternatif en français"
+"You have been signed out.","FR: You have been signed out."
```
jzbahrai, line 1871: Is this translation incorrect? -  it needs to be in FR

jzbahrai (commented):
Tested the changes locally (so I could have a smaller session timeout), got all 3 scenarios!

app/translations/csv/fr.csv
andrewleith: Oh yeah need to get that one translated...

andrewleith:
```suggestion
"You have been signed out.","Déconnexion réussie"
```

jzbahrai (approved):

PR #1844, opened 2024-05-15
bug: fix stats info

jzbahrai:
# Summary | Résumé

We are calculating stats for api key usage on the notifications table. If the key has been used, before 7 days, we do not indicate that.

This is the func that gets the stat: https://github.com/cds-snc/notification-api/blob/main/app/dao/fact_notification_status_dao.py#L325-L343

as you can see it only calls notifications. We do not want this query to also call notification_history. Changing the wording to in

Problem:
Screenshot 2024-05-15 at 11 08 10 AM
dicate its only for the last 7 days


andrewleith (approved):
LGTM

PR #1848, opened 2024-05-22
Feat/tou dialog

andrewleith:
# Summary | Résumé

---

---


YedidaZalik:
@andrewleith after consulting with Melissa, we'd like to change "By using GC Notify, you agree to accept our terms of use, including to:" to "By using GC Notify, you accept our terms of use. This includes your agreement to:"

app/__init__.py
```diff
@@ -216,6 +218,11 @@ def get_locale():
     application.jinja_env.globals["gca_url_for"] = gca_url_for
     application.jinja_env.globals["current_service"] = current_service
 
+    # cross-cutting concerns for TOU/login events
+    application.jinja_env.globals["show_tou_prompt"] = show_tou_prompt
+    application.jinja_env.globals["parse_ua"] = parse
+    application.jinja_env.globals["events_key"] = EVENTS_KEY
```
andrewleith, line 224: This allows the TOU prompt logic to be accessed from any route.  Not sure if there is a cleaner way to do this without having to import TOU stuff into every controller.

app/main/views/index.py
```diff
@@ -304,6 +306,19 @@ def activity_download():
...
+def agree_terms():
+    accept_terms()
+
+    next_url = request.form["next"]
+    if next_url and is_safe_redirect_url(next_url):
+        url = next_url
+    else:
+        url = url_for("main.show_accounts_or_dashboard")
+
+    return redirect(url)
```
andrewleith, line 319: This is where the user navigates when they accept the terms.  It sets a session object to indicate the terms have been accepted so they don't get prompted on subsequent requests.

app/main/views/two_factor.py
```diff
@@ -87,6 +88,8 @@ def parse_ua(ua):
             user_agent = parse(ua)
             return str(user_agent)
 
+        session[EVENTS_KEY] = user.login_events
+
```
andrewleith, line 92: This adds the `login_events` to the session so they can be displayed from any route, in case the user tries to navigate around without accepting the terms.

app/templates/admin_template.html
```diff
@@ -14,11 +16,15 @@
...
-    </div>
+      <div class="grid-row contain-floats">
+        <main role="main" class="w-full float-left {% block main_content_class %}py-10{% endblock %} px-gutterHalf box-border column-main">
+          {% if show_tou_prompt() %}
+            {{ tou_prompt() }}    
+          {% else %}
+            {% include 'partials/flash_messages.html' %}
+            {% block maincolumn_content %}{% endblock %}
+          {% endif %}
+        </main>
```
andrewleith, line 27: This checks if the TOU have been accepted on every route.  If they have, business as usual, if not, they will be prompted.

scripts/run_tests.sh
```diff
@@ -37,7 +37,7 @@ display_result $? 1 "Import order check"
 mypy ./
 display_result $? 1 "Type check"
 
-npx prettier --check app/assets/javascripts app/assets/stylesheets
+npx prettier --check app/assets/javascripts app/assets/stylesheets tests_cypress/cypress/e2e
```
andrewleith, line 40: Adding cypress tests to the prettier command so they get formatted consistently, too.

tests/__init__.py
```diff
@@ -46,12 +47,15 @@ def delete(self, key):
 
 
 class TestClient(FlaskClient):
-    def login(self, user, mocker=None, service=None):
+    def login(self, user, mocker=None, service=None, agree_to_terms=True):
```
andrewleith, line 50: Agree to the terms by default so we dont need to edit a thousand tests.

tests_cypress/cypress.config.js
```diff
@@ -14,6 +14,9 @@ module.exports = defineConfig({
           "prefer-native-element": ["error", {
             "exclude": ["button"]
           }],
+          "no-redundant-role": "off",
+          "no-dup-class": "off",
+          "require-sri": "off",
```
andrewleith, line 19: Move HtmlValidate config centrally instead of inside the `a11yScan` command.

tests_cypress/cypress/Notify/Admin/Pages/LoginPage.js
```diff
@@ -15,7 +17,7 @@ let Actions = {
         Components.TwoFactorCode().type(code);
         Components.SubmitButton().click();
     },
-    Login: (email, password) => {
+    Login: (email, password, agreeToTerms=true) => {
```
andrewleith, line 20: For backwards compatibility this is set to true, which will automatically agree to the terms on login.

tests_cypress/cypress/support/commands.js
```diff
@@ -121,8 +117,8 @@ Cypress.Commands.add('getByTestId', (selector, ...args) => {
     return cy.get(`[data-testid=${selector}]`, ...args)
 });
 
-Cypress.Commands.add('login', (username, password) => {
-    cy.session([username, password], () => {
-        LoginPage.Login(username, password);
+Cypress.Commands.add('login', (username, password, agreeToTerms = true) => {
+    cy.session([username, password, agreeToTerms], () => {
+        LoginPage.Login(username, password, agreeToTerms);
```
andrewleith, line 122: Adding the third parameter to the `cy.session` command allows us to share 2 sessions throughout the tests: those that expect the user to login and already have accepted the terms (the majority, and the default if `agreeToTerms` is not provied) and then those who done (really just the tests for this particular feature).

app/main/views/index.py
jzbahrai: I might be missing it, but what is the session obj here and where is it being set for future checking?

app/templates/admin_template.html
jzbahrai: this is excellent!

app/tou.py
```diff
@@ -0,0 +1,34 @@
...
+
+    if current_app.config["FF_TOU"] is False:
+        return False
+
+    is_gca_route = False
+
+    for route in GC_ARTICLES_ROUTES.values():
+        if request.path == route["en"] or request.path == route["fr"]:
+            is_gca_route = True
+            break
```
jzbahrai, line 24: should this be return?

app/tou.py
```diff
@@ -0,0 +1,34 @@
...
+
+    if current_user.is_authenticated and not session.get("terms_agreed") and "/contact" not in request.url and not is_gca_route:
+        return True
+
+    return False
+
+
+def accept_terms():
+    session[TERMS_KEY] = True
```
jzbahrai, line 33: ok this is the session obj but where is it being used above?

app/tou.py
andrewleith: As soon as it matches one route it just sets the var to true, then its evaluated on line 26.

app/tou.py
```diff
@@ -0,0 +1,34 @@
...
+        return False
+
+    is_gca_route = False
+
+    for route in GC_ARTICLES_ROUTES.values():
+        if request.path == route["en"] or request.path == route["fr"]:
+            is_gca_route = True
+            break
+
+    if current_user.is_authenticated and not session.get("terms_agreed") and "/contact" not in request.url and not is_gca_route:
```
andrewleith, line 26:
```suggestion
    if current_user.is_authenticated and not session.get(TERMS_KEY) and "/contact" not in request.url and not is_gca_route:
```

app/main/views/index.py
andrewleith: It actually happens inside of `accept_terms()` - not sure why i commented that here 😊  Checking it is encapsulated in `tou.py`.

jzbahrai (approved):

whabanks (approved):
Followed the test instructions with `FF_TOU` on and off. Looks good, I was unable to find a way to see the TOU with the feature flag off.

PR #1850, opened 2024-05-22
Fix/sitemap tests gca footer

andrewleith:
# Summary | Résumé
This PR updates the sitemap UI test that checks for the link in the footer on GCA pages.  The structure on GCA is slightly different that within the app so the test needed to be changed.


whabanks (approved):
LGTM!

PR #1855, opened 2024-05-29
Enhance UX and add features to the Callback pages

whabanks:
# Summary | Résumé
This PR contains UX and feature enhancements to the callbacks pages.

- Additional validation:
    1. That the URL entered is valid (is `https://` and passes the URL format check in [validators](https://yozachar.github.io/pyvalidators/stable/))
    2. That the URL is reachable and returns  `http 200` by performing a small ping to their callback service
- Instead of clearing the form and saving to delete a callback config, a dedicated delete link is now available with a confirmation prompt before deletion
- A "Test response time" button was added so users can easily gauge how long their service takes to respond
- Footer form controls are sticky, so users do not need to scroll down to the bottom of the page to test response times.

1. Create a temporary callback url: https://webhook.site/ or https://beeceptor.com/
2. Navigate to the api callbacks page

### Test 1 - Valid callback
1. Use the URL you created earlier and create an API callback
2. Note that when you save a callback configuration a message with the response time is included
3. Send some notifications and make sure callbacks are working

### Test 2 - Invalid URL
1. Use the URL you created earlier but make it `http://` instead of `https://`
2. Note the error message: `Enter a URL that starts with https://`
5. Note the hint text: `Must start with https://`

### Test 3 - Valid URL but server doesn't exist
1. Modify the callback URL so it points to an invalid endpoint
2. Note the error message: `Check your service is running and not using a proxy we cannot access`
4. Note the hint text: `Your service must be running and reachable from the internet.`

## Delete button
1. Delete callback you created
2. Note the confirmation message before deletion
3. Note that during confirmation the Submission controls are not present (Save, Test response time, delete)
4. Confirm the deletion
5. Note you are returned to the API page that links to API keys, callbacks, etc.

### Test 1 - Button visibility 
1. Navigate to the callback config page
2. Note that there is no `Test response time` button
3. Create a callback again then navigate back to the callback page
4. Note that when returning to the callback config page, the `Test response time` button is visible

### Test 2 - Response time button
1. With a callback configured - click the `Test response time` button
2. Note the message indicating the response time
5. Note that you aren't taken back to the `API integration` page but remained on the Callback config page

### Test 3 - Differing URLs
1. With a callback already configured - change the URL of the callback but do not click save
2. Test the response time
3. Note that the URL input into the form was tested

### Test 4 - Validation
1. All the validation cases from the above tests should also apply to the `Test response time` button. 

## Inbound SMS callbacks page
To be able to access the Inbound SMS Callbacks page we first need to enable the `Receive inbound SMS` setting for the service.

1. Add a row to the `inbound_numbers` table:
```
INSERT INTO inbound_numbers (id,"number",provider,service_id,active,created_at,updated_at) VALUES
	 ('2a7b8671-3b71-4ba0-abcf-df880802b0f9' ,'<any phone number>','pinpoint','<your service id',true,'2024-09-09 00:00:00.000',NULL);

```
2. In the `service_permissions` add a row to give your service the `inbound_sms` permission
```
INSERT INTO service_permissions (service_id,"permission",created_at) VALUES
	 ('<your service id>','inbound_sms','2024-09-09 15:57:18.958');

```
3. Clear all `service-.*` keys in Redis
8. Navigate to the service settings and verify that the `Receive inbound SMS` setting is turned on
9. Navigate back to `API integration` -> `Callbacks` 
10. Note you can choose between `Delivery receipts` and `Received text messages`
11. Choose `Received text messages` and repeat the validation & test response time button tests to ensure feature parity between the two pages.


tests/app/main/views/test_api_integration.py
```diff
@@ -507,11 +507,22 @@ def test_should_validate_safelist_items(
...
 )
 @pytest.mark.parametrize(
-    "url, bearer_token, expected_errors",
+    "url, bearer_token, response, expected_errors",
     [
-        ("https://example.com", "", "This cannot be empty"),
-        ("http://not_https.com", "1234567890", "Must be a valid https URL"),
-        ("https://test.com", "123456789", "Must be at least 10 characters"),
+        ("https://example.com", "", None, "This cannot be empty"),
+        ("http://not_https.com", "1234567890", None, "Enter a URL that starts with https://"),
```
whabanks, line 513: Thoughts on testing these validation messages in Cypress, and instead test that we're logging appropriately here?

app/main/validators.py
```diff
@@ -141,6 +143,48 @@ def __call__(self, form, field):
...
+class ValidCallbackUrl:
+    def __init__(self, message="Enter a URL that starts with https://"):
+        self.message = message
+
+    def __call__(self, form, field):
+        if field.data:
+            validate_callback_url(field.data, form.bearer_token.data)
+
+
+def validate_callback_url(service_callback_url, bearer_token):
```
andrewleith, line 155: As discussed in slack, this might be an issue since it now expects the callback servers to respond with a 200 when we call them with `{"health_check": "true"}`.  If we want to do this we would probably have to advertise this in the docs so users are aware.

tests/app/main/views/test_api_integration.py
andrewleith: Yeah its a good question.  We could definitely test the error messages in Cypress, although you do already have it done here.  I think if we were building up this feature from scratch we would probably have created a cypress test suite, but where we are just tacking something on, maybe this is fine?

I like the idea of testing the logging, especially if those logs' formats are tied to other parts of the system like alarms.  Testing logging is pretty easy, I had to do it [for system status here](https://github.com/cds-snc/notification-lambdas/blob/main/system_status/tests/test_determine_web_status.py#L34-L48).

andrewleith (commented):
So this looks great overall, just the question on if we can actually enforce a particular response at setup time without advising the clients of this in advance.

app/main/forms.py
```diff
@@ -1442,6 +1488,7 @@ class ServiceDeliveryStatusCallbackForm(CallbackForm):
             Length(min=10, message=_l("Must be at least 10 characters")),
         ],
     )
+    test_response_time = SubmitField()
```
whabanks, line 1491: This is the only actual change in this file. Seems like the Ruff formatter went all out on this file. Not sure how it was missed in the initial reformat...

app/main/validators.py
```diff
@@ -141,6 +144,57 @@ def __call__(self, form, field):
...
+    try:
+        response = requests.post(
+            url=service_callback_url,
+            allow_redirects=True,
+            data={"health_check": "true"},
+            headers={"Content-Type": "application/json", "Authorization": f"Bearer {bearer_token}"},
+            timeout=2,
+        )
+
+        g.callback_response_time = response.elapsed.total_seconds()
```
andrewleith, line 183: I was going to say I was concerned that using the flask global object might leak the value here to other users, but just learned that values stored here are request-specific and last for only one request.  Good to know!

andrewleith (approved):
LGTM - works as expected!  Will test some scenarios in staging once its there.

PR #1856, opened 2024-05-30
Terms of use dialog (create account)

andrewleith:
# Summary | Résumé

This PR adds a terms of use component to the register page. The users must open, read, and agree to the terms in order to create a new account.

## Before
image

## After
https://github.com/cds-snc/notification-admin/assets/823749/5d3ad82b-e637-4435-84fa-d21d42fe884e

- Users should not be able to register until they accept the TOU
- Users should be presented with the TOU dialog when clicking the 'Read and agree to the terms of use'
- Dialog should close when pressing agree/cancel/close/or ESC
- ⚠️ Dialog should not be present if FF is off, and you should still be able to register an account

### A11Y
- Is accessible and has valid HTML
- Has focus set to terms when dialog opens
- Has a trigger that is described by its status
- Has terms that are described by the heading
- Closes when user presses ESC
- Has scrollable terms that are also focusable
### Trigger
- Shows as not-complete on initial load
- Remains as not-complete if user opens the dialog and clicks [cancel, ESC, close]
- Updates to complete after clicking agree
- Displays error message/error summary if terms not agreed and user submits form
- Error message/summary is removed after agreeing to terms and submitting the form
- Can click on error summary and land on dialog trigger
### Dialog
- Opens when trigger is clicked
- Has agree button disabled when dialog opens
- Enables agree button when user scrolls to bottom of terms
- Closes when user clicks close/cancel


jzbahrai (commented):
lgtm, but will wait for one more review

app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  #tou-dialog {
+    margin: 0;
+    max-height: 100%;
+
+    @screen lg {
+      margin: auto;
+      margin-top: 5%;
+      max-height: 80%;
```
amazingphilippe, line 21: To center everything. I recommend we apply auto margins all around, and set max height to something like `calc(100vh - 6rem)`. (6rem is the value of doubleGutter, would be nice to be able to leverage the token somehow)

app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -0,0 +1,38 @@
...
+    }
+  }
+
+  #tou-heading {
+    max-width: 85%;
+    @screen md {
+      max-width: 100%;
+    }
+  }
```
amazingphilippe, line 30: I don't think you need this!

app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -0,0 +1,38 @@
...
+
+  ::backdrop {
+    background: linear-gradient(to right, #140f16, #372e38);
+    opacity: 0.85;
+    transition: 1s all;
+  }
+
+  #tou-dialog {
+    margin: 0;
+    max-height: 100%;
```
amazingphilippe, line 16: max height of `100vh` could be more robust

app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -0,0 +1,38 @@
...
+  #tou-heading {
+    max-width: 85%;
+    @screen md {
+      max-width: 100%;
+    }
+  }
+  #tou-terms {
+    @apply bg-white border border-gray-200;
+    background-color: #f7f7f7;
+    max-height: 70%;
```
amazingphilippe, line 34: I don't think you need max-height either!

app/assets/stylesheets/tailwind/style.css
```diff
@@ -42,6 +42,7 @@
 @import "./components/empty-list.css";
 @import "./components/autocomplete.css";
 @import "./components/menu.css";
+@import "./components/tou-dialog.css";
```
amazingphilippe, line 45: I think we can go for a generic dialog. Given my comments in the `tou-dialog.css` file, we could remove or abstract the parts of the dialog to make sense for any dialog

app/templates/components/task-list.html
```diff
@@ -1,19 +1,20 @@
...
     {% if status == "in-progress" %}
       <strong class="task-list-indicator-not-completed" id="{{ label|id_safe }}">
         {{ _('In progress') }}
       </strong>
     {% elif status %}
-      <strong class="task-list-indicator-completed" id="{{ label|id_safe }}">
-        <img aria-hidden="true" src="{{ asset_url('images/tick.svg') }}" class="w-6 h-6 inline-block mr-2 -mt-1" alt="">
-        {{ _('Completed') }}
-      </strong>
+      {{ task_list_item_complete(label) }}
```
amazingphilippe, line 15: Love this!

app/templates/components/terms.html
```diff
@@ -0,0 +1,197 @@
...
+
+    <h3 class="heading-medium">{{ headings[0] }}</h3>
+    <p>
+    <em>For CDS responsibilities, read GC Notify’s </em
+    ><a href="https://notification.canada.ca/service-level-agreement"
+        ><em>service level agreement</em></a
+    ><em>.</em>
+    </p>
+
+    <h3 class="heading-medium">{{ headings[1] | safe }}</h3>
```
amazingphilippe, line 24: This seems a bit difficult to read as code with the headings in an array. Can we make the terms an include that contains html only? 

How about translation?

app/templates/components/tou-dialog.html
```diff
@@ -0,0 +1,62 @@
+{% from "components/task-list.html" import task_list_item_complete, task_list_item_not_complete with context %}
+{% from "components/terms.html" import terms_of_use%}
+
+{% set error_msg = _('You must read and accept the terms to continue') %}
+
+{% macro tou_dialog() %}  
+  <dialog id="tou-dialog" class="max-w-screen-lg p-6 pt-6 md:pb-20 md:px-20 lg:rounded-xl flex flex-col w-full lg:border-4 lg:border-gray-300" data-testid="tou-dialog" data-error-prefix="{{ _('Terms of use:') }}" data-error-msg="{{ error_msg }}">
```
amazingphilippe, line 7: I'd love to clean up some of your tokens overall in this component. 

- For padding, lets try to use the token `-gutterHalf`, `-gutter` or `doubleGutter` instead of a mix of 6 and 20. This is an easy way to enforce visual consistency. 
- We can avoid rounded corners altogether. 
- We can also avoid the borders. The contrast between the white dialog and black backdrop is sufficient.

app/templates/components/tou-dialog.html
```diff
@@ -0,0 +1,62 @@
...
+    <button id="tou-close" class="absolute top-0 right-0 p-4 mr-10 mt-6" data-testid="tou-close-button">
+      <i aria-hidden="true" class="mr-2 fa-solid fa-fas fa-x"></i> 
+      <span class="sr-only">{{ _('Close') }}</span>
+    </button>
+    <h2 id="tou-heading" class="heading-medium md:heading-large mt-0 mb-4 py-4" data-testid="tou-heading">{{ _('Read and agree to the terms of use') }}</h2>
```
amazingphilippe, line 12: If we can group the heading and button into a dialog header, this could help position both items as a group. I think we usually want the top bar to stay together. We would also not need absolute positioning at all. 

Same comments about padding and margins. (gutterHalf, gutter and doubleGutter)

app/templates/components/tou-dialog.html
```diff
@@ -0,0 +1,62 @@
...
+    <div id="tou-terms" class="overflow-y-scroll overflow-x-hidden p-4 lg:p-8" tabindex="0" data-testid="tou-terms" aria-describedby="tou-heading">
```
amazingphilippe, line 13: I've tested this, and your dialog component already deals with overflow. We don't need another overflow inside!

This way we can leverage sticky positioning to show the buttons once the terms have been scrolled:

https://github.com/cds-snc/notification-admin/assets/5230720/7c5eaa0b-bf98-40ca-943f-f9085924192a

app/assets/javascripts/touDialog.js
```diff
@@ -0,0 +1,131 @@
...
+    // Close the modal when the user presses cancel or the X button
+    [close_button, cancel_button].forEach(function (button) {
+      button.addEventListener("click", function (e) {
+        closeModal();
+      });
+    });
+
+    // Enable the agree button when the user scrolls to the bottom
+    terms.addEventListener("scroll", function () {
+      if (terms.scrollHeight - terms.scrollTop <= terms.clientHeight + 5) {
```
amazingphilippe, line 81: I wonder if binding the disabled state to scroll might have unintended consequence for alternative navigation methods?

(you already know how I feel about disabled buttons too 😛)

app/templates/components/tou-dialog.html
```diff
@@ -0,0 +1,62 @@
...
+{% endmacro %}
+
+{% macro tou_dialog_trigger(button_text, is_in_error_state=false) %}
+  <div id="tou_trigger_element" class="form-group {{ 'form-group-error' if is_in_error_state }}" data-testid="tou_trigger_element">
+    {% if is_in_error_state %}
+      <span id="tou-error-message" class="error-message" data-testid="tou-error-message">
+        {{ error_msg }}
+      </span>
+    {% endif %}
+    <ul class="task-list">
```
amazingphilippe, line 52: There is a lot of margin around this UL when the error message pops up. 
Since we're not actually making a list, could we just use the task list item? As long as it isn't an LI.... Maybe we should make the styles more portable so that this task list component can be used as a list or as a form control. 

Capture d’écran, le 2024-06-06 à 14 23 07

amazingphilippe (commented):
This works, but I've outlined changes I'd like to see in future iterations. I'm not expecting you to fix everything right now! So I'll just comment instead of requesting changes.

app/templates/components/tou-dialog.html
andrewleith: Yes that would be great!  I'm never really sure what to use, so I usually just tinker until it looks good.  I don't really understand the gutter classes you've added so maybe a quick sync at some point to go  over them would be good!

app/assets/stylesheets/tailwind/components/tou-dialog.css
andrewleith: I think its still needed. On mobile, the dialog is full screen.

app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -0,0 +1,38 @@
...
+  /*! purgecss start ignore */
+
+  #tou-dialog:not([open]) {
+    display: none;
+  }
+
+  ::backdrop {
+    background: linear-gradient(to right, #140f16, #372e38);
+    opacity: 0.85;
+    transition: 1s all;
```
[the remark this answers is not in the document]
andrewleith: Yeah this is a left over from when i was playing around with some animations.

andrewleith, line 11:
```suggestion
```

app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -0,0 +1,38 @@
+@layer components {
+  /*! purgecss start ignore */
+
+  #tou-dialog:not([open]) {
+    display: none;
+  }
+
+  ::backdrop {
+    background: linear-gradient(to right, #140f16, #372e38);
```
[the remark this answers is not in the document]
andrewleith: Yeah the gray will work fine!  You don't think we should have any opacity?  The opacity is there to provide users context, i.e., they are in a modal and haven't navigated to a new page.

app/templates/components/tou-dialog.html
andrewleith: Interesting! I didn't think the terms div would scroll without that.

andrewleith: Oh.. without that the inner div does NOT scroll.  Not sure what you are trying to do, but removing the overflow alone doesn't have nice end-result.

app/templates/components/tou-dialog.html
andrewleith: I was trying to re-use the component without re-working too much of the system.  Might be easier to just override the CSS in this case?
```
<ul class="task-list mt-6">
```

app/assets/stylesheets/tailwind/components/tou-dialog.css
andrewleith: Yeah not totally sure why I did it like this :/

app/assets/javascripts/touDialog.js
andrewleith: Hopefully the fable tests will shed some light on that!

andrewleith: Another option would be to enable the agree button and give them an error if its clicked and they haven't scrolled down.

app/assets/stylesheets/tailwind/components/tou-dialog.css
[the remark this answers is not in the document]
amazingphilippe: Yes! In this context opacity helps a lot. I'd just replace the gradient with one of our gray color tokens.

app/assets/stylesheets/tailwind/components/tou-dialog.css
amazingphilippe: The dialog: yes. This seems to target the dialog-heading (as the H2 element), which already takes all the width available because its a block element.

app/templates/components/tou-dialog.html
amazingphilippe: It needs a bit more than removing the overflow scroll. My concern is that we have nested overflow scrolls, which I'm not sure is a great idea. Not sure its a terrible idea either....

app/assets/javascripts/fontawesome.js
```diff
@@ -15,6 +15,7 @@ import { faCircleQuestion } from "@fortawesome/free-solid-svg-icons/faCircleQues
 import { faTriangleExclamation } from "@fortawesome/free-solid-svg-icons/faTriangleExclamation";
 import { faCircleExclamation } from "@fortawesome/free-solid-svg-icons/faCircleExclamation";
 import { faInfoCircle } from "@fortawesome/free-solid-svg-icons";
+import { faX } from "@fortawesome/free-solid-svg-icons/faX";
```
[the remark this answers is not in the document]
andrewleith: Nice!

app/assets/stylesheets/tailwind/style.css
andrewleith: Yes agreed.  There's a bit of work to get the dialog component extracted, so lets do this when we attempt that!

app/templates/components/terms.html
andrewleith: Oh yeah I didn't actually finish this.  What I'm trying to solve here is a way for the terms page and the various places we list only the headings to stay in sync.

app/templates/components/tou-dialog.html
andrewleith: Nested as in the body scrolls and so does a div?  

I think hiding the buttons might be worse than the disabled agree button though.  That way needs even more ui intervention to explain to the user why there are no buttons and how to interact with the UI instead of it being intuitive (not sure the disabled button is any better).  

I mentioned this in another comment but repeating it here: a better option than both of these might be to enable the agree button and give them an error if its clicked and they haven't scrolled down.

app/assets/stylesheets/tailwind/components/tou-dialog.css
andrewleith: Oh, I see what you mean. Yeah, this is likely not needed.

andrewleith: Ah ok after playing around a bit, this css is to make the heading and the close button not overlap on mobile when space gets tight.  I think you had a better idea of how to lay this out without absolute positioning, but I'll leave this here until we do that!

app/assets/stylesheets/tailwind/components/tou-dialog.css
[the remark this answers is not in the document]
andrewleith: Fixed!

app/assets/stylesheets/tailwind/components/tou-dialog.css
andrewleith: Fixed!

app/assets/stylesheets/tailwind/components/tou-dialog.css
andrewleith: Fixed!

andrewleith:
@amazingphilippe ok I think I've made all the changes we discussed.  If you could take a final look next week and then we'll get this merged into staging for the next steps!

andrewleith:
Hey @amazingphilippe, still waiting on your sign-off to get this merged! 😄

PR #1860, opened 2024-06-06
Revert "First crack at SRE app token"

ben851:
Reverts cds-snc/notification-admin#1859

Using app token, and it's too late to continue troubleshooting. Reverting for now


sastels (approved):
revert!

PR #1866, opened 2024-06-12
Workflow map try 2

ben851:
# Summary | Résumé

This will automatically generate a workflow map every time the workflows change and is merged to main

After merging, manually run the workflow to see if it works


jimleroyer (approved):
Curious about this, let's see

PR #1874, opened 2024-06-20
Fix/update terms headings

andrewleith:
# Summary | Résumé

This PR fixes the heading structure on the `/terms` page.  It changes them from `h3` to `h2` so that there is no jump in headings.  It also moves the a11y test for this page from the GCA suite to the APP suite, since this page is now an app page.

- Cypress tests should pass


jimleroyer (approved):
LG! :shipit:

PR #1876, opened 2024-06-24
A11y and design tweaks to Dialog component for Terms of Use

amazingphilippe:
# Summary | Résumé

Improvements for Terms of Use dialog after Fable testing and design review:

- Have an instruction to scroll down to accept
- Remove some borders so that the layout doesn't end-up looking like a text box. Especially when zoomed at 400%. 
- Adjusting the header so that all controls align to the left. Easier to find with screen magnification
- Removed header, since the footer instructions and first TOU heading made it look redundant (In the future, there will still be opportunities to have a dialog heading)
- Defined some more focus styles that were missing

1. On the account creation screen
2. Open the terms of use dialog
3. No more heading?
4. Close button moved to left
5. Close button focusable in yellow
6. Dialog focus ring
7. Dialog body focus ring
8. The terms are scrollable
9. The pre-footer is sticky
10. Scrolling to the end reveals the butttons
11. Can be scrolled with arrows and tabs
12. Cancel closes the dialog
13. Accept closes the dialog and updates the task-list component


app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -19,20 +27,5 @@
...
-    @screen md {
-      max-width: 100%;
-    }
-  }
-  #tou-terms {
-    @apply bg-white border border-gray-200;
-    background-color: #f7f7f7;
-
-    h3:first-of-type {
-      @apply mt-0;
```
amazingphilippe, line 34: Moved this to the actual terms

app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -1,6 +1,14 @@
 @layer components {
   /*! purgecss start ignore */
 
+  dialog {
+    @apply text-smaller;
+  }
+
+  .dialog-sticky-footer {
+    @apply py-gutterHalf bg-white sticky bottom-0 border-0 border-t-1 border-gray-300;
+  }
```
amazingphilippe, line 10: Just a sticky footer recipe to keep the html clean

app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -1,6 +1,14 @@
 @layer components {
   /*! purgecss start ignore */
 
+  dialog {
+    @apply text-smaller;
```
amazingphilippe, line 5: Since the dialog is out of the main div, we need to apply the base font-size. It could be cleaner to move up base font-size to a place where both dialog and main can apply it.

app/templates/components/tou-dialog.html
```diff
@@ -3,60 +3,58 @@
...
-{% macro tou_dialog() %}  
-  <dialog id="tou-dialog" class="max-w-screen-lg p-gutter flex flex-col w-full" data-testid="tou-dialog" data-error-prefix="{{ _('Terms of use:') }}" data-error-msg="{{ error_msg }}">
-    <button id="tou-close" class="absolute top-0 right-0 p-4 mr-10 mt-6" data-testid="tou-close-button">
-      <i aria-hidden="true" class="mr-2 fa-solid fa-fas fa-xmark"></i> 
-      <span class="sr-only">{{ _('Close') }}</span>
+{% macro tou_dialog() %}
+<dialog id="tou-dialog"
+  class="focus:outline-yellow overflow-y-scroll max-w-screen-lg p-0 flex flex-col w-full border-1 border-gray-300"
+  data-testid="tou-dialog" data-error-prefix="{{ _('Terms of use:') }}" data-error-msg="{{ error_msg }}">
+  <div class="dialog-header bg-gray-200 flex items-baseline gap-gutter p-gutter py-2">
```
amazingphilippe, line 10: `dialog-header` classes could be extracted later on

andrewleith (commented):
Code-wise this looks good, and I like the simplified design.  A couple of minor issues:
1. The div scrolls the focused close button out of view when the dialog opens - was that intentional?  I don't think there is an issue with it per se, just wanted to point out it was happening.
2. I find the instruction to read and scroll blends in quite a bit and may be hard for users to find, which could lead to some of the same issues we saw with the other design where users weren't entirely sure how to move forward (although they all did figure it out in the end, and probably would here too).  Could an icon or some color help as an aide here?

Idea:
image

amazingphilippe:
1. It was intentional and I had the same debate internally! I went for scrollable to make sure we leave enough scrollable area. The modal is also closable from the bottom with the cancel button. 
2. Also had the same thought. Maybe an icon could help, or a bit more colour...

Both need some amount of design iteration. Should we merge this now or later? Just so we have most of it out there?

andrewleith:
> 1. It was intentional and I had the same debate internally! I went for scrollable to make sure we leave enough scrollable area. The modal is also closable from the bottom with the cancel button.
> 2. Also had the same thought. Maybe an icon could help, or a bit more colour...
> 
> Both need some amount of design iteration. Should we merge this now or later? Just so we have most of it out there?

Yeah not show stoppers certainly! Approving!

amazingphilippe:
@andrewleith I went with an idea for the sticky footer. Should we keep it or revert before merging?


https://github.com/cds-snc/notification-admin/assets/5230720/0a400cdf-09db-46b4-a81c-97efeaa3952b

andrewleith:
> @andrewleith I went with an idea for the sticky footer. Should we keep it or revert before merging?
> 
>  Enregistrement.d.ecran.le.2024-06-25.a.11.57.53.mov

Oh yeah that looks sharp, I think we should go with this!

One tiny issue if you want to fix it is that when you first open the dialog, it focus on the content well with a yellow border, and when you scroll to the bottom the border touches that informational box. We could add a little margin top to push them apart (if you think it matters!).  Here's what im talking about:

image

amazingphilippe:
Yep! I noticed it too, I have something to move the oultline inset, I can also add some padding to the content

PR #1878, opened 2024-06-25
Remove background color from login terms

andrewleith:
# Summary | Résumé
This PR removes the background color from the login terms.  The same id was used for the tou dialog which caused a styling conflict.

# Before:
image

# After
image

Login to notify, ensure the terms dont have a background color


amazingphilippe (approved):
:shipit:

PR #1885, opened 2024-07-03
Add missing translations to TOU account form

amazingphilippe:
# Summary | Résumé

We were missing a few terms in our translation file for:

Capture d’écran, le 2024-07-03 à 08 54 11
Capture d’écran, le 2024-07-03 à 08 54 16
Capture d’écran, le 2024-07-03 à 08 55 17

We were also referencing prod directly, I changed some links to be relative, or use the `url_for()` function where it was easy to do so. 

1. Go to /register
2. Toggle to French
3. Don't fill out form, click continue
4. Error message should not contain English text
5. Open the modal
6. The button to close the modal should be labelled "Fermer"
7. The sticky info-box should be in French


YedidaZalik:
@amazingphilippe for step 6, the button to close the modal is labeled "Annuler" not "Fermer"
Steps 1-5 and step 7 all pass

amazingphilippe:
"Fermer" is at the top
"Annuler" is the option opposed to "Accepter" but also closes the modal without accepting.

andrewleith (approved):
Thanks for catching and fixing these, Phil!

Tested all three ✅

PR #1893, opened 2024-07-12
Removing workflow map

ben851:
# Summary | Résumé

Removing the workflow map, which was a good idea that I couldn't get to work the way I wanted.

N/A.  Workflow change.


andrewleith (approved):
Farewell

PR #1896, opened 2024-07-17
send freshdesk api

jzbahrai:
# Summary | Résumé

If the template creation/ edit form has data in the Other field - We will send that data to freshdesk

Unfortunately we don't have a freshdesk integration for staging. The code is try/ except blocks for this exact reason.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -31,6 +31,7 @@ from app import (
     template_category_api_client,
     template_folder_api_client,
     template_statistics_client,
+    user_api_client,
 )
 from app.extensions import redis_client
 from app.main import main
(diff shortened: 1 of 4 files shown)


app/main/views/templates.py
```diff
@@ -921,6 +937,25 @@ def edit_service_template(service_id, template_id):  # noqa: C901 TODO: remove t
...
+                    current_service.id,
+                    form.template_category_other.data if is_english else None,
+                    form.template_category_other.data if not is_english else None,
+                    new_template.id,
+                )
+            except HTTPError as e:
+                current_app.logger.error(
+                    f"Failed to send new template category request to Freshdesk: {e} for template {new_template.id}, data is {form.template_category_other.data}"
+                )
+
```
andrewleith, line 958: I think this code should be down near  `service_api_client.update_service_template()` on line 984.  As-is I think there is a chance this is going to send the request when a user is previewing their template.

app/main/views/templates.py
```diff
@@ -980,6 +1015,22 @@ def edit_service_template(service_id, template_id):  # noqa: C901 TODO: remove t
...
+                current_service.id,
+                form.template_category_other.data if is_english else None,
+                form.template_category_other.data if not is_english else None,
+                new_template.id,
+            )
+        except HTTPError as e:
+            current_app.logger.error(
+                f"Failed to send new template category request to Freshdesk: {e} for template {new_template.id}, data is {form.template_category_other.data}"
+            )
+
```
andrewleith, line 1033: Should this be here, again?

andrewleith (changes requested):
A question about the placement of the ticket creation in `edit_service_template`, and I think it would be good to update the tests to ensure `send_new_template_category_request()` is called as appropriate during template creation and template editing.

andrewleith (approved):
LGTM! Tested locally and behaves as expected!

PR #1897, opened 2024-07-18
fix: ensure FF is on before trying to get `template_category_id` value

andrewleith:
# Summary | Résumé

This PR adds a check when editing/creating templates so that if `FF_TEMPLATE_CATEGORY` is off, it doesnt try to read a value that won't be there.

- [ ] Should be able to create a new template with `FF_TEMPLATE_CATEGORY` off
   - [ ] Should be able to send a message w/ the template
- [ ] Should be able to edit/save an existing template with `FF_TEMPLATE_CATEGORY` off 
   - [ ] Should be able to send a message w/ the template


jzbahrai (approved):

whabanks (approved):
Tested locally with the FF off. LGTM!

PR #1898, opened 2024-07-22
Experiment/create ephemeral UI test accts

andrewleith:
# Summary | Résumé

This is an experiment to use ephemeral test account aliases in order to increase concurrency potential for UI tests.  It still uses a single gmail inbox, but now can create an unlimited amount of aliases on demand so that if tests are running in parallel they won't try to login to the same account in the same environment.

---

### DIFF
--- a/tests_cypress/cypress.config.js
+++ b/tests_cypress/cypress.config.js
@@ -1,7 +1,8 @@
 var config = require('./config');
 
 const { defineConfig } = require("cypress");
-const EmailAccount = require("./cypress/plugins/email-account")
+const EmailAccount = require("./cypress/plugins/email-account");
+const CreateAccount = require("./cypress/plugins/create-account");
 const htmlvalidate = require("cypress-html-validate/plugin");
 
 module.exports = defineConfig({
@@ -28,8 +29,8 @@ module.exports = defineConfig({
           return null
         },
         // Email Account ///
-        getLastEmail() {
-          return emailAccount.getLastEmail()
+        getLastEmail(emailAddress) {
+          return emailAccount.getLastEmail(emailAddress)
         },
         deleteAllEmails() {
           return emailAccount.deleteAllEmails()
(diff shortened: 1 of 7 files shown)


tests_cypress/cypress.config.js
```diff
@@ -28,18 +29,27 @@ module.exports = defineConfig({
...
         },
         deleteAllEmails() {
           return emailAccount.deleteAllEmails()
         },
         fetchEmail(acct) {
           return emailAccount.fetchEmail(acct)
         },
-        createEmailAccount() {
-          return emailAccount.createEmailAccount();
+        createAccount({ baseUrl, username, secret }) {
```
andrewleith, line 41: The result of this call is "cached" in the `global` object here, which means only one set of test users is generated for every test run or every time you launch the cypress UI.  To get a new user, just quit cypress and launch it again.

tests_cypress/cypress/e2e/admin/a11y/app_pages.cy.js
```diff
@@ -82,7 +82,7 @@ const pages = [
 describe(`A11Y - App pages [${config.CONFIG_NAME}]`, () => {
   for (const page of pages) {
     it(`${page.name}`, () => {
-      cy.login(Cypress.env("NOTIFY_USER"), Cypress.env("NOTIFY_PASSWORD"));
+      cy.login();
```
andrewleith, line 85: The login command has been simplified to use the generated user and the password automatically.

tests_cypress/cypress/plugins/email-account.js
```diff
@@ -75,49 +75,57 @@ const emailAccount = async () => {
...
-                        subject: mail.subject,
-                        text: mail.text,
-                        html: mail.html,
+                    let latestMail = null;
+                    let uidsToDelete = [];
+                    for (const message of messages) {
+                        const mail = await simpleParser(message.parts[0].body);
+                        const to_address = mail.to.value[0].address;
+
+                        if (to_address == emailAddress) {
```
andrewleith, line 101: Only get 2FA codes for the specific email alias we are testing with 🎉

smcmurtry (approved):
Looks good! I tried this locally and was able to run the new `make` commands. Thanks for adding the documentation!

Some of the cypress tests are randomly failing for me but this is not specific to your PR - it happens on the `main` branch as well.

PR #1902, opened 2024-07-23
updating newrelic agent to 9.12.0

P0NDER0SA:
# Summary | Résumé

updating newrelic agent to 9.12.0

[Investigate Why New Relic 8.11 Causes Pods To Take Too Long to Start](https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/313)

Major version upgrade to see if there are any issues lingering in our Admin Pod after upgrade.  This could involve slow startups and response times up to 10 seconds during the first hour of the pod's life.

- > Upon being merged to main -- watch the pod and track its liveliness health and time it takes to get healthy
- > Monitor response times for Admin to ensure they haven't increased


pyproject.toml
```diff
@@ -62,7 +62,7 @@ redis = "3.5.3" # pinned for now
 xlrd = "1.2.0" # this is pinned so that we can continue to support .xslm files; see https://github.com/pyexcel/pyexcel-xls/issues/53
 # Putting upgrade on hold due to v1.0.0 using sha512 instead of sha1 by default
 itsdangerous = "2.2.0"  # pyup: <1.0.0
-newrelic = "8.10.0" # Pinned to 8.10.0 for now, 8.11.0 caused a performance regression: https://gcdigital.slack.com/archives/C012W5K734Y/p1709668046344929
+newrelic = "~9.12.0"
```
jimleroyer, line 65: It might be safer to pin this to an exact version and let renovate upgrade it for us. I have the impression the `~` would upgrade the minor version and potentially an automatic upgrade.

ben851: yeah probably a good idea

P0NDER0SA: Yep, that's fair.  The same thought did cross my mind, and I'll probably do the same for API.

PR #1903, opened 2024-07-23
Run unit tests using multiple configurations

andrewleith:
# Summary | Résumé
## Background
Recently we've had some issues with bugs popping up due to feature flags.  While development is ongoing in staging with a particular feature flag on, it is easy to miss ensuring everything will still work in PROD while the feature flag remains off.  The idea of this PR is to run our unit tests twice on each PR when feature flag settings differ between staging and prod

## Changes
This PR adds support to run 2 instances of our unit tests, in parallel.  

By default, our unit tests run using `NOTIFY_ENVIRONMENT=test` which is provided by `pytest.ini`.  The test workflow [has been duplicated](https://github.com/cds-snc/notification-admin/pull/1903/files#diff-6cb9ad7a3f7ae7dc858a932eacc676d72edb32a2d163d8a66e99b6b0cd267cba) and modified to run the unit tests [using an alternative value](https://github.com/cds-snc/notification-admin/pull/1903/files#diff-ce6f95ab671f9e81034bede3d0c91092dbb853ddf87b59460ae4876b0d34738eR40) for `NOTIFY_ENVIRONMENT`. 

A [new config section](https://github.com/cds-snc/notification-admin/pull/1903/files#diff-84f1a9419471e289c6b6e2b0209b329e20df6cef81d1f7f0a193ddc2fc6ad69dR230) for mimicking prod's feature flag configuration has been created, and the existing [Test](https://github.com/cds-snc/notification-admin/pull/1903/files#diff-84f1a9419471e289c6b6e2b0209b329e20df6cef81d1f7f0a193ddc2fc6ad69dR196) config has been updated to have the same feature flags set as staging does.


.github/workflows/test_prod_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+      with:
+        path: ~/.npm
+        key: ${{ runner.os }}-node-${{ hashFiles('package-lock.json') }}
+        restore-keys: |
+          ${{ runner.os }}-node-
+    
+    - name: Check for missing translations
+      run: make test-translations
+    - name: Update NOTIFY_ENVIRONMENT for pytest
+      run: sed -i 's/NOTIFY_ENVIRONMENT=test/NOTIFY_ENVIRONMENT=production_FF/' pytest.ini
```
andrewleith, line 40: This is the only difference between the original workflow and the new one. Not sure if we can make these re-usable somehow and not have to copy them.

.github/workflows/test_prod_config.yaml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Echo contents of config.py
+      run: cat pytest.ini
```
andrewleith, line 42: Remove debugging
```suggestion
```

whabanks (approved):

jzbahrai (approved):

PR #1910, opened 2024-07-26
Don't require description if the category is hidden

andrewleith:
# Summary | Résumé

---

---


whabanks (changes requested):
Tested this locally following the test steps. Couple of things I noticed

✅  When setting the category to `hidden` both the list page and edit page show `Other` 
🛑  When I try to save an edit to the template (non-category related edit) the validation fires on the category field. The `Other` option is not selected. 
🛑  Selecting `Other` and saving does not preserve the original Category and instead assigns the template to the generic `low category`

app/main/views/templates.py
```diff
@@ -315,14 +315,16 @@ def choose_template(service_id, template_type="all", template_folder_id=None):
...
 
-    # Get the full list of template categories, excluding hidden ones
-    template_categories = list(
-        {
-            template.template_category[template_category_name_col]
-            for template in template_list
-            if template.template_category and not template.template_category.get("hidden")
-        }
-    )
+    # Get the full list of template categories, any hidden ones will be called 'Other'
```
amazingphilippe, line 318: I wonder if it would be simpler to allow the hidden category as it is labelled, as long as it is pre-selected. For example:
- User chooses category A on template A
- Admin makes category A invisible
- User still see Category A when browsing and editing template A
- User cannot choose Category A on new templates, or templates not categorized with Category A
- User can choose Category B on Template A, save template
- Template A can no longer be of Category A

amazingphilippe: So looping back to your comment line, Something like: 
`# Get the full list of template categories, any hidden ones matching the current category can be added to the list.`

amazingphilippe: You can ignore the above, I see you have it working now. I am able to 
- ✅ Edit and Save a template, after its category was set to hidden
- ✅ Open up the category radio list, see the checked "Other", and save the template

andrewleith: I'm not sure.  Part of the problem here is we exposed a piece of functionality in the platform admin UI meant to be in the background for our default categories (low/med/high) and now we have all these edge cases trying to control this.

I don't think this scenario is very likely to happen, and I think we've already spent quite a bit of time on this, so personally I don't want to try to re-work it another way unless there is a clear benefit!

andrewleith:
> Tested this locally following the test steps. Couple of things I noticed
> 
> ✅ When setting the category to `hidden` both the list page and edit page show `Other` 🛑 When I try to save an edit to the template (non-category related edit) the validation fires on the category field. The `Other` option is not selected. 🛑 Selecting `Other` and saving does not preserve the original Category and instead assigns the template to the generic `low category`

I think I have this fixed up now, if you want to test again @whabanks.

whabanks (approved):
LGTM!

✅ `Other` label present when assigned to a hidden Category
✅ Updating a template still works when assigned to a hidden category - defaults to other
✅ With `Other` selected the original (hidden) category is preserved in the DB

PR #1923, opened 2024-08-06
bootstrapping the xray sdk/middleware

P0NDER0SA:
# Summary | Résumé

Just like the title says :) 
add the middleware and sdk to admin for aws Xray.

Includes a fix for 500s when it should be a 404 on xxxxxxx/services/wrongendpoint

https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/319

- [x] This PR does not break existing functionality.
- [x] This PR does not violate GCNotify's privacy policies.
- [x] This PR does not raise new security concerns. Refer to our GC Notify Risk Register document on our Google drive.
- [ ] This PR does not significantly alter performance.
- [x] Additional required documentation resulting of these changes is covered (such as the README, setup instructions, a related ADR or the technical documentation).

> ⚠ If boxes cannot be checked off before merging the PR, they should be moved to the "Release Instructions" section with appropriate steps required to verify before release. For example, changes to celery code may require tests on staging to verify that performance has not been affected.


app/__init__.py
```diff
@@ -699,6 +700,9 @@ def useful_headers_after_request(response):
 
 def register_errorhandlers(application):  # noqa (C901 too complex)
     def _error_response(error_code):
+
+        xray_recorder.begin_segment("error_response")
```
jimleroyer, line 704: Would we need to call `xray_recorder.end_segment` somewhere? The [basic usage doc](https://docs.aws.amazon.com/xray-sdk-for-python/latest/reference/basic.html) mentions it might have to be called before `begin_segment` sometimes?

jimleroyer: I wonder too if x-ray would detect the error segment automatically with the error response that is generated. Are we testing the segment feature of x-ray here or trying to fix an issue where x-ray fails to detect errors?

P0NDER0SA: This was added as a failsafe for certain errors where the code path results in an empty/null segment.  This code will verify that all errors are working as expected and that x-ray is still reporting traces.

P0NDER0SA: Stay tuned for code updates

jimleroyer (approved):
LGTM - Let's also monitor if that slows down our application boot up significantly (i.e. like NewRelic latest updates).

PR #1925, opened 2024-08-07
Add int test for create template TC FF

whabanks:
# Summary | Résumé
This PR adds the front end changes necessary to correct a bug where we were unable to override a template's priority to the same priority that it's associated Category uses. It also adds an end-to-end integration test suite in Cypress to test all the scenarios listed in the [bug card](https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1631). The tests treat API as a black box, performing actions in the front end and checking if the results are as expected in the DB.

# Related Tasks
* https://github.com/cds-snc/notification-api/pull/2243

These tests assume that you are logged in as admin and `FF_TEMPLATE_CATEGORY` is `True`.

- [x] Cypress tests are passing
- [ ] Template priority can be overridden to the *same type* that the category is using
    * Create a new template 
    * Select a category 
    * Set the template to use the category's priority (Information Blast = bulk)
    * Edit the template and override the priority to bulk
    * Note that the `process_type` column in the DB is set to bulk
- [ ] Template priority can be overridden to the other priorities (normal, priority)
- [ ] Template with an overridden priority can be set to use the template
     * Edit the template you created above
     * Set it to use the template category
     * Note in the DB the `process_type` is null, and the `template_category_id` corresponds to `Information Blast`


andrewleith (approved):
LGTM! 🚢

PR #1931, opened 2024-08-20
Hide sensitive services on admin

jzbahrai:
# Summary | Résumé

Change using self.sensitive_services to using the data from the DB
We want to be able to set the sensitive_services from the platform admin settings under the Service settings page. 
Please use this API backend PR: https://github.com/cds-snc/notification-api/pull/2258

1. Go to settings under any given service.
2. There is a new setting called "Set Sensitive Service" - it also exists in french
3. Toggle the setting successfully
4. Look at the code where we used to use config['SENSITIVE_SERVICE'] - it has now been replaced.


app/main/views/service_settings.py
```diff
@@ -388,6 +388,23 @@ def service_switch_live(service_id):
     )
 
 
+@main.route("/services/<service_id>/service-settings/set-sensitive-service", methods=["GET", "POST"])
+@user_is_platform_admin
+def set_sensitive_service(service_id):
+    title = _("Set sensitive service")
+    form = ServiceOnOffSettingForm(name=title)
```
andrewleith, line 395:
```suggestion
    form = ServiceOnOffSettingForm(name=title, enabled=current_service.sensitive_service)
```

andrewleith (changes requested):
Looks good, works as expected.  One little suggestion to have it show the current value of the setting when you go to the page.

I tested that the the string "Sensitive" would be logged as appropriate and it seems to work.  I'll re test this in staging in the cloudwatch logs.

andrewleith (approved):

PR #1937, opened 2024-08-29
Add admin interfact to suspend callback if they exist

jzbahrai:
# Summary | Résumé

We want to suspend an API callback if they exist.

Use this PR for the API: https://github.com/cds-snc/notification-api/pull/2275
1. Go to a service
2. Add a callback
3. Go to service setting page
4. If you are admin you should see a toggle at the bottom "Suspend Callback"
5. Based on the toggle you should see the correct data in your local DB

Test 2:
1. Go to a service which doesn't have a callback setup
2. Go to the service setting page - there should be no way for you to Suspend a Callback for that service.


whabanks (approved):
LGTM! Services with callbacks configured can be suspended and services without callbacks configured do not have the suspension option in the service settings.

PR #1957, opened 2024-09-25
Squash a few routing / UI bugs

whabanks:
# Summary | Résumé

This PR fixes a few bugs found during testing on staging for the callbacks changes:

- The delete button is not longer displayed when there isn't a callback configured
- Fixed an issue where the delete confirmation banner wasn't displaying in some circumstances
- Fixed a routing issue where deleting a received-text-messages callback was deleting the delivery status callback
- Fixed an issue where the Test callback button was not remaining on the same page after clicking

### DIFF
--- a/app/main/views/api_keys.py
+++ b/app/main/views/api_keys.py
@@ -28,13 +28,11 @@ dummy_bearer_token = "bearer_token_set"
 @main.route("/services/<service_id>/api")
 @user_has_permissions("manage_api_keys")
 def api_integration(service_id):
-    callbacks_link = ".api_callbacks" if current_service.has_permission(
-        "inbound_sms") else ".delivery_status_callback"
+    callbacks_link = ".api_callbacks" if current_service.has_permission("inbound_sms") else ".delivery_status_callback"
     return render_template(
         "views/api/index.html",
         callbacks_link=callbacks_link,
-        api_notifications=notification_api_client.get_api_notifications_for_service(
-            service_id),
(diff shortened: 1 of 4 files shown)


app/main/views/api_keys.py
```diff
@@ -250,26 +250,34 @@ def delivery_status_callback(service_id):
...
-                        _l("The service {} responded in {} seconds.").format(
-                            url_hostname,
-                            response_time,
-                        ),
-                        "default_with_tick",
-                    )
-                    return redirect(url_for("main.delivery_status_callback", service_id=service_id))
-
+            # If the user is just testing their URL, don't send them back to the API Integration page
+            if request.form.get("button_pressed") == "test_response_time" and g.callback_response_time >= 1:
```
jzbahrai, line 254: in this case is the callback saved or not? the flash message doesn't indicate that, whereas if it is <1s, it does say the callback is saved

whabanks: Hmm good catch on this case. Right now, if the url or bearer token have changed it will call the update method, then check if the test button was clicked and display the response time.

I think what we want is to check if the test button was clicked first, and put the url/bearer token check + update call in the else. It would be confusing to perform a save and tell them that after they click the test button.

jzbahrai: ok we dont need to change this for now, I guess it is saved? but maybe we need stronger language? your callback should respond <1s for us to count it performant or something like that

jzbahrai (approved):

PR #1970, opened 2024-10-15
dep updates

ben851:
# Summary | Résumé

Updated babel, ruff, and types with latest minor versions.

Smoke test when complete

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -28,8 +28,8 @@
         "textarea-caret": "3.1.0"
       },
       "devDependencies": {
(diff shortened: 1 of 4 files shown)


sastels (changes requested):
You have to update the package-lock file too, [I think](https://stackoverflow.com/questions/57867267/how-to-update-package-lock-json-without-doing-npm-install) with something like
```
npm i --package-lock-only
```

sastels (approved):
LGTM!

PR #1975, opened 2024-10-18
Tweak chown to skip .git folder

amazingphilippe:
# Summary | Résumé

I've been having issue building my container for a while. It was failing on post Create command. I narrowed it down to an issue with this line of code, where the command was not able to chown my local git logs and objects. 

After some experimenting, I was able to figure out that we can filter this command to not run on the .git folder. My container was able to build after doing so. 

So I'm suggesting this as a fix for everyone!

Rebuild your container.


.devcontainer/scripts/notify-dev-entrypoint.sh
```diff
@@ -10,5 +10,7 @@ set -ex
 # Tell git the workspace repository is safe, else upcoming commands will fail.
 git config --global --add safe.directory /workspaces/notification-admin
 
-chown -R vscode:vscode /workspaces/notification-admin
+find /workspaces/notification-admin \
+    -name .git -prune -o \
+    -exec chown vscode:vscode {} +
```
jimleroyer, line 15: Sorry you wasted some time on that one. I think we can remove the `chown` command! It is not necessary anymore afaik.

jimleroyer: Otherwise, there might be this exclusion pattern that could be possible using `chown` command itself:
```sh
chown -R vscode:vscode /workspaces/notification-admin/!(.git)
```
Coming from this [stackoverflow solution](https://superuser.com/a/1742875).

amazingphilippe: removed!

amazingphilippe:
We can close this given the recently merged docker changes

PR #1976, opened 2024-10-22
Add service setting to change annual limits for platform admins

whabanks:
# Summary | Résumé

This PR adds capability for platform admins to set annual email and sms limits for a service. It also changes the SMS and Email annual limit displays so they are no longer hard coded but fetched from the DB.

All `service-{id}` keys need to be cleared when deployed to staging / prod

1. If it is not yet merged, check out [this API PR](https://github.com/cds-snc/notification-api/pull/2329/files#diff-02a4832ab85a57344c4179cf331c533fa2bf4367833127a54e479dcf986a66d3) and run `flask db upgrade`
2. Login as a platform admin
3. Navigate to a service's settings
4. In the platform admin section, modify both the email and SMS annual limit
5. Note that the limits are reflected in the email and text message sections above


app/config.py
```diff
@@ -239,6 +261,13 @@ class ProductionFF(Config):
     NO_BRANDING_ID = "0af93cf1-2c49-485f-878f-f3e662e651ef"
 
     FF_RTL = False
+    FF_SPIKE_SMS_DAILY_LIMIT = True
+    FF_SMS_PARTS_UI = False
+    FF_BOUNCE_RATE_V1 = True
+    FF_BOUNCE_RATE_V15 = True
+    FF_BOUNCE_RATE_BACKEND = True
+    FF_ABTEST_SERVICE_ID = ""
```
andrewleith, line 269: These should be removed as they are no longer in use.

app/config.py
```diff
@@ -239,6 +261,13 @@ class ProductionFF(Config):
...
+    FF_ANNUAL_LIMIT = True
```
andrewleith, line 270: This section is to mimic what values we will use in prod.  I assume we will release this with `FF_ANNUAL_LIMIT` off, so this should be set to `False` for now.

andrewleith (approved):
LGTM in general, works as expected with the API PR. 

Just the nagging issue of the formatting but that doesnt impact this functionally.

app/config.py
whabanks: Fixed!

app/config.py
whabanks: Fixed!

PR #1984, opened 2024-10-31
Limits for go-live form step 2

amazingphilippe:
# Summary | Résumé

- Changes the go-live form to gather more structured data. As outlined inthis [shorthand form doc](https://docs.google.com/document/d/1acZV4WHwj1w2zbb5PAhUPG3YhRkHPiuzaSClZXM_dyA/edit?tab=t.0).

1. Log into Notify Staging
2. Create a new trial service if needed
3. In the request to go live checklist, choose the "Tell us about how you intend to use GC Notify" step
4. Fill in page 1 of the form. 
5. The new form elemetns are on page 2 of the form.
6. You should be able to pick a range for the volumes of sms and emails you want to send daily/monthly. 
7. For the daily limit, if you want to send a lot more, there is a conditional field that opens when you select the "More than X" option.


YedidaZalik (changes requested):
@amazingphilippe now that I can see how this appears in the interface, I agree that "How many" works!
I also realized that we could shorten the headings from 
"About emails you want to send" to "Email" and "About text messages you want to send" to "Text messages" 
I think that could reduce cognitive load. Is it too late to do that?

amazingphilippe:
> "About emails you want to send" to "Email" and "About text messages you want to send" to "Text messages"

I agree! Not too late. Good catch!

app/main/forms.py
```diff
@@ -1800,46 +1801,198 @@ class GoLiveAboutServiceFormNoOrg(StripWhitespaceForm):
...
+        self.min = min
+        self.max = max
+        self.message = message
+
+    def __call__(self, form, field):
+        trigger_data = getattr(form, self.trigger_field).data
+
+        # If trigger radio isn't selected, Stop Validation
+        if trigger_data != self.trigger_value:
+            print("DEBUG: Trigger condition not met, stopping validation chain")
```
andrewleith, line 1817: Debug statements should be removed

app/main/forms.py
```diff
@@ -1800,46 +1801,198 @@ class GoLiveAboutServiceFormNoOrg(StripWhitespaceForm):
...
+            # First check if empty
+            if field.data is None or field.data == '':
+                raise ValidationError(self.message or _l("This cannot be empty"))
+                # Then check range if value is provided
+            if self.min is not None and field.data < self.min:
+                raise ValidationError(_l("Number must be more than {min}").format(min=format_thousands_localized(self.min)))
+            if self.max is not None and field.data > self.max:
+                raise ValidationError(_l("Number must be less than {max}").forma(max=format_thousands_localized(self.max)))
+        else:
+            print("DEBUG: Trigger condition not met, skipping validation")
```
andrewleith, line 1832: Debug statements should be removed

app/main/forms.py
```diff
@@ -1800,46 +1801,198 @@ class GoLiveAboutServiceFormNoOrg(StripWhitespaceForm):
...
+    annual_email_volume = RadioField(
+        _l("How many emails do you expect to  send in a year?"),
+        validators=[DataRequired()],
+    )
+    daily_sms_volume = RadioField(
+        _l("How many text messages do you expect to send on a busy day?"),
+        validators=[DataRequired()],
+    )
+    annual_sms_volume = RadioField(
+        _l("How many text messages do you expect to  send in a year?"),
```
andrewleith, line 1880: There are some extra spaces in some of these strings.

app/main/forms.py
andrewleith, line 1: Since this form is quite complex, it would be good to have some more tests around it.  Specifically, that the `OptionalIntegerRange` is working how you expect it to (for example, that validations only fire when the last radio is selected and no value is given, or the value is out of range, etc.)  

There only seems to be extra validation (and extra "other/specify" fields) on two of the four "how many" fields.  Is that intentional?

app/main/forms.py
```diff
@@ -1800,46 +1801,198 @@ class GoLiveAboutServiceFormNoOrg(StripWhitespaceForm):
...
+        ]
+
+    def more_validators(self, limit, notification_type):
+        return [
+            OptionalIntegerRange(
+                trigger_field=f"daily_{notification_type}_volume",
+                trigger_value=f"more_{notification_type}",
+                min=limit * 10,
+            )
+        ]
```
andrewleith, line 1946: This validation doesn't trigger when you input the the min boundary.  I.e. where it says "More than 100 000`, you can enter exactly 100,000 and there is no validation error.

andrewleith (changes requested):
Looks great overall.  A few comments about adding some tests due to the overall complexity of this particular form.

app/main/forms.py
amazingphilippe: Absolutely, I'll work on some tests. 

And yes! The form only has 2 conditional fields total.

app/main/forms.py
amazingphilippe: I've added +1 to the "min" to fix this. Which makes sense with how the attribute should work

app/main/forms.py
```diff
@@ -1778,80 +1779,221 @@ class GoLiveAboutServiceForm(StripWhitespaceForm):
...
+            if self.min is not None and field.data < self.min:
+                raise ValidationError(_l("Number must be more than {min}").format(min=format_thousands_localized(self.min)))
+            if self.max is not None and field.data > self.max:
+                raise ValidationError(_l("Number must be less than {max}").format(max=format_thousands_localized(self.max)))
+            else:
+                # If nothing wrong found, stop validation
+                raise StopValidation()
+
+
+class BaseGoLiveAboutNotificationsForm:
```
amazingphilippe, line 1899: Made a "Base" GoLiveAboutNotificationsForm

andrewleith (approved):
LGTM

PR #1988, opened 2024-11-04
Fix content on the email address settings page

amazingphilippe:
# Summary | Résumé

After changing the creat a service flow, we must've forgottent to check the contents of the settings page where one can change a service sending email address. 

this PR updates the content of that page, so that it matches other changes we did, and most importantly fixes a missing translation. 

1. In any service
2. In the settings
3. Click on the sending email address settings
4. You should be on "Change your sending email address"
5. Toggle to FR and make sure there is no English content
6. Test the live preview functionality
7. Save the form and check that the value was saved properly.


app/config.py
amazingphilippe, line 1: Auto formatting...

app/templates/views/service-settings/email_from.html
```diff
@@ -17,15 +18,24 @@
 
   <div class="form-group">
     <p>{{ _('Users will see your sending email address in the \'From\' field of the emails you send.') }}</p>
-    <p>{{ _('Your sending email address helps build trust with recipients. A good email address is relevant and as concise as possible.') }}</p>
-    <p>{{ _('You may use letters, numbers, dots, dashes or underscores.') }}</p>
+    <p>{{ _('Your sending email address helps build trust with recipients. A good email address is relevant and as concise
+      as possible.') }}</p>
```
andrewleith, line 22: There is a line break here breaking the translation.

andrewleith, line 22:
```suggestion
    <p>{{ _('Your sending email address helps build trust with recipients. A good email address is relevant and as concise as possible.') }}</p>
```

andrewleith (changes requested):
Thanks for cleaning this up.  Looks like there is just a small issue with a line break that is breaking one translation, see below.

app/templates/views/service-settings/email_from.html
amazingphilippe: Ah yes I see that now... For some reason format-on-save really wanted to break this line... I'll just accept your change here!

andrewleith (approved):
LGTM!

PR #1999, opened 2024-11-20
Upgrade to latest New Relic version

jimleroyer:
# Summary | Résumé

Upgrade to latest New Relic version and bringing additional changes that were performed on the API side. 

1. Upgrade to New Relic v10.3.0.
2. Pass in environment to New Relic init so that it can get proper environment configuration defined in the `newrelic.ini` file. 
3. Added a total running timer for the Gunicorn process.
4. Increased gunicorm hard timeout to 90 seconds. 
5. Disable New Relic package reporting feature as we already have renovate to detect vulnerabilities and this is not worth the feature which adds a lot more time to startup (20+ seconds). 
6. Added `fzf` command line utility to the devcontainer along with proper shell config.

Check the kubernetes admin pods startup to make sure that these do not get into a crash loop. It is possible we might have to align the admin k8s liveness, readiness and startup probes with the ones in API with their latest changes.

### DIFF
--- a/gunicorn_config.py
+++ b/gunicorn_config.py
@@ -26,6 +26,28 @@ on_aws = environment in ["production", "staging", "scratch", "dev"]
 if on_aws:
     keepalive = 75
 
+    # The default graceful timeout period for Kubernetes is 30 seconds, so
+    # make sure that the timeouts defined here are less than the configured
+    # Kubernetes timeout. This ensures that the gunicorn worker will exit
+    # before the Kubernetes pod is terminated. This is important because
+    # Kubernetes will send a SIGKILL to the pod if it does not terminate
+    # within the grace period. If the worker is still processing requests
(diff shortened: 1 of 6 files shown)


gunicorn_config.py
```diff
@@ -20,10 +22,34 @@
 # to be larger than the idle timeout configured for the load balancer.
 # > By default, Elastic Load Balancing sets the idle timeout value for your load balancer to 60 seconds.
 # https://docs.aws.amazon.com/elasticloadbalancing/latest/application/application-load-balancers.html#connection-idle-timeout
-on_aws = os.environ.get("NOTIFY_ENVIRONMENT", "") in ["production", "staging", "scratch", "dev"]
```
jimleroyer, line 23: I don't know why we fell back on an empty string instead of the default `None` value. It seems unnecessary considering the following condition with the AWS env check.

PR #2010, opened 2024-12-04
Feat/validate annual limit 2

andrewleith:
# Summary | Résumé

---

---


app/main/views/templates.py
```diff
@@ -125,6 +126,31 @@ def get_char_limit_error_msg():
     return _("Too many characters")
 
 
+def get_limit_stats(notification_type):
```
jzbahrai, line 129: This function feels like a replica for notification_counts_client.get_limit_stats which you call underneath on l#131. Can we just rename it to format_limit_stats? Just so in the future the caller knows this is being set elsewhere

app/notify_client/notification_counts_client.py
```diff
@@ -0,0 +1,149 @@
...
+                    "remaining": service.sms_daily_limit - sent_today["sms"],
+                },
+            },
+        }
+
+        return limit_stats
+
+
+# TODO: consolidate this function and other functions that transform the results of template_statistics_client calls
+def _aggregate_notifications_stats(template_statistics):
```
jzbahrai, line 120: what is template_statistics? a dict being set from where?

app/notify_client/notification_counts_client.py
```diff
@@ -0,0 +1,149 @@
...
+        notifications[stat["template_type"]] += stat["count"]
+
+    return notifications
+
+
+def _filter_out_cancelled_stats(template_statistics):
+    return [s for s in template_statistics if s["status"] != "cancelled"]
+
+
+def _aggregate_stats_from_service_api(stats):
```
jzbahrai, line 133: same q as above

app/main/views/templates.py
andrewleith: Yeah this is just transforming the data for easy use in this case.  I can rename.

app/notify_client/notification_counts_client.py
andrewleith: template_statistics is the return structure from the API.  Ideally these methods would be closer to that API (i.e. in the `template_statistics_client` - but it seemed like a big lift for here.

app/notify_client/notification_counts_client.py
andrewleith: Same sort of thing.  These methods were taken from various controllers and act on the return value of the API calls.  It is quite messy.

jzbahrai: any chance you can add that to the documentation for this class?

jzbahrai (approved):

PR #2011, opened 2024-12-04
Change/annual limit

jzbahrai:
# Summary | Résumé

I changed the email_annual_limits and sms_annual_limits in the DB do that it is set to the default of 20million and 100k.
The service settings page is not behind a feature flag anymore - it will always read the result from the DB

1. You can test this on the review app (the feature flag is on ) - make sure your cache is clear - you will see the above amounts on the service_settings page
2. You can test it locally with feature flags switched off. you should see all  4 rows have the above data (the two in the sms/ email section and the 2 under platform admin)


whabanks (approved):
LGTM! Tested locally and in the review app.

PR #2014, opened 2024-12-05
Remove FF for updating annual limits

jzbahrai:
# Summary | Résumé

There is a FF on the function that calls the API to update the annual limits. I removed that.


andrewleith (approved):
LGTM

PR #2017, opened 2024-12-10
fix(dashboard): update check to see if redis has today's data

andrewleith:
# Summary | Résumé

Fix for checking if the annual_limits redis client has data for today.


whabanks (approved):
LGTM!

PR #2018, opened 2024-12-12
Feat/provide list of sending domains

andrewleith:
# Summary | Résumé

This PR updates the sending domain admin page to use a list of verified sending domains from AWS instead of a drop down list which could allow users to enter invalid domains into.

Since this lists rarely changes the request to SES is cached for 5 minutes so we don't send too many requests.  

## Related cards
- https://github.com/cds-snc/notification-planning/issues/1363

## Before - textbox
image

## After - dropdown
image

- Login as an administrator
- Navigate to a service
- Go to **Settings**
- Go to the **Sending domain** row and click **Change**
   - [ ] Ensure the list of domains are in a dropdown
- Select a different domain
- Save your setting
- Send yourself a test email
   - [ ] Ensure the sending domain is the one you selected

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -33,15 +33,12 @@
 				"vsliveshare.vsliveshare"
 			],
 			"settings": {
+				"ruff.lint.ignore": ["F401"],
+				"ruff.lint.run": "onSave",
+				"ruff.organizeImports": false,
 				"[python]": {
-					"ruff.lint.run": "onSave",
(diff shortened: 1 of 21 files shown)


sastels:
Is it easy to clear this from the cache, ie when we add a new one to AWS? (waiting 5 minutes is fine if this isn't already baked in somehow)

jzbahrai (approved):
I tested this locally and it worked, it doesn't work on admin preview though.

PR #2025, opened 2025-01-02
Add error message when notification size > 256KB

whabanks:
# Summary | Résumé
This PR adds a front end error message, with guidance, when attempting to send a one off notification that exceeds the SQS size limit of 256KB. 

image

# Related work
* https://github.com/cds-snc/notification-api/pull/2085

Pull and run the above API PR when testing locally. From the UI there are only a couple scenarios where a notification can reach the 256KB limit by using variables in one-offs. This is because:
-  The email content and subject fields have character limit validation. 
- Files can only be attached when using the API. 
- Our `RecipientCSV` class has built in validation on column values and overall content length that prevent users from reaching the 256KB limit when performing bulk sends.

### Test 1
1. Create an email template using [template_content.txt](https://github.com/user-attachments/files/18292997/template_content.txt)
2. Send using that template with this content for the variable [var1_content.txt](https://github.com/user-attachments/files/18293005/var1_content.txt)
3. Note the error message that guides you to the largest variable in your message

### Test 2
1. Edit your template, adding two more variables
2. Chunk the above `var1_content.txt` into 3 and insert them into the newly added variables, keeping track of which variable holds the largest chunk
3. Attempt the send again and note that the largest variable is specified in the error message

### DIFF
--- a/app/translations/csv/fr.csv
+++ b/app/translations/csv/fr.csv
@@ -2032,3 +2032,5 @@
 "Sending paused until 7pm ET. You can schedule more messages to send later.","FR: Sending paused until 7pm ET. You can schedule more messages to send later."
 "Sending paused until annual limit resets","FR: Sending paused until annual limit resets"
 "These messages exceed the annual limit","FR: These messages exceed the annual limit"
+"Your message exceeds the size limit of 256KB","FR: Your messages exceeds the size limit of 256KB"
+"{} holds the most content. You may be able to send your message if you reduce or remove that content.","FR: {} holds the most content. You may be able to send your message if you reduce or remove that content."
--- a/app/main/views/send.py
+++ b/app/main/views/send.py
@@ -1103,6 +1103,7 @@ def _check_notification(service_id, template_id, exception=None):
 
 
 def get_template_error_dict(exception):
+    exception_fields = {}
     # TODO: Make API return some computer-friendly identifier as well as the end user error messages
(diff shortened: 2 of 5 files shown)


app/translations/csv/fr.csv
```diff
@@ -2031,4 +2031,6 @@
 "email messages","courriels"
 "Sending paused until 7pm ET. You can schedule more messages to send later.","FR: Sending paused until 7pm ET. You can schedule more messages to send later."
 "Sending paused until annual limit resets","FR: Sending paused until annual limit resets"
-"These messages exceed the annual limit","FR: These messages exceed the annual limit"
\ No newline at end of file
+"These messages exceed the annual limit","FR: These messages exceed the annual limit"
+"Your message exceeds the size limit of 256KB","FR: Your messages exceeds the size limit of 256KB"
+"{} holds the most content. You may be able to send your message if you reduce or remove that content.","FR: {} holds the most content. You may be able to send your message if you reduce or remove that content."
```
amazingphilippe, line 2036:
```suggestion
"Your message exceeds the size limit of 256KB","Votre message dépasse la limite de taille de 256 Ko"
"{} holds the most content. You may be able to send your message if you reduce or remove that content.","{} contient le plus de contenu. Vous pourriez être en mesure d’envoyer votre message si vous réduisez ou supprimez ce contenu."
```

PR #2027, opened 2025-01-02
Add a hint only for admins when navigating outside of their teams.

amazingphilippe:
# Summary | Résumé

admins can't send messages from client services. 

Sometimes we can wonder why the send buttons are not displaying when investigating issues. This PR adds a bit of a hint to let us know that we can't send messages from some services. (When we are not part of their team)

1. With an admin account, navigate to a template in a service where you are not part of the team. You should see the hint
2. With an admin account, navigate to a template in one of your services/teams. You should NOT see the hint
3. With a regular account, navigate to a template in any of your services. You should NOT see the hint.


smcmurtry (approved):
I followed the testing steps on the review app and it looks like it is working correctly. Nice!

PR #2028, opened 2025-01-02
Add Validation Summary component to edit template forms. 

amazingphilippe:
# Summary | Résumé

The add-template page has gotten a bit long recently with the introduction of template categories. Suggesting we incorporate the validation summary component to highlight missing fields or errors. 

1. Create a new template (sms or email)
2. Submit form without filling in the form
3. Receive a validation summary

After: We can see all the errors even if not in view. 
Capture d’écran, le 2025-01-03 à 08 41 00
Capture d’écran, le 2025-01-03 à 08 42 12


app/templates/views/edit-sms-template.html
```diff
@@ -18,20 +18,22 @@
...
-            link_url=gca_url_for('personalisation_guide'),
-            link_text=_("Guide: Send messages with custom content"),
-            icon="arrow-up-right-from-square"
-          )
-        }}
+          description=_("You can put double brackets around a variable to insert custom&nbsp;content."),
+          link_url=gca_url_for('personalisation_guide'),
+          link_text=_("Guide: Send messages with custom content"),
+          icon="arrow-up-right-from-square"
+        ) }}
```
smcmurtry, line 34: This looks like a code formatting change?

app/templates/views/edit-email-template.html
```diff
@@ -51,14 +54,20 @@
...
       <h2 class="heading-medium">{{ _('Template category') }}</h2>
       {% call template_category(form.template_category_id, true if template_category_mode == 'expand' else false) %}
-        {{ select(form.template_category_id, hint=_('Template categories help improve delivery of your messages'), option_hints=template_category_hints, option_conditionals=other_category, testid="template-categories", use_aria_labelledby=false) }}
+      {{ select(
+        form.template_category_id,
+        hint=_('Template categories help improve delivery of your messages'),
+        option_hints=template_category_hints, option_conditionals=other_category,
+        testid="template-categories",
+        use_aria_labelledby=false
+      ) }}
```
smcmurtry, line 68: I think it's best to avoid including code formatting changes that are unrelated to the actual change you are trying to make, since it makes the PR bigger than it needs to be. But maybe your editor is just doing this on its own?

smcmurtry (approved):
I tested this in the review app and it seems to be working. Nicely done!

You included a few code formatting changes - you can leave them in this PR but in the future you could try to avoid that.

app/templates/views/edit-email-template.html
amazingphilippe: Yes I had to undo an earlier change as the AC for this story was changed a bit. And I couldn't quite get that section back to what it was exactly. And I'm also having some auto-formatting issues as well on html files... Maybe we should see if we can fix that at the container level by adding the correct extensions.

PR #2029, opened 2025-01-02
Fixed the send to get data for notification_count_for_the_year

jzbahrai:
# Summary | Résumé

When we get the monthly notification count, we need the count for this current fiscal year. 
The code was previously sending "2025" as the year to get the monthly data information, instead 
in the time period april 2024 - march 2025, we need to call the monthly_stats function with "2024".

# Testing:
I tested this locally, and I am sending the correct data.

# Release plan
Release this to production, and then manually test the change to see if templates can be fetched.

### DIFF
--- a/app/notify_client/notification_counts_client.py
+++ b/app/notify_client/notification_counts_client.py
@@ -1,5 +1,3 @@
-from datetime import datetime
-
 from notifications_utils.clients.redis import (
     email_daily_count_cache_key,
     sms_daily_count_cache_key,
@@ -7,6 +5,7 @@ from notifications_utils.clients.redis import (
 
 from app import redis_client, service_api_client, template_statistics_client
 from app.models.service import Service
+from app.utils import get_current_financial_year
 
 
 class NotificationCounts:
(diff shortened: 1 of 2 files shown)


smcmurtry (approved):
I was able to reproduce the bug locally by doing the following on the `main` branches of admin and api:
- create a new template
- upload a csv via the admin interface and do a bulk send
- visit the template page that was used for the send
- observe the 500 error page: `Sorry, we’re experiencing technical difficulties Try again later.` and on the api:
```  File "/workspace/app/service/statistics.py", line 112, in add_monthly_notification_status_stats
    data[month][row.notification_type][row.notification_status] += row.count
KeyError: '2025-01'
```

Then I check out Jumana's branch and refresh the page, and I no longer see an error.

This change has the side effect of making the annual limits apply to the financial year and not the calendar year, which is what we want. 

Looks good!

app/notify_client/notification_counts_client.py
```diff
@@ -83,8 +82,10 @@ def get_limit_stats(self, service: Service):
                 }
         """
 
+        current_financial_year = get_current_financial_year()
```
andrewleith, line 85: Yep my bad, I was passing the current year here where it should have been fiscal.  Thanks and sorry for the alarms 😞

andrewleith (approved):
LGTM

PR #2031, opened 2025-01-06
Get placeholders and values translated and highlighted.

amazingphilippe:
# Summary | Résumé

Needs to be merged first: https://github.com/cds-snc/notification-utils/pull/345/files

This PR uses the new `user_language` parameter on each of `SMSPreviewTemplate` and `EmailPreviewTemplate` clasees in utils to display the recipient placeholder in FR or EN depending on current language. 

Doing so, we also needed to tweak the placeholder translation in both language when rendering the template. 

1. In admin, take an SMS or Email template
2. Toggle the FR and EN, the recipient placeholder should be wrapped with (()) and display in the current language.
3. Prepare a one-off send
4. After providing a value for recipient placeholder, before sending. Toggle FR and EN and make sure the placeholder has been translated to the value.
5. Send the message


smcmurtry (changes requested):
Tested and it works great! Can you update to the new version here as well? https://github.com/cds-snc/notification-admin/blob/main/.github/workflows/test_endpoints.yaml#L61

smcmurtry (approved):

PR #2032, opened 2025-01-06
Changed data to json for correct encoding

jzbahrai:
# Summary | Résumé

We want to send the test using the clients real bearer token.
Also encoding the request with data vs json could cause an issue. I tested the change locally.


smcmurtry (approved):
Looks good to me!

PR #2047, opened 2025-01-13
Followed through some tailwind migration guides items.

amazingphilippe:
# Summary | Résumé

Noticed some bugs with outlines and backgound images after migrating to tailwind 3. This made me revisit their migration guides (1 to 2 and 2 to 3)

I've found some important changes that we needed to fix:

- purgecss isn't included above tw2 --- I've used their safelist config technique, but ideally we should stop using dynamic classes. (Eg. `class="bg-{color}"`, where we feed in a "color" somewhere else)
- `no-wrap` --- becomes `nowrap`
- outlines and focus rings are kind of different. --- I've resorted to tweaking our main focus css so that our focus ring stays more consistent with what we had previously. Its a tad thicker, but at least I removed the outline offset.


andrewleith (approved):
LGTM - tested locally.

PR #2048, opened 2025-01-13
Discovery/load gcds

amazingphilippe:
# Summary | Résumé

Configured the gulp pipeline to copy GCDS styles and scripts over to our static folder. 

- I also needed to add headers since GCDS loads fontawesome from cloudflare. Should they be more precise? I feel like its bad that we allow any css and js from cloudflare...
- I also removed some series() directives in gulp. They were useless the way they were configured. 

---


app/templates/main_template.html
```diff
@@ -12,6 +12,16 @@
   {% include 'partials/google-tag-manager-head.html' %}
   {% include 'partials/qualtrics-head.html' %}
 
+  <!-- GCDS icons -->
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.2/css/all.min.css"
+    crossorigin="anonymous">
+
+  <!-- GC Design System -->
+  <link rel="stylesheet" media="screen" href="{{ asset_url('stylesheets/gcds.css') }}" />
```
andrewleith, line 20: I wonder why we need to load 2 separate CSS files here - it would be nice if these were consolidated.

app/templates/main_template.html
```diff
@@ -12,6 +12,16 @@
   {% include 'partials/google-tag-manager-head.html' %}
   {% include 'partials/qualtrics-head.html' %}
 
+  <!-- GCDS icons -->
+  <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/font-awesome/6.4.2/css/all.min.css"
```
andrewleith, line 16: One thing to note here is that we are likely loading font-awesome twice now.  Once here for GCDS, and then again for notify itself.

andrewleith (commented):
This looks good to me.  Probably wouldn't merge it in just yet though.

Maybe we can use this as a base to try out some components and just make sure there are no incompatibilities with the rest of the app?

app/templates/main_template.html
amazingphilippe: I think they'll be removing fontawesome soon in favour of their own icons.

app/templates/main_template.html
amazingphilippe: We have our build pipeline to import a subset at least. So some icons will load twice for sure ... 😞

amazingphilippe:
> wouldn't merge it in just yet though.
h yeas! I don't know why I did a PR, What I wanted to do is to start our discoveries from a common branch. I'll close this!

PR #2049, opened 2025-01-15
Fix/check terms page for bad links

andrewleith:
# Summary | Résumé

This PR updates the gca pages test suite to include `/terms` even though it is technically an app page.  This way, we will still check it for dead links.

- [ ] `/terms` is tested as part of the gca suite


ben851 (approved):
LGTM

PR #2050, opened 2025-01-17
Add CSS generation on-the-fly

andrewleith:
# Summary | Résumé

This PR updates our frontend build pipeline so that we can take better control of it.  In particular, these changes enable us to implement better developer experience by allows us to generate our tailwind + css on the fly as we are developing in a few seconds, vs. waiting the 25s+ it was taking before to run `npm run tailwind`

## Improvements
- New npm script to watch for changes and rebuild CSS (`npm run watch`) 🎉 
- New make target for running this when you're developing (`make watch`)
- Smaller bundled CSS size 🎉 
- Increased productivity for frontend tasks 🎉 🎉 
- Added tasks and launch settings into devcontainer to automatically do this for you:
   - On the run and debug sidebar (1), choose "Python: Remote Attach" (2)
image
   - Press `cmd+shift+p`
   - Select **Run task...**
   - Select **Dev Environment**
   - Hit **Enter**

### Rebind `F5` keyboard shortcut for admin only
- Press `cmd+shift+p` and type `Open keyboard shortucts (JSON)`
- Add the following:
```
{
        "key": "f5",
        "command": "workbench.action.tasks.runTask",
        "args": "Dev Environment",
        "when": "config.useAdminDebug"
    }
```
### Build pipeline before
```mermaid
    graph LR
    
    subgraph js[Webpack / JS Pipeline]
        direction TB
        J0[style-loader: extract CSS from JS]
        J0 --> J1[babel-loader: Convert ES6+ to ES5]
        J1 --> J2[minification: Compress JS]
    end

    subgraph css[Webpack / CSS Pipeline]
        direction TB
        P0[css-loader: handles css imports]
        P0-->P1[style-loader: takes CSS from css-loader and injects into page]
        P1-->P2[postcss-loader: calls postcss]
        P2 --> P3[MiniCssExtractPlugin: Extract to .css file]
    end

    subgraph postcss[PostCSS]
        direction TB
        PC2[postcss-import: Combine @import files]
        PC2 --> PC3[tailwindcss: Generate utility classes]
        PC3 --> PCN[cssnano: minify css]
        PCN --> PC4[autoprefixer: Add vendor prefixes]
        PC4 --> PC5[purgecss: Remove unused CSS]
    end
    

    subgraph Gulp[Gulp]
        CB[Compile] --> CSS
        CB --> JavaScript
        CB --> Assets
        
        subgraph Assets
            direction TB
            E1[Copy Images] --> E2
            E2[Copy Fonts] --> E3
            E3[Copy Gov Assets]
        end
        
        subgraph JavaScript
            direction TB
            D1[Concatenate] --> D2
            D2[Minify] --> D3
            D3[Source Maps]
        end
        
        subgraph CSS
            direction TB
            C3[Minify]
        end
    end
    js --> css --> postcss --> Gulp
```

### Build pipeline after
```mermaid
graph LR
    subgraph tw[Tailwind]
        direction TB
        TW1[Compile CSS @imports]
        TW1 --> TW2[Remove unused CSS]
        TW2 --> TW3[Minify CSS]
    end

    subgraph js[Webpack / JS Pipeline]
        direction TB
        J0[style-loader: extract CSS from JS]
        J0 --> J1[babel-loader: Convert ES6+ to ES5]
        J1 --> J2[minification: Compress JS]
    end

    subgraph Gulp[Gulp]
        
        
        subgraph Assets
            direction TB
            E1[Copy Images] --> E2
            E2[Copy Fonts] --> E3
            E3[Copy Gov Assets]
        end
        
        subgraph JavaScript - all.min.js
            direction TB
            D1[Concatenate most internal js files] --> D2
            D2[Babel plugin] --> D3
            D3[Concatenate js from libs we depend on] --> D4
            D4[Minify] --> D5
            D5[Output to /static/javascripts/all.min.js]
        end
        
        subgraph JavaScript - indiviual components
            direction TB
            JS2[Babel plugin] --> JS4
            JS4[Minify using Uglify] --> JS5
            JS5[Output invidual files to /static/javascripts]
        end
        
        subgraph CSS
            direction TB
            C1[Minify using cleanCSS] --> C3
            C3[Output to /static/stylesheets/index.css]
        end
    end
    tw --> js --> Gulp
```


app/templates/main_template.html
```diff
@@ -167,7 +167,7 @@
 {% include 'partials/google-analytics.html' %}
 
 {% block page_script %}
-<script nonce="{{ request_nonce }}" type="text/javascript" src="{{ asset_url('javascripts/main.min.js') }}"></script>
+<script nonce="{{ request_nonce }}" type="text/javascript" src="{{ asset_url('javascripts/index.min.js') }}"></script>
 <script nonce="{{ request_nonce }}" type="text/javascript" src="{{ asset_url('javascripts/all.min.js') }}"></script>
```
amazingphilippe, line 171: Do we need to load both "index" and "all"? It seems like we prepend all with index in gulpfile.

gulpfile.js
```diff
@@ -86,35 +86,44 @@ const javascripts = () => {
...
     .pipe(dest(paths.dist + "javascripts/"));
 };
 
+const minifyIndividualJs = () => {
+  return src([
+    paths.src + "javascripts/branding_request.js",
+    paths.src + "javascripts/formValidateRequired.js",
+    paths.src + "javascripts/sessionRedirect.js",
+    paths.src + "javascripts/touDialog.js",
+    paths.src + "javascripts/templateFilters.js"
```
amazingphilippe, line 106: Are these ever bundled? Do we load them piece by piece in our html templates? If not, we could remove this step?

gulpfile.js
```diff
@@ -86,35 +86,44 @@ const javascripts = () => {
...
-      ])
-    )
     .pipe(plugins.concat("index.css"))
-    .pipe(
-      dest(paths.dist + "stylesheets/")
-    );
+    .pipe(plugins.cleanCSS({ 
+      level: 2,
+    }, (details) => {
+      console.log(`${details.name}: Original size:${details.stats.originalSize} - Minified size: ${details.stats.minifiedSize}`)
```
amazingphilippe, line 124: Nice!

amazingphilippe: I'd love to keep this. Maybe as Original size - New size. Maybe it would help us keep track of bundle size changes every time we add stuff.

package.json
```diff
@@ -9,9 +9,9 @@
     "test": "jest --config tests/javascripts/jest.config.js tests/javascripts",
     "webpack": "webpack --config webpack.config.js",
     "build": "gulp",
-    "watch": "gulp watch",
+    "watch": "npx tailwindcss -i ./app/assets/stylesheets/tailwind/style.css -o ./app/static/stylesheets/index.css --watch ",
```
amazingphilippe, line 12: Do we need gulp in there?

webpack.config.js
```diff
@@ -7,12 +7,7 @@ module.exports = {
   // mode: "development", //development
   mode: "production",
   entry: {
-    main: ["./app/assets/javascripts/index.js", "./app/assets/stylesheets/tailwind/style.css"],
-    branding_request: ["./app/assets/javascripts/branding_request.js"],
-    formValidateRequired: ["./app/assets/javascripts/formValidateRequired.js"],
-    sessionRedirect: ["./app/assets/javascripts/sessionRedirect.js"],
-    touDialog: ["./app/assets/javascripts/touDialog.js"],
-    templateFilters: ["./app/assets/javascripts/templateFilters.js"],
+    index: ["./app/assets/javascripts/index.js"],
```
amazingphilippe, line 10: lovely

webpack.config.js
```diff
@@ -44,17 +39,6 @@ module.exports = {
...
-        test: /\.css$/i,
-        include: [
-          path.resolve(__dirname, "app/assets/stylesheets/tailwind")
-        ],
-        use: [
-          MiniCssExtractPlugin.loader,
-          "css-loader",
-          "postcss-loader"
-        ]
-      },
```
amazingphilippe, line 57: Good riddance!

gulpfile.js
andrewleith: These few are not bundled.  The thought when they were created was that since they are not often used, they could exist separately.  Although, the templateFilters one probably should be bundled now that I look more closely.

webpack.config.js
andrewleith: Yeah this file being named `main.js` but having a source of index.js when we ALREADY HAVE a file named `main.js` was just so super confusing.

This is still confusing - but less so :D

app/templates/main_template.html
andrewleith: We dont combine them - but i have no idea why.  I think we probably could just have the one.

package.json
amazingphilippe: So watch will only update on css changes, and not script changes?

gulpfile.js
andrewleith: So here is the original size coming in from the tailwindcli step. and out is the further minification that cleanCSS performs.  I don't think its exactly what you were thinking.

amazingphilippe: Ah! Gotcha.

package.json
andrewleith: Yeah - I didnt tackle watching the js - could be another PR!

Makefile
```diff
@@ -53,7 +53,8 @@ coverage: venv ## Create coverage report
 
 .PHONY: run-dev
 run-dev:
-	poetry run flask run -p 6012 --host=localhost
+	npm run watch & \
```
smcmurtry, line 56: This part isn't working for me - can we add this to a new `make watch` command?

pyproject.toml
smcmurtry, line 1: are there any real changes in this file? or is it just changing the order?

smcmurtry (changes requested):
I tested this locally and if I run `poetry run flask run -p 6012 --host=localhost` and `npm run watch` in separate terminals it all works as expected! The css is regenerated and reloading the browser gets the new styles.

I did have to run `npm run tailwind` first to generate the static assets, so I think we should change `npm run build` to `npm run tailwind` over here: https://github.com/cds-snc/notification-admin/blob/main/.devcontainer/scripts/installations.sh#L53

The debugger config also worked.

I am seeing an error at the end of the dev container set up script but that is not related to your PR.

Makefile
andrewleith: Good idea, ill make those changes!

pyproject.toml
andrewleith: I added `debugpy` to the project - i guess it tries to update other packages when you do that which is not ideal.  There is probably a way to add it without it doing that - I'll take a look.

andrewleith:
> I did have to run `npm run tailwind` first to generate the static assets, so I think we should change `npm run build` to `npm run tailwind` over here: https://github.com/cds-snc/notification-admin/blob/main/.devcontainer/scripts/installations.sh#L53

Could you expand on this?  I tried deleting my `static/` folder, and then rebuilding the devcontainer.  When I did, everything loaded normally.  

I don't mind making this change, because its probably what should be in that script anyway (`npm run build` only runs gulp) but I'm just worried we're missing some other bug.

smcmurtry:
@andrewleith okay I tried again and I think you are right. When building a new dev container, I was seeing `.devcontainer/scripts/installations.sh` fail when running `poetry run pre-commit install`, which is unrelated to your PR. Once I commented that line out, I am seeing the `app/static` folder being created correctly and the app loading in the web browser. All good!

pyproject.toml
smcmurtry: if you add the line manually and then run `poetry install` the diff is smaller. But what you did is okay, I just wanted to make sure I wasn't missing anything 😄

smcmurtry (approved):
Looks good!

PR #2051, opened 2025-01-17
Add docstring to function

smcmurtry:
# Summary | Résumé

I added a docstring to one of our api client functions. This will make it more obvious that the function takes the financial year as an argument.

None

### DIFF
--- a/app/notify_client/service_api_client.py
+++ b/app/notify_client/service_api_client.py
@@ -382,13 +382,13 @@ class ServiceAPIClient(NotifyAdminAPIClient):
 
     # TODO: cache this once the backend is updated to exlude data from the current day
     # @flask_cache.memoize(timeout=_seconds_until_midnight())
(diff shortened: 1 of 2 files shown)


app/notify_client/service_api_client.py
```diff
@@ -382,7 +382,24 @@ def get_service_history(self, service_id):
 
     # TODO: cache this once the backend is updated to exlude data from the current day
     # @flask_cache.memoize(timeout=_seconds_until_midnight())
-    def get_monthly_notification_stats(self, service_id, year):
+    def get_monthly_notification_stats(self, service_id: str, year: int):
```
andrewleith, line 385: Nice! Should we rename the param to make it even more obvious, too? i.e. `financial_year` instead of `year`?

smcmurtry: yes good idea!

whabanks (approved):
Once Andrew's suggestion is incorporated then this LGTM!

PR #2056, opened 2025-01-24
Get gc organisations data from S3

smcmurtry:
# Summary | Résumé

We now have a cached version of the gc-organisations datafile in S3 (both in Prod and Staging). This PR updates admin to use the file in S3.

* working on https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1151

The review app is broken because of an issue we are having getting new environment variables into that environment. So you will need to test locally.

1. Run locally with the following new line in your .env file:
```
GC_ORGANISATIONS_BUCKET_NAME=notification-canada-ca-staging-gc-organisations
``` 
2. Create a new service
3. on step 3, verify that the list of organisations looks correct
4. verify that you are able to finish creating the service successfully


jzbahrai (approved):
ran and tested this locally!

PR #2064, opened 2025-02-04
Feat/add caretaker notice

andrewleith:
# Summary | Résumé
This PR adds the caretaker notice behind a feature flag (`FF_CARETAKER`) to the top of the TOU prompt on sign in.

It also introduces a new component, `notice`, that we can use going forward.

## Storybook page with component examples:
- https://i73keltgogkvgmhuqbqhyut6ca0wxyyr.lambda-url.ca-central-1.on.aws/_storybook?component=notice

## Screenshot of TOU prompt
image

## Notice component options
image

- Login
- [ ] Ensure that the caretaker notice **DOES NOT** display
- Update your .env to have `FF_CARETAKER=true`
- Login
- [ ] Ensure that the caretaker notice **DOES** display


tailwind.config.js
```diff
@@ -71,10 +71,12 @@ module.exports = {
         hover: "#990c1a",
         border: "#6a0812",
         mellow: "#df3034",
+        light: "#fef6f6",
         300: "#D74D42",
       },
       white: "#FFF",
       blue: {
+        light: "#f5f9ff",
```
amazingphilippe, line 79: Curious to know if blue-100 would work for this? and if it doesn't, we could probably make `blue-100: "#f5f9ff"` For context, I'd rather we get rid of the named variants eventually in the future.

tailwind.config.js
```diff
@@ -132,6 +135,10 @@ module.exports = {
         100: "#D3E766",
         700: "#545E00",
       },
+      orange: {
+        light: "#fef9f6",
+        DEFAULT: "#C86117",
```
amazingphilippe, line 140: Same here. If we need an orange scale, I can work on a numerical scale. Otherwise, I believe I've got a numerical scale for yellow. Maybe that would work? Its a bit muddier, but it depends on what policy wants.

app/templates/components/notice.html
```diff
@@ -0,0 +1,60 @@
...
+        'bg': 'bg-red-light'
+      },
+      'warning': {
+        'icon': 'fa-circle-exclamation',
+        'color': 'text-orange',
+        'border': 'border-orange',
+        'bg': 'bg-orange-light'
+      }
+    }
+  %}
```
amazingphilippe, line 34: I feel like this could be more readable in a .CSS file. They're all classes, grouped by type of banner.

andrewleith: Yeah you may be right - I went back and forth.  In the end I didnt need a css file at all like this. 

I think it could all move, but do you think the icon will work via css?

app/templates/components/notice.html
```diff
@@ -0,0 +1,60 @@
...
+  <div class="notice notice-{{ type }} overflow-hidden rounded-lg px-12 pt-12 border-4 {{display[type]['border']}} {{display[type]['bg']}}" role="alert">
+    <div class="flex gap-4 items-baseline">
+        <div class="notice-icon {{ display[type]['color'] }}">
+          <i aria-hidden="true" class="mb-2 fa-solid fa-lg {{ display[type]['icon'] }}"></i>
+          <span class="sr-only">{{ screen_reader_text[type] }}: </span>
+        </div>
+        <div class="notice-content ml-6 pb-12">
+            {% if headingLevel >= 1 and headingLevel <= 6 %}
+                <h{{ headingLevel }} class="notice-heading heading-small md:heading-large mt-4" >{{ title }}</h{{ headingLevel }}>
```
amazingphilippe, line 52: I would be a bit more defensive for the heading tags. There is a possibility to create non-html tags using this technique. I would probably do an if/else with the entire h1/h2/h3. Lets not bother with levels 4-6. Seems unlikely in this context. Even h3 seems unlikely! 

If you want to push this further, this could be an optional parameter. As in: Always an h2, unless we want an h1 specifically.

app/templates/components/notice.html
```diff
@@ -0,0 +1,60 @@
...
+            {% else %}
+                <div>{{ title }}</div>
+            {% endif %}
+            <div>{{ message | safe }}</div>
```
amazingphilippe, line 56: I would use the {{ caller() }} trick here. In use, this means that we can put anything in the banner, including buttons and forms. It also looks more readable in templates

app/templates/components/notice.html
andrewleith: I don't think you can create non-html tags (could be wrong though!) 
Right now if you give it something other than a number, it crashes - I can make that cleaner.  I can reduce from h1 to h3 as well.

andrewleith: Just pushed a change so that it will only ever draw `<h1>` through `<h3>`, and if the dev gives an invalid param it defaults to a `<div>`

app/templates/components/notice.html
andrewleith: Very nice, will make that change!

app/templates/components/notice.html
```diff
@@ -0,0 +1,60 @@
...
+      'warning': _('Warning'),
+    }
+  %}
+  <div class="notice notice-{{ type }} overflow-hidden rounded-lg px-12 pt-12 border-4 {{display[type]['border']}} {{display[type]['bg']}}" role="alert">
+    <div class="flex gap-4 items-baseline">
+        <div class="notice-icon {{ display[type]['color'] }}">
+          <i aria-hidden="true" class="mb-2 fa-solid fa-lg {{ display[type]['icon'] }}"></i>
+          <span class="sr-only">{{ screen_reader_text[type] }}: </span>
+        </div>
+        <div class="notice-content ml-6 pb-12">
```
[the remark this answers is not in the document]
andrewleith: Yeah there is a lot of divs for sure.  What do you mean though? How would this reduce them?

tailwind.config.js
andrewleith: Yeah i wasnt sure if we needed a background color, but i thought it looked cool so i forged ahead.  I just took the icon/border colors, converted the to HSL, and cranked the L to 98%.  Happy to use whatever values you suggest!  And we can for sure get rid of the named ones and use -100 - that makes sense to me!

andrewleith: Yeah i wasnt sure if we needed a background color, but i thought it looked cool so i forged ahead. I just took the icon/border colors, converted the to HSL, and cranked the L to 98%. Happy to use whatever values you suggest! And we can for sure get rid of the named ones and use -100 - that makes sense to me!

smcmurtry (approved):
I tested locally and the feature flag is working correctly for me. French appears as expected. Nice!

PR #2078, opened 2025-02-10
remove javascript manipulating links into buttons

amazingphilippe:
# Summary | Résumé

A while back, we introduced some javascript to add a role=button to anchor tags styled like buttons. Users using voice control had some trouble with those links. After reviewing best practices, we've decided to go back and remore this javascript and let links be links, even if they are emphasized by a background colour (and look similar to a button)

On the home page, inspect the sign-in and create account links in the hero section. Both should be anchor tags, none should have role=button


smcmurtry (approved):
Looks good! I tested with Voiceover and the affected <a> elements are now behaving like links.

PR #2082, opened 2025-02-11
fix cache clearing

ben851:
# Summary | Résumé

Update the cache clearing action to latest since the previous version is deprecated.

Verify cache clear action works


andrewleith (approved):
As discussed, it might not be necessary to have the pip cache in there any more, but we can weed that out later.

PR #2088, opened 2025-02-17
Fix/use latest redis client

andrewleith:
# Summary | Résumé

This PR updates the `notification_counts_client` to use the annual limits redis client for stats, falling back to the new `/annual-limit-stats` endpoint when redis is not yet populated.

This data is used in 3 places:

## On the dashboard
- In the **Annual usage** section, the number of email+sms sent come from the redis client
image

## On the usage report
- On the **Usage report** page (`/monthly`), today's daily sends are used from the redis client, and aggregated with the monthly breakdown taken from the API 
image

## On the ready to send page
- On the ready to send page, the # of notifications left daily and annually are calculated using numbers from the redis client:
image


app/notify_client/service_api_client.py
```diff
@@ -59,6 +59,11 @@ def get_service_statistics(self, service_id, today_only, limit_days=None):
             params={"today_only": today_only, "limit_days": limit_days},
         )["data"]
 
+    def get_year_to_date_service_statistics(self, service_id):
+        return self.get(
+            "/service/{0}/annual-limit-stats/".format(service_id),
```
andrewleith, line 64:
```suggestion
            "/service/{0}/annual-limit-stats".format(service_id),
```

app/notify_client/notification_counts_client.py
```diff
@@ -1,50 +1,35 @@
...
+        if annual_limit_client.was_seeded_today(service.id):
+            # get data from redis
+            annual_limit_data = annual_limit_client.get_all_notification_counts(service.id)
+        else:
+            # get data from db
+            annual_limit_data = service_api_client.get_year_to_date_service_statistics(service.id)
+
+        # transform data so dashboard can use it
+        return {
+            "sms": annual_limit_data["total_sms_fiscal_year_to_yesterday"]
```
jzbahrai, line 26: can we pull these keys from the redis client? Kind of like this? 
```
from notifications_utils.clients.redis.annual_limit import (
    EMAIL_DELIVERED_TODAY,
    EMAIL_FAILED_TODAY,
    SMS_DELIVERED_TODAY,
    SMS_FAILED_TODAY,
    ```

File: app/notify_client/notification_counts_client.py [hunk 3]
```diff
@@ -82,69 +67,66 @@ def get_limit_stats(self, service: Service):
...
-        sent_today = self.get_all_notification_counts_for_today(service.id)
-        # We are interested in getting data for the financial year, not the calendar year
-        sent_thisyear = self.get_all_notification_counts_for_year(service.id, current_financial_year)
-
+        annual_limit_data = {}
+        if annual_limit_client.was_seeded_today(service.id):
+            # get data from redis
+            annual_limit_data = annual_limit_client.get_all_notification_counts(service.id)
+        else:
+            # get data from db
```
jzbahrai, line 75: # get data from the api which might be the db or redis

app/notify_client/notification_counts_client.py
```diff
@@ -82,69 +67,66 @@ def get_limit_stats(self, service: Service):
...
+        #     total_sms_fiscal_year_to_yesterday: int,
+        #     total_email_fiscal_year_to_yesterday: int,
+        # }
         limit_stats = {
             "email": {
                 "annual": {
                     "limit": service.email_annual_limit,
-                    "sent": sent_thisyear["email"],
-                    "remaining": service.email_annual_limit - sent_thisyear["email"],
+                    "sent": annual_limit_data["total_email_fiscal_year_to_yesterday"]
```
jzbahrai, line 90: same comment as above, using the keys from the utils client

app/notify_client/notification_counts_client.py
```diff
@@ -82,69 +67,66 @@ def get_limit_stats(self, service: Service):
...
+            annual_limit_data = service_api_client.get_year_to_date_service_statistics(service.id)
+
+        # notifications_v2: {
+        #     sms_delivered_today: int,
+        #     email_delivered_today: int,
+        #     sms_failed_today: int,
+        #     email_failed_today: int,
+        #     total_sms_fiscal_year_to_yesterday: int,
+        #     total_email_fiscal_year_to_yesterday: int,
+        # }
```
jzbahrai, line 85: shall we put a comment here on the difference between annual limit data vs daily limit data?
aka annual limit uses the receipt data, daily just uses the send data through the api/admin?
(we don't have to if you think it is overkill)

pyproject.toml
```diff
@@ -41,7 +41,7 @@ gunicorn = "22.0.0"
 itsdangerous = "2.2.0"  # pyup: <1.0.0
 newrelic = "10.3.0"
 notifications-python-client = "6.4.1"
-notifications-utils = { git = "https://github.com/cds-snc/notifier-utils.git", tag = "53.1.1"}
+notifications-utils = { git = "https://github.com/cds-snc/notifier-utils.git", branch = "update-redis-client" }
```
jzbahrai, line 44: lets update this

smcmurtry (changes requested):
The annual limit data on the dashboard looks correct based on my testing, but what I see on the usage report is not. I think the issue is that the usage report doesn't include today's totals.
Also the "Ready to send?" numbers look correct.

Dashboard:
Screenshot 2025-02-18 at 4 32 25 PM

Usage report:
Screenshot 2025-02-18 at 4 32 37 PM

smcmurtry (approved):
I re-tested and the usage page totals look good

PR #2094, opened 2025-02-19
Display error banners when saving duplicate template category and email branding entities

whabanks:
# Summary | Résumé
- Banners will now be displayed when saving duplicate template categories or email branding
- Added error message FR translations

TODO:
- [x] Add cypress tests

# Related work
* https://github.com/cds-snc/notification-api/pull/2464


app/main/views/email_branding.py
```diff
@@ -89,17 +90,22 @@ def update_email_branding(branding_id, logo=None):
...
+                colour=form.colour.data,
+                brand_type=form.brand_type.data,
+                organisation_id=form.organisation.data if form.organisation.data != "-1" else None,
+                alt_text_en=form.alt_text_en.data,
+                alt_text_fr=form.alt_text_fr.data,
+            )
+        except HTTPError as e:
+            if e.status_code == 400 and "already exists" in e.message:
+                flash(_(e.message), "error")
+                return redirect(url_for(".create_email_branding"))
```
andrewleith, line 108: This works well functionally, but you lose the record you were working on if its a new one, or if its an existing one you end up on a blank form.  Is there any way for the user not to lose context?

andrewleith (commented):
This looks good to me - just a question about losing context below.

app/main/views/email_branding.py
whabanks: We should be able to capture their current input and pass it back to the form when we redirect. I'll dig into it.

app/main/views/templates.py
```diff
@@ -1420,17 +1425,22 @@ def template_category(template_category_id):
...
+                description_fr=form.data["description_fr"],
+                hidden=form.data["hidden"],
+                email_process_type=form.data["email_process_type"],
+                sms_process_type=form.data["sms_process_type"],
+                sms_sending_vehicle=form.data["sms_sending_vehicle"],
+            )
+        except HTTPError as e:
+            if e.status_code == 400 and "already exists" in e.message:
+                flash(_(e.message), "error")
+                return redirect(url_for(".add_template_category"))
```
andrewleith, line 1443: I think the same context loss is happening here for TCs.  Also, when adding new TCs or branding, and the name conflicts, you end up losing all your changes and have to start over.  Not sure if thats worth fixing cause this is pretty edge-casey as it is.

whabanks: Fixed!

andrewleith (approved):
LGTM context is no longer lost! 🎉

PR #2098, opened 2025-02-24
Add guidance to nav

amazingphilippe:
# Summary | Résumé

This add guidance to the nav. 

- The guidance link is also added through GC Articles
- The guidance link is added to the desktop nav
- The guidance link is added to the mobile nav too. 

1. Guidance should link to the guidance page
2. Guide de reference should link to the FR guide de reference
3. Links should work in mobile and desktop views
4. Test from the contact us page (public non GCA page)


app/navigation.py
```diff
@@ -115,6 +115,10 @@ class HeaderNavigation(Navigation):
         "documentation": {
             "documentation",
         },
+        "guidance": {
+            "guidance",
+            "guide-de-reference",
+        },
```
andrewleith, line 121: Where does this get used? 🤔

andrewleith (approved):
LGTM just curious (and have been for awhile) how does navigation.py even get used in admin? Looking around quickly at the references in VS Code makes it seems like it isnt actually used but I'm sure I must be missing something.

app/navigation.py
amazingphilippe: This is weird, but it checks if the menu item is the current page or not. Its a bit useless since we've moved to GCA pages, but if this key is missing, the contact-us page won't render properly. 

When on GCA pages, there's another way to get the current page handled in one of the Aticles controller. ITs more automatic, so I did not need to change anything!

andrewleith: Can you link to the code though? Just curious on how this works!

amazingphilippe: Its all in this file!

Basically, my change is in one of the mappings, which is just some sort of dictionary/object.

Then in this function:
```python
def is_selected(self, navigation_item):
        if request.endpoint in self.mapping[navigation_item]:
            return self.selected_attribute
        return ""
```

It tries to match the current endpoint to one of the items in the mapping. Here, `is_selected("guidance")` would return the `selected_attribute` `(selected_attribute = "class=selected")` (yikes)

amazingphilippe: Seems like we could drastically improve this in terms of controller logic, but also accessibility. I had never noticed this only added a class.... I wonder if we have a way to also add the aria-current somehow...

PR #2100, opened 2025-02-26
Make legends as headings more robust

amazingphilippe:
# Summary | Résumé

When we create or send templates, we have screens for which the heading is the same as the legend for some radios. 

This cause a a11y error:
> Description: The `<legend>` element has no text.
> WCAG 2.2 Success Criterion:A 3.3.2 Labels or Instructions 

To fix this, we must make sure we set up forms to have actual labels. Then in our page templates, we can use that forma label as the page header, minimizing the risk that there is a mismatch with the form legend.

We also need to tweak the legend in select-input, to make sure each radio field has a label association. we just need to add-in the `id="{{ field.name }}-label"` attribute, which is reference via aria-labelledby later on.

Run automated tests like axe or ARC toolkit on:
- [ ] Create template
- [ ] Add recipients


smcmurtry (approved):
Looks good! tested in the review app

PR #2060, opened 2025-01-28
Restrict email reply-to domains to domains used by current team members

whabanks:
# Summary | Résumé
This PR updates the email reply-to validation, allowing only email domains matching the email domains of current team members.

# Test instructions | Instructions pour tester la modification
## Add a reply-to with an invalid domain
1. Add a new reply to where the email's domain is not used by any current team members
2. Note the error message

## Add a reply-to with a valid domain
1. Add a new reply to where the email domain **is** used but one of the current team members. Use a `+someString` alias to identify the notification
2. When you reach the "verifying this email address" screen, go to the notification table in your local DB and find the verification notification by searching for the alias you used in step 1
3. Change the status to `delivered`
4. Note that the process completes as expected

## Edit an existing reply-to to an invalid domain
1. Edit the reply-to you created in the steps above to a domain not used by any team members
2. Note the error message preventing you from moving forward

## Edit an existing reply-to to a valid domain
1. Edit the reply-to again and use a domain used by one of the service's team members, again using a `+someString` alias
2. Repeat the steps above, setting the notification to `delivered` in your local DB to complete the verification process
3. Note the reply-to was successfully updated.

PR #2102, opened 2025-03-05
fix: only get email domains from active users

sastels:
# Summary | Résumé

We added some code in #2060 to strict reply to domains to those of the service users. However, the code fails if there in an invited (ie hasn't accepted) user on the team

- invite a user to your team
- try to add a reply-to
- success!

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -64,7 +64,6 @@ from app.main.forms import (
     SMSPrefixForm,
 )
 from app.main.views.email_branding import get_preview_template
-from app.models.user import User
 from app.s3_client.s3_logo_client import upload_email_logo
 from app.utils import (
     DELIVERED_STATUSES,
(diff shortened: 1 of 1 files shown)


sastels:
Note that cypress is failing for the same test (Team Members) that is currently always failing.

P0NDER0SA (approved):
looks good!

app/main/views/service_settings.py
```diff
@@ -592,7 +592,9 @@ def service_add_email_reply_to(service_id):
     form = ServiceReplyToEmailForm()
     first_email_address = current_service.count_email_reply_to_addresses == 0
     is_default = first_email_address if first_email_address else form.is_default.data
-    g.team_member_email_domains = set([team_member.email_domain for team_member in current_service.team_members])
+    g.team_member_email_domains = set(
+        [team_member.email_domain for team_member in current_service.team_members if hasattr(team_member, "email_domain")]
```
jimleroyer, line 596: 👨‍🍳

PR #2106, opened 2025-03-10
Fix duplicate empty container

amazingphilippe:
# Summary | Résumé

This PR moves the empty list component out of the table component. This gives us more control on what should be shown when a table list is empty. 

the empty list was added a few weeks back as general design improvements, but caused a side effect on the notification table when empty:

image

To fix:
- reverted the change in table.html
- Added the empty-list component where it made sense. 

- Test each screen where the empty list is shown


smcmurtry (approved):
Looks good!

PR #2107, opened 2025-03-10
Format team permission page with actual user name

amazingphilippe:
# Summary | Résumé

In a team member's profile, we can use the actual user name to avoid having a gendered sentence in French. Easy fix in 3 strings:

- 'Remove {} from this team'
- 'Folders {} can see'
- 'Not available because {} has not added a phone&nbsp;number to their profile'

Before: 
image

After: 
Capture d’écran, le 2025-03-10 à 15 19 17

I've also adjusted the French strings. 

1. On a service with more than 1 team member:
2. Check another team member's settings
3. Check that the user name renders in first 2 strings
4. On a service with email auth activated, and one team member with no phone number (might need to be set directly in db)
5. Check that the radio hint displays name correctly in the 3rd string.


smcmurtry (approved):
Tested and it looks good!

PR #2112, opened 2025-03-13
Fix underscores in sending address

smcmurtry:
# Summary | Résumé

Fixing the bug described here: https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1688

I did a more explicit js re-write of the python function [`email_safe`](https://github.com/cds-snc/notification-admin/blob/main/app/utils.py#L338-L350) that actually sets the sending address. 

I added unit tests and included a test for the bug case. And I sync'd up the test cases between the js function and the python fn so that we know they are behaving the same way.

1. Open the review app 
2. Get to step 2 of the "create a new service" flow
3. observe that when you enter the sending address "rsp_uat", the correct sending address is displayed on the page:
Screenshot 2025-03-13 at 2 38 34 PM


app/assets/javascripts/utils.js
```diff
@@ -31,6 +31,38 @@
...
+      .join("");
+
+    // Replace multiple consecutive dots with a single dot
+    string = string.replace(/\.{2,}/g, ".");
+
+    // Replace sequences like ".-." or "._." with just "-" or "_"
+    string = string.replace(/(\.)([-_])(\.)/g, "$2");
+
+    // Disallow repeating ., -, or _
+    string = string.replace(/(\.|-|_){2,}/g, "$1");
```
whabanks, line 60: Minor nitpick. Repeating underscores are technically valid for the local part of the email address. Not sure, other than stylizing, why you'd want to use them though. I think removing them here is appropriate.

LGTM!
image

whabanks (approved):

app/assets/javascripts/utils.js
smcmurtry: I could see it being more confusing if you are writing the email by hand: `my__service` vs `my_service` vs `my___service` is hard to tell apart. 

But ultimately we just need to match what the existing python function is doing here.

PR #2118, opened 2025-03-18
Controlling navigation outside of templates

amazingphilippe:
# Summary | Résumé

This PR proposes a change to how we create and use our navigation menus. Instead of creating them manually in html, we can build them as various methods in navigation.py. 

That way, we can consistently loop through navigation items and navigation front-ends independently.

Each navigation method define a dict or items with the following props:

- **key**: the key of the item, related to the existing mappings. This is used to set aria-current for example.
- **label**: translated link text
- **view**: the name of the view associated with the link item. i.e. where the link should point to, minus the `main.` in templates, we can use this with `url_for()` to construct urls. 
- **url**: this can be used to directly set the url when needed. Useful when you need a static, or more specific `url_for()` parameters, such as within services. 
- **context**: when to show the item, based on any condition. i.e. permissions, user type. When none is set, everyone can see the nav item. 

## Controllers available

Each navigation can be called as needed in templates, but the controller also has a `get_nav()` where we can quickly group different navs that would usually occupy the same front-end area, i.e. how the header navigation show links for public, user and service pages.


smcmurtry (approved):
Code looks good! There is nothing to test since this is just groundwork for a follow up PR.

PR #2119, opened 2025-03-18
French formatting fixes

amazingphilippe:
# Summary | Résumé

1. Changing instances of email branding to become "image de marque DES courriels"
2. Removing the only median dot gender neutral formulation into something that doesn't require it. 
3. Formatting on the password requirement bullet points. 

1. View the email branding pages in FR. Check that the new plural form is in use. (change no. 1)
2. Reset a password for one of your account through admin, review the reset form (changes no. 2 and no. 3)


smcmurtry (approved):
Looks good!

PR #2121, opened 2025-03-21
Fix ugly focus outline that is visible when you log in

smcmurtry:
# Summary | Résumé

This has been bothering me for a while.

Before:
Screenshot 2025-03-21 at 3 54 26 PM

After:
Screenshot 2025-03-21 at 3 52 46 PM

1. log into the review app and verify that the focus outline is less ugly


andrewleith (approved):
LGTM 🔍

PR #2123, opened 2025-03-24
Contact form updates

amazingphilippe:
# Summary | Résumé

For https://github.com/cds-snc/notification-planning/issues/1771

Add content to the contact page when logged out to inform the public that we can't answer their questions if they are not public servants. 

Add a hint to the message box on step 2 to remind people not to submit protected information. 

1. When logged off
2. Go to the contact us page
3. Review content and link in FR and EN
4. Enter some information, and continue to step 2
5. Review hint content and link in FR and EN


smcmurtry:
This doesn't match what I see in Figma, am I looking in the wrong spot? https://www.figma.com/design/f6QILuxf94EHaStf2ejWJb/Contact-us-%7C-Nous-joindre?node-id=1522-857&t=RroaGRHAj7F3ogak-4

app/templates/views/contact/contact-us.html
```diff
@@ -35,7 +35,13 @@
...
         </div>
     </div>
-    <p class="mb-12">{{_("For other questions about GC Notify, use this form.")}}</p>
+    <div class="space-y-gutterHalf my-gutter">
+        {% set contact_link = '<a href="{url}" target="_blank">{title}</a>'.format(title=_("Government of Canada contacts page"), url=_("https://www.canada.ca/en/contact.html")) | safe %}
+    {% set hint_message = _("Do not include {} or personal information.").format(protected_link) %}
+        <p class="">{{_("We support government employees using GC Notify. We cannot answer questions from members of the public.")}}</p>
+        <p class="">{{_("Use this form to ask a question, get technical support, or give feedback.")}}</p>
+        <p class="">{{_("For other enquiries, go to the {} to find the department or agency that can help.").format(contact_link)}}</p>
+    </div>
```
andrewleith, line 44: This messages shows up for both authenticated and unauthenticated users.  The PR description mentioned it was intended only for unauthenticated users.

andrewleith (changes requested):
Looks good just need to make it so the message only shows to unauthenticated users (as stated in the PR description).

There are a few textual instructions at the top of the form, which might make this easy to miss.  Should we do anything to make this stand out more? Maybe use the notice component, or put the messages into a bullet list?

We could also reiterate this message right before send under the button like we do on other screens using this pattern:
image

amazingphilippe:
Ope, didn't use the right content! Thanks for catching it. I've used the latest I could find in the card/figma link

app/templates/views/contact/contact-us.html
amazingphilippe: Good catch. Fixed.

smcmurtry (commented):
I think the content is supposed to be split into 2 paragraphs: 
Screenshot 2025-03-25 at 9 56 58 AM

Aside from that this looks good!

amazingphilippe:
Ah yes, that wasn't clear from the Figma 😬  Hold on...

PR #2126, opened 2025-03-27
Allow users who just signed in to view terms of use before accepting them. 

amazingphilippe:
# Summary | Résumé

Adding a conditional so the terms page can be accessed from the login page. This is needed since we moved that page away from GC Articles. 

1. Log in
2. Before you continue, check all links on that page


andrewleith (approved):
LGTM!

smcmurtry (approved):

PR #2138, opened 2025-04-09
Fix a bug in the create service flow where `email_from` is missing

smcmurtry:
# Summary | Résumé

Fix a bug where the user was not being redirected back to step 2 in the create service flow when their step 2 data was missing.

This is the bug:
1. open the review app
2. go to create a new service
3. complete step 1
4. don't fill out anything in step 2
5. change the url to replace `/add-service?current_step=choose_service_name` with `/add-service?current_step=choose_organisation`  and hit enter
6. choose your organisation and click "create service"
7. you will see a 500 error 

If you follow these steps and don't see a 500 error, the bug is fixed.

### DIFF
--- a/app/main/views/add_service.py
+++ b/app/main/views/add_service.py
@@ -222,7 +222,7 @@ def add_service():
     data.update(form.data)  # add newly submitted data from POST
     # iterate through all forms and validate
     for step in WIZARD_ORDER:
-        temp_form_cls = get_form_class(current_step, government_type)
+        temp_form_cls = get_form_class(step, government_type)
         temp_form = temp_form_cls(data=data)
         if not temp_form.validate():  # something isn't right, jump to the form with bad / missing data
             return redirect(url_for(".add_service", current_step=step, government_type=government_type))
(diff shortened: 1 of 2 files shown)


app/main/views/add_service.py
```diff
@@ -222,7 +222,7 @@ def add_service():
     data.update(form.data)  # add newly submitted data from POST
     # iterate through all forms and validate
     for step in WIZARD_ORDER:
-        temp_form_cls = get_form_class(current_step, government_type)
+        temp_form_cls = get_form_class(step, government_type)
```
smcmurtry, line 225: This for loop is supposed to loop through the forms from previous steps and check that they have valid data. It was not doing that - it was only checking the current step.

jzbahrai:
when changing the url tp: /add-service?current_step=choose_organisation
it went to this page instead https://tmamjtpqyb3yil7e4h3gxbnomu0cdsqq.lambda-url.ca-central-1.on.aws/add-service and wouldn't move forward even when I clicked:

smcmurtry:
@jzbahrai okay that is also fine. I edited the test instructions. As long as the 500 error is prevented then the bug is fixed.

jzbahrai (approved):

PR #2139, opened 2025-04-09
Disclosure menu module improvements and examples

amazingphilippe:
# Summary | Résumé

This PR improves how we manage states for the the more-menu and menu modules.

- Kept javascript class handling simple and removed most of the class toggles, to move them in our CSS components
- The More-Menu module is mostly rewritten to better handle overflowing elements. Logic is separated in 3 steps:
    1. Calculate menu items and when they will overflow
    2. Group items that should be in the more-menu
    3. Run the grouping again when items are in the more-menu
- Wrote a `.nav-menu-item` CSS component to help align the different styles of elements in the navs.
- Moved our generic focus styles to `:focus-visible` so that they only appear when navigating with a keyboard. This means the focus style no longer trigger when using the mouse. 

## Test the main nav and more-menu
1. On the public site, in FR, resize the window to trigger the more menu and mobile navigation.
2. The more menu should appear when window is around 900px wide
3. Check that the menu can open, and be navigated with the keyboard. It navigates a bit differently from the "menu" module. This is already the case and this PR does not address this. It should still be able to be navigated by tabbing through the links. 

## Test the new storybook components
1. Navigate to /_storybook
2. Select the disclosure menu component
3. Test the menus with the keyboard. They should allow you to:
4. Enter the menu with ENTER
5. Navigate the menu with ARROWS
6. Exit the menu with ESC


app/templates/partials/nav/gc_header_nav.html
```diff
@@ -76,7 +76,7 @@
             {{ nav_menu_item(url=gca_url_for('whynotify'), localised_txt=_('Why GC Notify'), id_key='nav-head-why-notify', is_active=header_navigation.is_selected('why-notify')) }}
             {{ nav_menu_item(url=gca_url_for('features'), localised_txt=_('Features'), id_key='nav-head-features', is_active=header_navigation.is_selected('features')) }}
             {{ nav_menu_item(url=documentation_url(), localised_txt=_('API documentation'), id_key='nav-head-doc', is_active=header_navigation.is_selected('documentation')) }}
-            {{ nav_menu_item(url=gca_url_for('guidance'), localised_txt=_('Guidance'), id_key='nav-head-doc', is_active=header_navigation.is_selected('guidance')) }}
+            {{ nav_menu_item(url=gca_url_for('guidance'), localised_txt=_('Guidance'), is_active=header_navigation.is_selected('guidance')) }}
```
andrewleith, line 79: This one seems to be missing its `id_key`, is that intentional?

app/templates/views/storybook/disclosure-menu.html
```diff
@@ -0,0 +1,82 @@
+<h2 class="heading-medium">Disclosure Menu</h2>
+With class of <code>menu-overlay</code>
```
andrewleith, line 2: This option seems to push all the content down when it opens - is that intentional? I would have thought overlay meant it laid over top of any adjacent content without reflowing the document.

If it is intentional, do we even want to support this option?  It is a bit jarring.

app/templates/views/storybook/disclosure-menu.html
```diff
@@ -0,0 +1,82 @@
...
+                    {{ item }}
+                </a>
+            </li>
+        {% endfor %}
+    </ul>
+</nav>
+
+
+<h2 class="heading-medium">Disclosure Menu Bar</h2>
+With custom overlay appearance
```
andrewleith, line 24: This option did what I expected the previous one to do.

app/templates/views/storybook/disclosure-menu.html
```diff
@@ -0,0 +1,82 @@
+<h2 class="heading-medium">Disclosure Menu</h2>
+With class of <code>menu-overlay</code>
+<nav class="relative">
```
andrewleith, line 3: Do we want to add some tests here to ensure the new JS is behaving correctly?  Or are we covered by our existing test suite?

andrewleith (approved):
LGTM just a couple of questions.

app/templates/partials/nav/gc_header_nav.html
amazingphilippe: not intentional! I'll add it back. When I added the link (in another PR), I duplicated the API docs link. I thought Only nav-head-doc had an ID, but missed that all of the other links had. I don't think they do anything, but I'll add one!

app/templates/views/storybook/disclosure-menu.html
amazingphilippe: We do want both, I think. Sort of like a mix with the summary-details component. It can be useful on smaller screen when you might not have room to tap outside of the menu. Its just an option though! Should we name it something else?

app/templates/views/storybook/disclosure-menu.html
amazingphilippe: We don't seem to have any tests for the older components. Cypress tests would be best no?

amazingphilippe: Just saw some existing tests for the disclosure menu. It should be enough. I haven't really changed the script. The focus was on the "more-menu" portion.

app/templates/views/storybook/disclosure-menu.html
andrewleith: Yeah maybe a rename would help. Something with the word push or flow or something? Naming things is the hardest part!!

amazingphilippe: After thinking about it, I made the overlay behave as you'd expect, and provided an example on how you would make a custom menu that expands into the flow instead.

PR #2140, opened 2025-04-14
Clean up reports page

smcmurtry:
# Summary | Résumé

This PR:
- cleans up the styling on the reports page, it should now match [Figma](https://www.figma.com/design/rzjRLTIzVjM4sHpzI6e0Xl/Reports---Rapports?node-id=133-1465&t=GnIm2JAvf6TN1iwg-4)
- adds missing styles for the "generating" and "error" states
- adds an "expired" state when the report has expired
- adds tests

TBD

### DIFF
--- a/app/templates/views/reports/reports-table.html
+++ b/app/templates/views/reports/reports-table.html
@@ -15,7 +15,7 @@
   ) %}
     {% call row_heading() %}
       <div class="flex flex-col">
-        <div class="mb-4">
+        <div>
           {% if item.status == "ready" %}
           <a class="file-list-filename" href="{{ item.url }}">
             <span><time class="local-datetime-full">{{item.requested_at}}</time></span>
@@ -38,19 +38,27 @@
     
     {% call field(align='right') %}
       <div class="flex flex-col justify-end">
-        {% if item.status == "requested" %}
-            <div class="mb-4">
-              <div class="loading-spinner"></div>
-            </div>
+        <div class="mb-4">
+          {% if item.status == "requested" %}
+            <div class="loading-spinner"></div>
+          {% else %}
+            <div class="size-12"></div>
+          {% endif %}
+        </div>
+        {% if item.status == "requested" or item.status == "generating" %}
             <div class="file-list-hint">{{ _("Preparing report") }}</div>
         {% elif item.status == "ready" %}
-            <div class="mb-4">
-              <div class="size-20"></div>
-            </div>
             <div class="file-list-hint">
                 {{ _("Download before") }}
                 <span><time class="local-datetime-short">{{item.expires_at}}</time></span>
             </div>
+        {% elif item.status == "error" %}
+          <div class="file-list-hint text-red">{{ _("Failed. Try again") }}</div>
+        {% elif item.status == "expired" %}
+          <div class="file-list-hint text-red">
+            {{ _("Deleted at") }} 
+            <span><time class="local-datetime-short">{{item.expires_at}}</time></span>
+          </div>
         {% else %}
           {{ _(item.status) }}
         {% endif %}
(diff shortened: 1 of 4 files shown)


app/templates/views/reports/reports-table.html
```diff
@@ -15,7 +15,7 @@
   ) %}
     {% call row_heading() %}
       <div class="flex flex-col">
-        <div class="mb-4">
+        <div>
```
andrewleith, line 18: Nitpick: this `<div>` is probably not needed.

app/templates/views/reports/reports-table.html
```diff
@@ -15,7 +15,7 @@
   ) %}
     {% call row_heading() %}
       <div class="flex flex-col">
```
andrewleith, line 17: Pretty sure this div is unclosed.  I am not seeing HTML validation errors which i don't understand however.

app/templates/views/reports/reports-table.html
andrewleith, line 18: Its a bit hard to see blocks of markup here, for example, line 26.  Consider making sure each element (or atl least block-level elements) opens/closes on its own line.

andrewleith (changes requested):
Looks great.  There is a possible markup issue, commented below.  

This page would be a great fit to add to the [storybook](https://ekkrcgemmvx5swbzxbokl4hnoa0hbqig.lambda-url.ca-central-1.on.aws/_storybook).  It would allow you to mock out various permutations of this page, without having the need to generate reports, etc. That way we can ensure it will render correctly in each state, with the appropriate icon, etc.  We can even add cypress tests to this if necessary ([example of simple component tests,](https://github.com/cds-snc/notification-admin/blob/bbd5c418b1c2de40cf4d23f4b77cd16309f6f6eb/tests_cypress/cypress/e2e/admin/components/notice.cy.js), [example of more complex component tests](https://github.com/cds-snc/notification-admin/blob/cf99ac3330af7ea0f618e27d6196cbbf12c90fb5/tests_cypress/cypress/e2e/admin/components/remaining_messages_summary.cy.js))


Here is a sample snippet that could possibly fix some markup issues (this would replace lines 16-37):

```
    {% call row_heading() %}
      <div class="flex flex-col">
        {% if item.status == "ready" %}
          <a class="file-list-filename" href="{{ item.url }}">
            <span><time class="local-datetime-full">{{item.requested_at}}</time></span>
            - {{ item.report_type}} - {{ current_service.name }} {{ _("report") }}
          </a>
        {% else %}
          <div class="file-list-filename text-gray-grey1" tabindex="0">
            <span><time class="local-datetime-full">{{item.requested_at}}</time></span>
            - {{ item.report_type}} - {{ current_service.name }} {{ _("report") }}
          </div>
        {% endif %}

        <p class="file-list-hint">
          <span>{{ _("Requested by") }} {{ item.requesting_user.name }}</span>
          {% if item.requesting_user_id == current_user.id %}
            <span>({{ _("You") }})</span>
          {% endif %}
        </p>
      </div>
    {% endcall %}
```

I'll mark this as "request changes" to either fix or dispel the potential markup issue, the rest is optional.

app/templates/views/reports/reports-table.html
smcmurtry: Yes I think you're right. Fixed!

app/templates/views/reports/reports-table.html
smcmurtry: Removed

smcmurtry: Thanks, I made some changes to address this

smcmurtry:
Thanks for the review @andrewleith! It looks like I would have to refactor this into a couple components in order to add it to the storybook. Probably a good idea - I'll plan to do that at some point in the future.

andrewleith (approved):
LGTM

PR #2141, opened 2025-04-14
Add the current language to the report request

smcmurtry:
# Summary | Résumé

Add the user's current language to the data that is sent when the user requests a report.

* working on https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1846

- log into the review app
- generate a report
- there should be no errors and the report should be the same as before (the api code to generate the FR report is still in progress)

### DIFF
--- a/app/main/views/reports.py
+++ b/app/main/views/reports.py
@@ -1,4 +1,5 @@
 from flask import (
+    current_app,
     flash,
     jsonify,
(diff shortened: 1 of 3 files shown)


jzbahrai (approved):
tested, no errors

PR #2144, opened 2025-04-16
Hide org invite button

andrewleith:
# Summary | Résumé
Epic: [Rethink authentication for GC Notify](https://github.com/cds-snc/notification-planning/issues/1763)

This PR removes the ability for platform admins to send invitations through organizations.  The underlying code and structure is still there in case we want to revive this concept again later. 

### Admin cannot invite users via org
- Login as a platform admin
- Navigate to the admin panel, organizations
- Choose an organization
- Click Team members
   - [ ] Ensure there is no Invite team member button

### Admin can invite users via org
- Login as a platform admin
- Navigate to the admin panel, organizations
- Choose an organization
- Click Team members
   - [ ] Ensure there is an Invite team member button


smcmurtry (commented):
The code is working fine - the button is hidden as expected.

Does this need to be behind a feature flag? We aren't using this feature currently. I tried inviting myself to be an "organisation admin" one time and it got me into a very buggy state. So I think it would be best to just remove the button altogether. If we want to revive this feature we can always add it again.

smcmurtry (approved):

PR #2148, opened 2025-04-17
Allow any `gc.ca` email address as a reply to

andrewleith:
# Summary | Résumé
Update the validation to ensure ANY `gc.ca` address can be a reply to.  Before it was only allowing `xyz@gc.ca` rather than `xyz@subdomain.gc.ca`.

- [ ] You can successfully add your Service Canada email as a reply to to your service


smcmurtry (commented):
The code is working fine but I think we should update the error message as well to include the new domains. Maybe we could just add `@*.gc.ca`, but not sure if our users would understand that

Screenshot 2025-04-17 at 9 28 06 AM

andrewleith:
> The code is working fine but I think we should update the error message as well to include the new domains. Maybe we could just add `@*.gc.ca`, but not sure if our users would understand that
> 

I think `*.gc.ca` is a good idea.  If people are confused, we'll hear about it!

smcmurtry (approved):
looks good, just 1 test to fix

PR #2149, opened 2025-04-17
tweak translations for optional phone

amazingphilippe:
# Summary | Résumé

This adds the official FR content for the optional phone number story. Found a few bugs and tried to make things equal with email messages. Reorganized the translations a bit to put similar strings near each other. 

Also made the remove button use the link styles, since we have them. 

1. In your profile
2. Add a phone number and follow the instructions. 
3. Remove the phone number and follow the instructions
4. While you do these steps, compare with [Figma in FR and EN](https://www.figma.com/design/T6gN6GxBubbXvAHGtgUxHl/Security-features?node-id=2174-2626&t=NU64ML9IjaQ7D2NN-4)


app/templates/views/user-profile/confirm.html
```diff
@@ -20,7 +20,7 @@
   <div class="grid-row contain-floats">
     <div class="md:w-3/4 float-left py-0 px-0 px-gutterHalf box-border">
       <p>
-        {{ _('We’ve sent a security code to your ') + thing }}.
```
andrewleith, line 23: I'm pretty sure this is used for boh email and SMS which is why there is a variable (aptly named `thing` 😛).  So this will create an issue when a email verify code is sent.

andrewleith (changes requested):
Looks good other than `confirm.html`, comments below!

app/templates/views/user-profile/confirm.html
amazingphilippe: @andrewleith Just coming back to this after 3 weeks, I don't even remember how to get to this string. I think this would work? I'll do some testing later

andrewleith: Well I think with this change it will say we sent you a security code by text message even when it goes by email.  This page is used for both flows!

andrewleith (approved):
Looks good!

PR #2150, opened 2025-04-17
Set the report filename

smcmurtry:
# Summary | Résumé

This PR:
- updates the report filename to use part of the uuid
- creates a new endpoint so we can set the filename of the downloaded report file
- adds tests for the new endpoint

Figma: https://www.figma.com/design/rzjRLTIzVjM4sHpzI6e0Xl/Reports---Rapports?node-id=349-1286&t=sJAB06ExbITlzDfq-0

1. open the review app
2. go to the reports page
3. generate a new report
4. download the report
5. verify that the report filename matches the one that is displayed, and matches figma


andrewleith (approved):
LGTM - works as described.

PR #2154, opened 2025-04-24
Feat/add generic component a11y test

andrewleith:
# Summary | Résumé

This PR adds a generic component a11y test that will test each page listed on the Storybook page.  This will make it easier as we add new components, as we won't have to add an accessibility test for each one.

- Ensure the a11y test runs in ci and against all component pages


smcmurtry (approved):
Looks good! I can see the new test running in ci.

PR #2156, opened 2025-04-24
chore: remove extension that causes crashing on the new macbooks

andrewleith:
# Summary | Résumé

This PR removes the extension `mechatroner.rainbow-csv` from the dev container.  This seems to have been causing extension host crashes.


smcmurtry (approved):

andrewleith:
I did an extension bisect and it said this extension was the issue, but then i kept getting issues. Since I can't be sure this extension was the source of the issue, I'll close this PR.

PR #2159, opened 2025-04-28
Removing specific filters for heartbeat notifications

jimleroyer:
# Summary | Résumé

Removing specific filters for heartbeat notifications. This won't be needed as we moved the heartbeat notifications into a dedicated service and after we rework the existing heartbeat in the database so they are associated with the new service.

Related task:
https://app.zenhub.com/workspaces/notify-planning-core-6411dfb7c95fb80014e0cab0/issues/gh/cds-snc/notification-planning-core/560

Test the dashboards that were affected with the filters, make sure these before and after the changes report the same numbers for previous periods of time.


app/utils.py
```diff
@@ -122,7 +122,7 @@ def get_latest_stats(lang, filter_heartbeats=None):
         elif notification_type == "email":
             emails_total += count
 
-    live_services = get_live_services_count()
+    live_services = len(service_api_client.get_live_services_data()["data"])
```
jimleroyer: @andrewleith Was this one an improvement that might have gone after that PR was made (many months ago)? Just want to have your eyes on that and what should be used.

app/utils.py
```diff
@@ -133,11 +133,6 @@ def get_latest_stats(lang, filter_heartbeats=None):
     }
 
 
-@cache.memoize(timeout=24 * 60 * 60)
-def get_live_services_count():
-    return len(service_api_client.get_live_services_data({"filter_heartbeats": True})["data"])
```
jimleroyer, line 138: Hmm I guess we could keep that one, especially for memoization, and let it be called in the other call site, but we need to remove filter_heartbeats. @andrewleith what do you prefer?

jimleroyer (commented):
LGTM - let's have @andrewleith review as well.

app/utils.py
andrewleith: Yeah I think we should keep the caching here - no need to hit the db often for these generalized statistics!

ben851 (approved):
LGTM, added the caching back

PR #2162, opened 2025-04-29
Debug/add dashboard timings

andrewleith:
# Summary | Résumé

This PR adds some logging statements to capture the duration of the various queries that happen in the dashboard in order to troubleshoot the slow response times for trial services we are seeing.


andrewleith:
I started from a branch instead of main, whoops.

PR #2163, opened 2025-04-29
debug: add some perf timing logs

andrewleith:
# Summary | Résumé

This PR adds some logging statements to capture the duration of the various queries that happen in the dashboard in order to troubleshoot the slow response times for trial services we are seeing.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -358,6 +358,21 @@ def aggregate_notifications_stats(template_statistics):
 
 
 def get_dashboard_partials(service_id):
+    import logging
+    import time
+
+    logger = logging.getLogger(__name__)
+    timings = {}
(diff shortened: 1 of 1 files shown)


app/main/views/dashboard.py
```diff
@@ -358,6 +358,12 @@ def aggregate_notifications_stats(template_statistics):
 
 
 def get_dashboard_partials(service_id):
+    import logging
```
jzbahrai, line 361: can this import statement be put on top? and we use the logger from flask?

andrewleith: Oh sorry i wasn't really paying close attention here, I intended to revert all this later.  I'll fix it up.

jzbahrai (approved):

PR #2164, opened 2025-04-29
Fix async report footer overflow

andrewleith:
# Summary | Résumé

This PR adds some classes to ensure the report footer bar is contained properly.  It also adds some css so that the padding is adjusted when the footer sticks while scrolling.

- [ ] Footer content is in the container when not stuck
- [ ] Footer content is in the container when stuck
- [ ] Footer content is in the container when stuck on load (there is a different set of classes for some reason)

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -1,4 +1,5 @@
 import calendar
+import time
 from datetime import datetime, timedelta
 from functools import partial
 from itertools import groupby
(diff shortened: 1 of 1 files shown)


smcmurtry (approved):
Looks good! I'm seeing a slight shift in the button's horizontal position when scrolling to the bottom of the page. But I think this is good enough for now.

PR #2165, opened 2025-04-29
Fix the report page back link

smcmurtry:
# Summary | Résumé

This PR:
- changes the back link on the report page to use the referring page, and stores it in the user's session so it isn't lost if they refresh the page. The referrer is used because it could be the "sms in the last week", "email in the last week", "job with id=x", and any number of notification status filters.
- adds a test for this 

1. check the branch out and set `FF_ASYNC_REPORTS=true` in your .env
2. run and go to your "email in the last 7 days page"
3. click the "view reports" link
4. click back from the reports page and make sure you go back to the correct page

Repeat the above steps, but refresh the reports page before you click back.

### DIFF
--- a/app/main/views/reports.py
+++ b/app/main/views/reports.py
@@ -7,6 +7,8 @@ from flask import (
     flash,
     jsonify,
     render_template,
+    request,
+    session,
     stream_with_context,
     url_for,
 )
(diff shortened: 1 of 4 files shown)


andrewleith (changes requested):
LGTM.  Since there are 2 different possible back pages, should you have a scenario to make sure they both work?

smcmurtry:
I was having some trouble before getting that test to work, since we didn't have a way to set the headers when calling `client_request.get`. But I figured out how to add that to the test fixture and added a new test.

andrewleith (approved):
Nicely done! LGTM!

PR #2167, opened 2025-04-30
Make the "prepare reports" footer more responsive

smcmurtry:
# Summary | Résumé

This PR adds some javascript that adds event listeners to the "prepare report" button and the associated form. These listeners make the loading spinner appear as soon as the button is clicked, and send the form data via AJAX and prevent the page from reloading.

1. Check out locally
2. set `FF_ASYNC_REPORTS=true` in your .env
3. `npm rub build` , `make run-dev` and hard refresh the "email in last week" page
4. try generating some reports and verify that the page does not reload when the button is clicked and that the loading spinner appears without a delay

### DIFF
--- a/.github/workflows/backstage-catalog-helper.yml
+++ b/.github/workflows/backstage-catalog-helper.yml
@@ -1,8 +1,6 @@
 name: Backstage Catalog Info Helper
 on:
   workflow_dispatch:
-  schedule:
-    - cron: "0 0 * * *"
 
 jobs:
   update-catalog-info:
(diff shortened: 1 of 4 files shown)


andrewleith (commented):
It may be best to put this into its own file so later on we can find it more easily.   You'd just need to add it to `/app/assets/javascripts` and then include it in gulp in the `javascripts` object, for example:

```
paths.src + "javascripts/reportFooter.js",
```

smcmurtry:
@andrewleith I tried moving this to its own file like you suggested, but the js no longer seemed to be working - the loading spinner disappeared before the report was ready

smcmurtry:
This is not working as well as I would like - will revisit later

PR #2170, opened 2025-05-01
Feat/more responsive ux gen reports

andrewleith:
# Summary | Résumé
This PR changes how the report generation process works under the hood:
- Removes the full form post/page reload, using a AJAX post in its place to avoid the page refreshing
- Updates the spinner and button state using CSS

- The spinner should be more responsive when clicking to generate the report and the buttons should be physically and visually disabled until it is complete


smcmurtry (approved):
Tested, looks good.

PR #2174, opened 2025-05-02
Fix bug where the report url expires early

smcmurtry:
# Summary | Résumé

Fix a bug where the report url expires early.

We have been using the presigned s3 url that is saved in the db. But it is not actually necessary since the user now no longer downloads the report directly from s3. They download it via the `download_report_csv` endpoint in admin so we can set the appropriate filename. Admin already has access to this bucket, so we can just get the s3 object directly. 

Bug card: https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1898

We can't test in the review app since it doesn't have the right s3 access. It will only be possible to test the actual bug case in staging (where the report is more than 1 day old).

1. check out locally and run
2. generate a report
3. verify that you can still download it successfully


jzbahrai (approved):
tested ✅

PR #2181, opened 2025-05-14
Handle the case where an invited user already belongs to the service

smcmurtry:
# Summary | Résumé

Handle the case where an invited user already belongs to the service. The api returns a 400 error in this case: https://github.com/cds-snc/notification-api/blob/main/app/service/rest.py#L469-L470

This is causing multiple 500 errors in admin in prod. I haven't been able to reproduce the bug on my end yet. I was able to track it down with logs. More details in this slack thread: https://gcdigital.slack.com/archives/CNWA63606/p1747243726409519

card: https://app.zenhub.com/workspaces/notify-planning-614b3ad91bc2030015ed22f5/issues/gh/cds-snc/notification-planning/1914

We haven't been able to reproduce the bug. Existing tests should pass.


app/main/views/invites.py
```diff
@@ -63,11 +65,18 @@ def accept_invite(token):
...
+                    existing_user.add_to_service(
+                        service_id=invited_user.service,
+                        permissions=invited_user.permissions,
+                        folder_permissions=invited_user.folder_permissions,
+                    )
+                except HTTPError as e:
+                    if e.status_code == 400:
+                        flash(_("You have already been added to this service."), "error")
+                    else:
+                        flash(_("There was a problem added to this service."), "error")
```
jzbahrai, line 78: The was a problem while being added to this service, please try again?

andrewleith:
In the stack trace on PROD, the error isn't being thrown at line 470, but at 478.  This means the code somehow passes the check on line 470 (`if user in service.users:`) but then fails at line 478 due to the user being in the service.  

We might have caching issue and/or some kind of race condition.

Full stack trace [is here](https://gcdigital.slack.com/archives/CV38DBNVA/p1747317578736079).

app/main/views/invites.py
smcmurtry: updated to some new content

app/main/views/verify.py
```diff
@@ -90,5 +91,12 @@ def _add_invited_user_to_service(invited_user):
     invitation = InvitedUser(invited_user)
     user = User.from_id(session["user_id"])
     service_id = invited_user["service"]
-    user_api_client.add_user_to_service(service_id, user.id, invitation.permissions, invitation.folder_permissions)
-    return service_id
+    try:
```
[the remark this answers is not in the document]
smcmurtry: I updated to use [409](https://stackoverflow.com/questions/3825990/http-response-code-for-post-when-resource-already-exists) at Jimmy's suggestion

smcmurtry:
> In the stack trace on PROD, the error isn't being thrown at line 470, but at 478. This means the code somehow passes the check on line 470 (if user in service.users:) but then fails at line 478 due to the user being in the service.

Thanks for the investigation @andrewleith ! I put up an [API PR](https://github.com/cds-snc/notification-api/pull/2536/files#diff-02a4832ab85a57344c4179cf331c533fa2bf4367833127a54e479dcf986a66d3R479) to handle that error.

jzbahrai (approved):

PR #2183, opened 2025-05-20
Fix links colour

amazingphilippe:
# Summary | Résumé

Fixing a bug with how links are coloured. We had pseudo states for :visited and :link, but also setting a default colour directly on the anchor `<a>` tag. When you apply a color directly to an anchor (<a>) element, it does affect the usefulness of the :link and :visited pseudo-classes.

Since our goal is to colour the links blue or white, I've tried to remove the most critical declarations linked to the links disappearing on :active. I had to remove a bit more than you might think because all of these styles are layered like an onion. 

Bref, now the footer links don't disappear into thin air when we click on them, they still have a bright yellow BG and color when keyboard focusing on them. 

- [ ] In the footer, click and hold a link. It shouldn't disappear. 
- [ ] Using TAB, tab through the footer links. They should have a yellow background and black text colour.


andrewleith (approved):
LGTM

PR #2190, opened 2025-06-05
feat(xray): add xray trace ids to HTTP response headers

andrewleith:
# Summary | Résumé

This PR leverages [this work done in API](https://github.com/cds-snc/notification-api/pull/2272) in order to add xray traces to the HTTP headers.

- Run admin 
- Inspect the response headers
- [ ] Ensure that there is a `x-amzn-trace-id` header with a value

### DIFF
--- a/app/xray/context.py
+++ b/app/xray/context.py
@@ -1,0 +1,109 @@
+import logging
+
+from aws_xray_sdk.core.context import Context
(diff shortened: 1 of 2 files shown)


andrewleith:
After a discussion with @jimleroyer, it is now clear that this will not do anything further to get the xray ids into the response headers.  Back to the drawing board.

PR #2191, opened 2025-06-06
Feature/atom stats

amazingphilippe:
# Summary | Résumé

Testing if we can integrate "live" stats in WP using an atom feed. This method is super clean, and follows how we also have a csv endpoint for these stats. 

The stats are refreshed on request, so when WP fetches the feed every 12 hours or so, the numbers will update. Not exactly live, but pretty near, and neat

1. Test the endpoint
2. See if it can be integrated in wp, and be styled how we want it to.


app/main/views/index.py
```diff
@@ -291,6 +291,33 @@ def activity():
     return render_template("views/activity.html", **get_latest_stats(get_current_locale(current_app), filter_heartbeats=True))
 
 
+@main.route("/activity/atom", endpoint="activity_atom")
```
andrewleith, line 294: Very cool!  Since this is an unprotected route, let's add some caching so we aren't querying for data every time. I'll grab you an example on how to do this.

app/main/views/index.py
```diff
@@ -291,6 +291,33 @@ def activity():
...
+def activity_atom():
+    stats = get_latest_stats(get_current_locale(current_app), filter_heartbeats=True)
+    now = datetime.now(timezone.utc)
+    updated_total = now.isoformat()
+    updated_services = (now - timedelta(minutes=2)).isoformat()
+    updated_emails = (now - timedelta(minutes=3)).isoformat()
+    updated_sms = (now - timedelta(minutes=4)).isoformat()
+
```
andrewleith, line 302: What is the purpose of these statements? Why are you calculating 2, 3, and 4 minutes ago?

app/main/views/index.py
```diff
@@ -291,6 +291,33 @@ def activity():
...
+            **stats,
+            updated_total=updated_total,
+            updated_services=updated_services,
+            updated_emails=updated_emails,
+            updated_sms=updated_sms,
+        ),
+        200,
+    )
+    response.headers["Content-Type"] = "application/atom+xml; charset=utf-8"
+    response.headers["Cache-Control"] = "max-age=3600"
```
andrewleith, line 315: I'm not sure you should be setting cache headers in the response like this explicitly.  Looking a little further, this header will be stripped by some of our middleware anyway.  Do you want to cache this response?

andrewleith (changes requested):
Very cool stuff!  I've left a couple of comments on caching, in particular:

- The data that is requested here (via `get_latest_stats`) is already cached, so even without adding anything this method won't hit the DB often
- you could cache this route using the `@cache` decorator (from `app`) like this:
   - `@cache.memoize(timeout=12 * 60 * 60)`
   - This would make it not even do this processing more than once every 12 hours
- Adding `cache-control` headers to responses won't work in notify because it is set to strip them out (except for static resources) - but the @cache decorator will solve that

And then also a question about the dates and what they are used for.

app/main/views/index.py
amazingphilippe: Sounds good! For context, WP only hits this endpoint twice a day by default to update the feed content. However doubling it up might be a good idea in case I'm wrong...

andrewleith (approved):
LGTM

PR #2196, opened 2025-06-10
Feat/annual totals redis

smcmurtry:
# Summary | Résumé

The `get_monthly_notification_stats` api call is taking ~1 second for some services. This data is monthly, but we are just aggregating in annually, adding today's totals, and using it to display SMS/email sent this year on the dashboard. We already have this data in Redis, if the service has sent today. So we can use that if it is available, and greatly reduce the load time for the dashboard.

This PR:
- adds a `get_annual_data` function, that gets data from redis if available, else it gets it from the same api call as before
- unit tests for `get_annual_data`
- new type definitions to keep track of all the data structures

1. log into the review app
2. verify that the annual totals on the dashboard match what you are seeing on staging, both for services that have sent today, and services that have not sent today.

### DIFF
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -403,13 +403,6 @@ def get_dashboard_partials(service_id):
     bounce_rate_data = get_bounce_rate_data_from_redis(service_id)
     timings["get_bounce_rate_data"] = (time.time() - start) * 1000
 
-    # get annual data from fact table (all data this year except today)
-    start = time.time()
-    # annual_data = service_api_client.get_monthly_notification_stats(service_id, get_current_financial_year())
-    # get annual_data from redis
-    annual_data = annual_limit_client.get_all_notification_counts(service_id)
-    timings["get_monthly_notification_stats"] = (time.time() - start) * 1000
-
     start = time.time()
     annual_data = get_annual_data(service_id, dashboard_totals_daily)
     timings["get_annual_data"] = (time.time() - start) * 1000
--- a/app/main/views/dashboard.py
+++ b/app/main/views/dashboard.py
@@ -31,6 +31,7 @@ from app.models.enum.bounce_rate_status import BounceRateStatus
 from app.models.enum.notification_statuses import NotificationStatuses
 from app.models.enum.template_types import TemplateType
 from app.statistics_utils import add_rate_to_job, get_formatted_percentage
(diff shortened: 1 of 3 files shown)


app/main/views/dashboard.py
```diff
@@ -361,21 +362,6 @@ def aggregate_notifications_stats(template_statistics):
 def get_dashboard_partials(service_id):
     timings = {}
 
-    def aggregate_by_type(data, daily_data):
```
smcmurtry, line 364: I moved this function out of the enclosing function and renamed to `aggregate_by_type_daily` since there was another function with the same name.

app/main/views/dashboard.py
```diff
@@ -467,14 +448,29 @@ def aggregate_by_type(data, daily_data):
...
+        "sms": counts["sms"] + daily_data["sms"]["requested"],
+        "email": counts["email"] + daily_data["email"]["requested"],
+    }
+
+
 def _get_daily_stats(service_id):
     # TODO: get from redis, else fallback to template_statistics_client.get_template_statistics_for_service
     all_statistics_daily = template_statistics_client.get_template_statistics_for_service(service_id, limit_days=1)
     stats_daily = aggregate_notifications_stats(all_statistics_daily)
-    dashboard_totals_daily = (get_dashboard_totals(stats_daily),)
```
smcmurtry, line 474: Why was this a tuple?! I have no idea. It just made the code more complicated.

app/main/views/dashboard.py
```diff
@@ -687,3 +683,34 @@ def get_column_properties(number_of_columns):
...
+    if not current_app.config["REDIS_ENABLED"] or not annual_limit_client.was_seeded_today(service_id):
+        annual_data = service_api_client.get_monthly_notification_stats(service_id, get_current_financial_year())
+        aggregated_annual_data = aggregate_by_type_daily(annual_data, dashboard_totals_daily)
+        return aggregated_annual_data
+
+    # get annual_data from redis
+    annual_data_redis = annual_limit_client.get_all_notification_counts(service_id)
+
+    return {
+        "email": dashboard_totals_daily["email"]["requested"] + annual_data_redis["total_email_fiscal_year_to_yesterday"],
```
jzbahrai, line 714: do we need to use dashboard_total_daily here? can we use annual_data_redis[sms/email_delivered_today] ?

PR #2204, opened 2025-06-18
chore: make text lowercase in stat section on home page

andrewleith:
# Summary | Résumé
This PR makes the text in the stats section on the home page all lowercase.

# Related cards
* https://github.com/cds-snc/notification-planning/issues/1003

Since the articles page is hooked directly to prod, there is no real good way to test the end result.

Locally, you can hit http://localhost:6012/activity/atom?lang=fr and http://localhost:6012/activity/atom?lang=en and ensure all of the `<title>` elements have lowercase text in them.


app/main/sitemap.py
```diff
@@ -18,7 +18,7 @@ def get_sitemap():
                     {"href": gca_url_for("features"), "link_text": _("Features")},
                     {"href": gca_url_for("new_features"), "link_text": _("New features")},
                     {"href": gca_url_for("register_for_demo"), "link_text": _("Register for a demo")},
-                    {"href": gca_url_for("whynotify"), "link_text": _("By and for the GC")},
+                    {"href": gca_url_for("whynotify"), "link_text": _("By and for the Government of Canada")},
```
andrewleith, line 21: This has nothing to do with the fix at hand, but was causing a cypress test to fail.  

Our rule is that all links to our articles pages must have the same link text as the page title.  This is to ensure that in this case.

PR #2209, opened 2025-06-23
Changed the TOU continue step to take the next_url

jzbahrai:
# Summary | Résumé

When using a security key, we are redirected to the login_events page twice.
The first time it has the TOU instruction, the second time it only has the login_events.

Changed the continue step to explicitly take the `next_url` so that the browser knows what page to go to.
In the previous flow, it redirects to the request.path, which in the case of fido2 is its login page -> it says its authenticated -> so goes to the login_events page (and doesn't have the TOU this time around)

Go to staging
1. Login
2. Set up your security key
3. Login using your security key
4. You will see two login_event pages (one with the TOU and one without)

Go to the review app
1. Login using your security key
2. you shouldnt see two login event pages

### DIFF
--- a/app/templates/components/tou-prompt.html
+++ b/app/templates/components/tou-prompt.html
@@ -2,7 +2,7 @@
 {% from "components/terms.html" import tou_heading_list %}
 {% from "components/notice.html" import notice %}
 
-{% macro tou_prompt() %}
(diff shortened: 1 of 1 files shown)


andrewleith (approved):
Works like a charm!

PR #2210, opened 2025-06-24
Fix for the mobile menu on iOS

smcmurtry:
# Summary | Résumé

We have a bug where none of the links work in the mobile menu on iOS, either with chrome or safari. 
This is a fix for that.

1. open the review app link on iOS (I use chrome `menu > Cast, Save and Share > Create QR Code` to avoid having to type the address in my phone)
2. try opening the home page menu and clicking the links to verify that it is working
3. test on both chrome and safari
4. Also test on desktop safari


andrewleith (approved):
The phone number now works in safari on small screen sizes!  I am not able to test on a mobile device though unfortunately.

Do we have to get rid of the functionality that clicking outside the menu closes it?  I think people do expect that on mobile.  

That said, if its critical to the fix, it is probably ok since we barely have any mobile users anyway.

smcmurtry:
I'm going to merge this to get the fix deployed, but I'll spend a bit of time today to investigate if it's possible to re-enable the "click outside the menu to close" functionality without re-introducing the bug

PR #2215, opened 2025-06-30
Remove phone number formatting on ios

amazingphilippe:
# Summary | Résumé

Fixes a bug where ios would add an href to 9 digit numbers with specific space formats, ie: "xxx xxx xxx" Which it believed to be phone numbers. 

before:
Capture d’écran, le 2025-06-30 à 12 00 57

after:
Capture d’écran, le 2025-06-30 à 12 01 25

1. With an ios device, check the home and activity pages to confirm numbers are not auto-formatted


amazingphilippe:
This was already fixed in #2212

PR #2216, opened 2025-07-02
Fix job report availability by retention

whabanks:
# Summary | Résumé
- Services can no longer download job reports that contain notification data older than their retention period. This applies to services with retention periods <= 3 days
- Add inline hints to indicate why the service can no longer access a given job report

TODO


whabanks:
Closing as this our policy was reworded to allow for data outside of a services retention period to exist temporarily in generated reports.

PR #2218, opened 2025-07-02
Prevent report gen on jobs older than retention

whabanks:
# Summary | Résumé

This PR updates the `view_job` page when viewing archived jobs i.e. Jobs that are older than the services retention period.

- Ensure that the "download report" section is hidden
- Ensure that the "prepare report" footer is hidden.
- Removed the `FF_ASYNC_REPORTS` feature flag

1. Perform a bulk send
2. Find the job in the DB and set `archived` to `True`
3. View the job from the dashboard
4. Note that both the download report section and prepare report footer is not visible

Alternatively / even better: If you have a service with jobs send > 7 days ago try viewing one of those jobs instead.

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -81,7 +81,6 @@ class Config(object):
     FF_RTL = env.bool("FF_RTL", True)
     FF_ANNUAL_LIMIT = env.bool("FF_ANNUAL_LIMIT", False)
     FF_CARETAKER = env.bool("FF_CARETAKER", False)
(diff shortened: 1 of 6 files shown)


whabanks:
> The logic to hide the reports footer for old jobs is working for me.
> 
> So this PR is intended to release the async reports feature for the jobs page? In that case you need to do one more thing: remove the old "download reports" button from here: 

I reinstated the FF for now and added a few checks so we can merge this and preserve the old download link for jobs until we're ready to release the feature.

smcmurtry (approved):
Looks good!

PR #2224, opened 2025-07-10
Feat/2fa badges

andrewleith:
# Summary | Résumé

This PR continues the work for auth types v2 and adds:
- Password challenge when changing 2fa options
- Displays list of existing 2fa keys when changing 2fa options
- Adds verification status badges to email and sms 2fa options
- Updates translations to latest
- Redirects to setting phone number if sms option is selected for 2fa but no number set

## Before
image

## After
image
image

# TODO: 
- Will add more tests in a follow-up PR

---

- [ ] Ensure you are challenged for a password when clicking 'change' on 2fa settings
- [ ] Ensure list of security keys is presented on the 2fa settings page
- [ ] Ensure "verified" badge is displayed for "receive a code by email"
- [ ] Ensure no badge is displayed for "receive code by text message" when no phone number set in profile
- [ ] Ensure "not verified" badge is set for "receive a code by text message" when phone number has not been verified
- [ ] Ensure  "verified" badge is set for "receive a code by text message" when phone number has been verified
- [ ] Translations are up to date
- [ ] Redirect occurs when choosing text message for 2fa but no phone number set


app/main/views/user_profile.py
```diff
@@ -420,14 +444,42 @@ def user_profile_2fa():
...
+
+    if form.validate_on_submit():
+        session[HAS_AUTHENTICATED] = True
+        return redirect(url_for(".user_profile_2fa"))
+
+    return render_template(
+        "views/user-profile/authenticate.html",
+        thing=_("email address"),
+        form=form,
+        back_link=url_for(".user_profile_email"),
```
whabanks, line 484: I think this should be `.user_profile`. Currently when you click the link to change 2FA then go back without entering the password, it goes to the add/change email address page. Clicking cancel sends you back to `.user_profile`.

whabanks (changes requested):
Just one small issue with a back button going to the wrong page. Otherwise LGTM!

app/main/views/user_profile.py
andrewleith: Makes sense, good catch!
```suggestion
        back_link=url_for(".user_profile"),
```

whabanks (approved):
LGTM!

PR #2229, opened 2025-07-15
Remove old job report download link

whabanks:
# Summary | Résumé
This PR removes the old download job report section as it has been replaced with the async report functionality.

1. Send a job
2. Navigate to the job from the dashboard
3. Note that the old download link section near the header is no longer present


smcmurtry (approved):
Looks good! I tested in the review app.

PR #2235, opened 2025-07-21
Use new security_key_auth_type

whabanks:
# Summary | Résumé
- When setting your 2FA method to a security key, the new security_key_auth_type will be used.
- The selected key ID is recorded in the user.fido2_key_id column
- The key's name is now displayed in the security section of the user profile
- Removing the last security key you have defaults your 2FA back to email if it was previously set up to use a security key

### Setting a security key as 2FA
1. Add a security key if you don't have one already
2. Set your 2FA to the security key you just added
3. Check your user in the DB
    - Note that the `auth_type` is set to `security_key_auth` 
4. Navigate back to the user profile page
5. Note that the name of the security key is visible in the 2FA row

### Changing 2FA to email and SMS from security key
1. Change your 2FA to email
2. Check the User table
    - Note that the `auth_type` is `email`
3. Add a security key and change your 2FA back to security key
4. Repeat the above steps, changing the 2FA to SMS this time

### Deleting your last 2FA key
1. Set your 2FA method back to security key
2. Delete the last security key that you have (and are also currently using for 2FA)
3. Note that the 2FA method was reverted back to the default `email` and the security key was successfully removed

### Security key 2FA can still be used if there is no mobile number
1. Remove your mobile number
2. Change your 2FA to security key
3. Note that it's successful, no redirect to a page asking to add a phone number

### Security key 2FA can still be used if there is an UNVERIFIED mobile number present
1. Add a phone number but do not verify it
2. Change your 2FA to security key
3. Note it's successful, no redirect to a page asking to verify your phone number


whabanks:
Regarding the last test: `Deleting your last 2FA key`. I'm thinking we should either:
1. Prevent the user from removing a security key that's in use for 2FA
2. Warn the user in the confirmation message that if they remove the chosen security key their 2FA with be defaulted back to email

app/main/views/user_profile.py
```diff
@@ -489,10 +500,16 @@ def user_profile_2fa():
...
+            kwargs = {}
             if new_auth_type == "email":
-                auth_type = "email_auth"
+                kwargs["auth_type"] = "email_auth"
+                kwargs["fido2_key_id"] = None
             elif new_auth_type == "sms":
-                auth_type = "sms_auth"
+                kwargs["auth_type"] = "sms_auth"
+                kwargs["fido2_key_id"] = None
+            elif new_auth_type.startswith("key_"):
```
jzbahrai, line 510: why is this key_ ? 
what is the difference between what you are trying to do here vs:
```
elif new_auth_type == "new_key":
                # Redirect to add a new security key
                return redirect(url_for(".user_profile_add_security_keys"))
                ```

**Comment by jzbahrai** [comment]
first test:
```auth_type              | security_key_auth
blocked                | f
additional_information | {}
password_expired       | f
verified_phonenumber   | t
fido2_key_id           | 1e7c2bc8-8e11-4cbd-819e-5cce5e7d8f38
```
works! 

all 3 test scenarios pass.

only issue:
1. have an unverified phone number
2. add a key
3. change auth to key
^ this forces you to verify the phone number, that should not be the case

app/main/views/user_profile.py
```diff
@@ -465,7 +468,7 @@ def user_profile_2fa():
         data = [
             ("email", _("Receive a code by email")),
             ("sms", _("Receive a code by text message")),
-            *[(f"key_{key['id']}", _("Use '{}' key").format(key["name"])) for key in getattr(current_user, "security_keys", [])],
+            ("security_key", _("Use security key")),
```
jzbahrai, line 471: let's remove this change so you dont get conflicts with this Pr https://github.com/cds-snc/notification-admin/pull/2241

PR #2238, opened 2025-07-21
Changed a few control flows for auth types

jzbahrai:
# Summary | Résumé

Changed a few control flows for auth types

### Change your phone number
1. Have an exsiting verified phone number on your profile, and sms_auth type
2. Go an edit the phone number (not remove)
3. Number should be unverified
4. Auth type should be email

# Change security key, redirect
1. Delete existing security keys 
2. Go to the 2fa change page
3. Click on the add security key
4. After you add the security key, you should be redirected to the 2fa page

### DIFF
--- a/app/main/views/user_profile.py
+++ b/app/main/views/user_profile.py
@@ -147,6 +147,9 @@ def user_profile_mobile_number():
 
         # if they are posting the button "edit"
         if edit_or_cancel_pressed == "edit":
+            current_user.update(verified_phonenumber=False)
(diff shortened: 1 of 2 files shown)


smcmurtry (dismissed):
The "Change your phone number" test worked as described.

For the "Change security key, redirect" test, I couldn't get past the "Add a security key" page. I give it a name, click "continue", but nothing happens and I'm stuck on that page. I am able to add security keys on staging.

PR #2241, opened 2025-07-22
changed how we display security keys

jzbahrai:
# Summary | Résumé
We want users to be able to pick any security key to login

Changed the wording on "use '{name_of_security_key}' security key" to "use existing security keys"
on the 2fa selection page.
This change should not show if there are no security keys for the user


whabanks (approved):
Merged this with #2235 locally and tested. LGTM, content change is present. Option isn't available if there are no security keys added to the user profile.

PR #2244, opened 2025-07-22
Fix 500 error when changing to an empty mobile number

smcmurtry:
# Summary | Résumé

We were getting a 500 error when someone took the following steps:
1. log into staging
2. visit the user profile page
3. Click "change number" on `/user-profile/mobile-number`
4. make the input field empty and click "Save"
5. Observe the 500 error

I added some logic so that this works the same as removing your number and updated a test.

1. log onto the review app
2. follow steps 2-4 above
3. observe that you are prompted for your password
4. your number is removed

### DIFF
--- a/tests/app/main/views/test_user_profile.py
+++ b/tests/app/main/views/test_user_profile.py
@@ -145,42 +145,6 @@ def test_should_redirect_after_mobile_number_change(
     )
 
 
-@pytest.mark.parametrize(
-    "bad_phone_number,expected_error_message",
-    [
-        ("123", "Not a valid phone number"),
-        ("abc123", "Not a valid phone number"),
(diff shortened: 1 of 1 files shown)


app/main/views/user_profile.py
```diff
@@ -164,6 +164,11 @@ def user_profile_mobile_number():
             return redirect(url_for(".user_profile"))
 
         elif form.validate_on_submit():
+            # If the mobile number is empty or None, treat it as a removal request
+            if not form.mobile_number.data or form.mobile_number.data.strip() == "":
```
jzbahrai, line 168: i think you need two lines here:
current_user.update(verified_phonenumber=False)
if current_user.auth_type = 'sms_auth':
current_user.update(auth_type='email_auth')

PR #2245, opened 2025-07-22
Fix bug where multiple "update" emails were being sent

smcmurtry:
# Summary | Résumé

This PR does a few things:
- fix the bug where multiple update emails were being sent after removing a phone number
- fix another bug where user data was being updated on the `/user-profile/mobile-number` page without the user clicking save
- fixed some other places where 2 API calls (`.update()` calls) were happening instead of 1

## Bug 1

1. open the review app
2. make sure you have a phone number set
3. remove your phone number
4. observe that you only receive 1 email

## Bug 2
1. open the review app
2. make sure you have a phone number set
3. click "change number"
4. click "cancel"
5. observe that you do not receive any emails

### DIFF
--- a/app/main/views/user_profile.py
+++ b/app/main/views/user_profile.py
@@ -147,14 +147,6 @@ def user_profile_mobile_number():
 
         # if they are posting the button "edit"
         if edit_or_cancel_pressed == "edit":
-            data_to_update = {
-                "verified_phonenumber": False,
-            }
-            if current_user.auth_type == "sms_auth":
-                data_to_update["auth_type"] = "email_auth"
-
-            # update once to avoid multiple emails to the user
-            current_user.update(**data_to_update)
             return render_template(
                 "views/user-profile/change.html",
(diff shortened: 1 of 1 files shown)


app/main/views/user_profile.py
```diff
@@ -148,9 +148,6 @@ def user_profile_mobile_number():
 
         # if they are posting the button "edit"
         if edit_or_cancel_pressed == "edit":
-            current_user.update(verified_phonenumber=False)
-            if current_user.auth_type == "sms_auth":
-                current_user.update(auth_type="email_auth")
```
smcmurtry, line 153: these updates were happening outside of the form submission - i.e. they would happen even if the user clicked "cancel".
I moved them inside the `if form.validate_on_submit()` logic below

smcmurtry: No - I ran the debugger, and when you click "Save", `edit_or_cancel_pressed=None` so this code isn't executed

PR #2246, opened 2025-07-22
Feat/auth v2 strings

amazingphilippe:
# Summary | Résumé

This PR updates some error messages and filed labels as described in this Figma. https://www.figma.com/design/T6gN6GxBubbXvAHGtgUxHl/Security-features?node-id=2901-3020&t=E44UFofbHFmzunvr-4

Here I focused on the quick fixes. Some will need some more complex PRs:
- Validation for phone numbers usually lives in utils. Do we want to re-architecture validation there? Or make a custom validator just for profile phone number?
- ~~Can we say that 2fa codes are 5 char long? Or is it preferable not to?~~ Not a problem!
- Validation for email adress now goes through the same validator, regardless if it is YOUR email address or AN email address. This has implications with the sending flow. 

Trigger screens as shown in the figma files, usually around account settings and login. Make sure to test in FR to spot missing translations.
## Verify code
- [x] Log-in, don't enter the right code right away
- [x] Enter a code with 4 chars
- [x] Enter a code with 6 chars
- [x] Enter the wrong code
## Profile settings
- [x] Try to change your phone
- [x] Delete your phone: check flash message
- [x] Try to add your phone: check flash message
- [x] Try to change your password: Enter the wrong current password, and an easy password: like 11111111 (8 identical digits)
- [x] Notice the new error messages
- [ ] Try to change your name, submit a blank field
- [x] Try to change your email, enter an email that already is in use for another account
- [x] Try to add a new security key, check hint text


jzbahrai:
Try to change your name, submit a blank field - throws a 500

PR #2248, opened 2025-07-23
Fix security key login when FF is off

whabanks:
# Summary | Résumé
- Previously if the FF was off, users with security keys would not be able to log in with their security key. Prior to having an auth_type for security key we blindly routed the user to the security key sign in flow if they had security keys. But their auth_type would still be set to sms or email.
- This PR re-instates the hard-coded routing and wrapped it in a FF check

1. Before launching locally, set `FF_AUTH_V2` to `False`
2. Set your 2FA to email
3. Add a security key to your profile
4. Log out, and clear your `user-<id>` redis key
5. Log back in
6. Note that you're taken to the security key sign in flow despite your auth_type set to `email_auth` in the DB

### DIFF
--- a/app/main/views/sign_in.py
+++ b/app/main/views/sign_in.py
@@ -1,4 +1,4 @@
-from flask import abort, flash, redirect, render_template, request, session, url_for
+from flask import abort, current_app, flash, redirect, render_template, request, session, url_for
 from flask_babel import _
 from flask_login import current_user, logout_user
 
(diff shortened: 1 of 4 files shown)


smcmurtry (approved):
Tested locally, looks good!

Why did you have to make the geolocation changes?

whabanks:
> Tested locally, looks good!
> 
> Why did you have to make the geolocation changes?

The CodeQL analyzer was complaining that it was a critical security vulnerability. I'm not sure why we did not see it on previous runs and just now. This code has been here for a long time.

PR #2252, opened 2025-07-24
Fix duplicate security key registration message

andrewleith:
# Summary | Résumé
This PR:
- Fixes where the success message is created when adding a key, preventing the user from seeing the message when the process doesn't actually complete
- Adds a check when creating a key that it isn't the same one they already had, and shows an error if it is (vs. doing nothing like it did before

- go to your user profile
- go to security keys
- add a key from chrome
- add another security key from chrome
   - [ ] You see the message `This security key is already registered. Please use a different key or remove the existing one first.`
- click cancel
   - [ ] You do not see the message `Added a security key` 
- add your yubi key
   - [ ] Adding yubi key succeeds


app/main/views/user_profile.py
```diff
@@ -371,19 +371,22 @@ def user_profile_security_keys_confirm_delete(keyid):
...
     form = SecurityKeyForm()
 
-    if form.keyname.data == "":
-        form.validate_on_submit()
-    elif request.method == "POST":
-        flash(_("Added a security key"), "default_with_tick")
-        result = user_api_client.register_security_key(current_user.id)
-        return base64.b64decode(result["data"])
+    if request.args.get("duplicate") is not None:
+        flash(_("This security key is already registered. Please use a different key or remove the existing one first."), "error")
```
jzbahrai, line 375: what is setting the duplicate arg?

jzbahrai (approved):

PR #2254, opened 2025-07-24
Gate security key changes behind password check

whabanks:
# Summary | Résumé
This PR is primarily for adding password checks when managing security keys similar to the checks we have when changing email address, phone number, etc. It also includes a few bug fixes and a bit of route refactoring.

- Adding or removing a security key from the manage security keys page now requires a password to be input
- Fix bug where back links on the add security key page weren't working when used from the 2FA flow
- Fix bug where a security key's UUID was being displayed in the confirmation message when a user is removing a key
- Added the /user-profile/security_keys/authenticate route
- Updated /user-profile//security_keys/authenticate to /authenticate-fido2 to differentiate between the password based authentication and testing of fido2 keys from the manage security keys page 

**Throughout the scenarios below, test the back links at all stages**

### Add a security key
1. Add a security key
2. Note that you're prompted for a password

### Remove a security key
1. Remove the key you just added
2. Note that you're prompted for a password

### Add a key from the 2FA page
1. Go to the manage 2FA page
2. Select the `Add a security key` radio button
3. You should not be prompted to enter a password when adding from this flow as the password prompt occurs up front when navigating to the 2FA management page
4. **Back link behaviour**: You should be brought back to the 2FA page when using the back links during this flow.


jzbahrai:
the first two tests work,

the third one: I deleted all my sc, but it keeps telling me I have ne (I archived the key in the pw manager as well), couldnt test scenario 3
Screenshot 2025-07-29 at 11 13 11 AM
Screenshot 2025-07-29 at 11 12 16 AM

PR #2266, opened 2025-07-30
Task/fix invite

jzbahrai:
# Summary | Résumé

When inviting a user, we need to respect their existing auth type if it is already set.

Case 1: exisitng user
1. Set your existing user(A)'s auth type to sms
2. Sign in with a new account (B) and set up a service
3. Invite (A) to the new service
4. A's auth type should be SMS

Case 2: new user
1. Invite a user to your service that doesn't exist in the Notify systems
2. When you login with that user - the auth type should be set to email


smcmurtry (approved):
I tested both scenarios in the review app and they worked great!

PR #2272, opened 2025-08-05
Template page changes for sample template work

andrewleith:
# Summary | Résumé
This PR adds the empty list component and Sample Library button to the templates page.

## Related cards
- https://github.com/cds-snc/notification-planning/issues/2124

## Figma
https://www.figma.com/design/47FhKMvOXm95IZFL9G8PYN/Bilingual-templates-and-services-%7C-Gabarits-et-services-bilingues?node-id=5342-7058&m=dev

Throughout these scenarios, test the action buttons to ensure that functionality is intact regardless of the state you're in with each scenario. Try to break stuff, theres a lot of states here and something is bound to be broken. 

### Scenario 1
1. Navigate to a service without any templates
2. Note the new empty state component is visible
3. Note that the `Sample library` button is visible with the `New` indicator/badge beside the button
4. Note that the page title is `Your templates`

### Scenario 2
1. Navigate to a service with a bunch of template 10 or so will do
2. Note the following buttons are visible
    - Create Template
    - Copy Template
    - New Folder
    - Sample Library (With `new` badge) 
5. Scroll up or reduce the size of your window so the page controls become sticky
6. Note all buttons are present while sticky and there are no layout issues
7. Select a template, and note the visible buttons change to:
    - Move
    - Add to new Folder
    - `1 Selection` Status text and `Clear` link are visible
8. Click the clear link -> note that the button visibility changes back to the state from step 2

### Scenario 3
1. In the same service as Scenario 2
2. Add a new folder that is empty
3. Enter the folder
4. Note that the page title is `Browse <folder name>`
5. Note the controls are the same as `Scenario 2 -> step 2`

### Scenario 4
1. Add an email template to the folder you just created (or if a folder with a template exists already navigate there)
2. Use the template filters to select only `SMS templates`
3. Note the empty state view is visible


app/templates/main_template.html
```diff
@@ -42,13 +44,23 @@
   />
   <script nonce="{{ request_nonce }}">
     window.APP_LANG = "{{ current_lang }}"
+    window.APP_COMPONENTS = {
```
whabanks, line 47: This is super janky, but was the only thing I could think of at the time to render and inject a Jinja template / component via JS. [See here for usage](https://github.com/cds-snc/notification-admin/pull/2272/files#diff-61e30dea530bede1567d39bca250e9e5430c4a7d1e264aff5dad5f585df349f6R348)

app/templates/main_template.html
```diff
@@ -42,13 +44,23 @@
...
       now: "{{ _('Now') }}",
+      new: "{{ _('New') }}",
       done: "{{ _('Done') }}",
       new_template_button: "{{ _('Create template') }}",
-      copy_template_button: "{{ _('Copy template') }}",
       new_folder_button: "{{ _('New folder') }}",
+      sample_library_button: "{{ _('Sample library')  }}",
+      copy_template_button: "{{ _('Copy template') }}",
       nothing_selected: "{{ _('Nothing selected') }}",
+      sample_library_url: "{% if current_service and current_service.id %}{{ url_for('.view_sample_library', service_id=current_service.id) }}{% else %}/sample-library{% endif %}",
```
whabanks, line 63: Normally we'd just set the `href` of the this link to `/sample-library` but since the library can be navigated to when your within a folder, where the url looks like: `.../templates/all/<folder_id>` simply appending that to the URL breaks the routing when you're navigating from there. Since this template is inherited by many other templates, we have to check if the `current_service` is actually available before building the URL or we get a `500` error when building the route.

smcmurtry (approved):
Tested in the review app and locally and it seems to be working fine.

It is alarming that we need to change so much code to add a button to this part of the app. We should look at rewriting this part of the codebase to make it easier to work on. 

Nice work!

PR #2274, opened 2025-08-05
De-dupicate menu with correct permissions

amazingphilippe:
# Summary | Résumé

This PR removes duplicate markup for the mobile nav. Instead relying on our disclosure component to render the main navigation for smaller screens. 

The nav controller needs to show the correct links based on user permissions.
- Most permissions make links visible
- The "view_activity" permission hides "Sent messages" and "Bulk sends"

A few thing to keep in mind when reviewing: This menu only show on pages that are not coming from GC Articles. So some public pages like sign-in, contact, activity. And all pages within the app when signed in.

### Log in as an admin:
1. The admin panel is available through the main nav

### In a different browser, log in as a non admin
1. Admin panel is not in the main nav

### In a common service with both your admin and non-admin sessions
1. Navigate to the teams page of any service you have in common
2. Toggle each permissions individually
3. The main nav on the other session should reflect the permissions for that team member

> Also important! Mirror your review/QA steps on Staging to make sure no regression are slipping through


smcmurtry (changes requested):
Tested in the review app. I found 2 regressions:
1. The order of the menu items for someone with the "Managa API integration" permission has changed:
Prod/Staging:
Image

Review app:
Image

2. The "Sign in" and "Français/English" links are causing some styling regressions on narrow screens:

Prod/Staging:
Image

Image

Review app:
Image

Image

amazingphilippe:
1. I think I can fix the order of the links, hopefully it doesn't break for another permission type
2. I should've marked the styling as an intentional change. I've been following some design ideas from the design system team.

PR #2279, opened 2025-08-12
current_password validation

amazingphilippe:
# Summary | Résumé

Adds a definition for current password form fields which bypass strength validation. This allows for the current password to validate against itself and render the correct error message. 

- [ ] Log in
- [ ] Go to change your password
- [ ] Enter the wrong password as your current password
- [ ] Enter weak password as your new password
- [ ] You should wee these messages:
- [ ] Test in French too. 

image

image


whabanks (approved):
LGTM!

PR #2301, opened 2025-08-26
Fix sample template edit/ preview flow

jzbahrai:
# Summary | Résumé

Fix sample template edit/ preview flow

Case 1:
1. Edit a sample template (not just click edit, but add data)
2. Click preview
3. Save

Case 2:
1. Try a different template from Case 1, edit a sample template (not just click edit, but add data)
2. Click preview
3. Click the backlink
4. Your data should be the one from (1)
5. Save

Case 3:
1. Edit a sample template (not just click edit, but add data)
2. click preview
3. Click the dashboard
4. Go back to viewing the same sample template from (1) - it should NOT have your data
5. Preview the templat (it should not have any changes)


smcmurtry (approved):
Tested in the review app, looks good!

PR #2302, opened 2025-08-27
Add an `example_subject` field

smcmurtry:
# Summary | Résumé

This is so we can fill variables in the subject with example content and display that on the preview page.

Figma:

Open the review app and navigate to some of the sample template preview pages. For the Appointment reminder templates should not have a variable in the subject on the preview page.


jzbahrai (approved):

whabanks (approved):
LGTM!

PR #2305, opened 2025-08-27
Fix translation for the `New` badge

whabanks:
# Summary | Résumé
- The new badge beside the sample templates button and inside the task blocks on the dashboard are now properly translated
- Adjusted the position of the new badge so it does not cover button text when displayed in french

1. Navigate to the dashboard
2. Change the language
3. Note that the `New` badge properly displays in French / English

1.  Navigate to the templates page
2. Change the language
3. Note that the `New` badge beside the Sample Library button is displayed in English / French
4. Note that in French the badge, which is longer, does not cover any button text


andrewleith (approved):
LGTM - the new badge on the dashboard and on the templates page are now localized.

PR #2312, opened 2025-08-27
Add a query param for Google Analytics

smcmurtry:
# Summary | Résumé

This will put the query parameter `?source=sample-template` in the url so we can track if someone saves a sample template to their service.

Test 1
1. open the review app
2. open a sample template and get to the edit page
3. click save
4. observe that `?source=sample-template` is in the url

Test 2
1. open the review app
2. open a sample template and get to the edit page
3. click preview
4. click save
5. observe that `?source=sample-template` is in the url


smcmurtry:
@andrewleith Yes that sounds good - I'll use `?source=ga_sample_template` though not `?source=ga_sample-template`

PR #2353, opened 2025-08-28
Put skip link in a nav, and label the nav.

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/2409

- [ ] Inspect the skip link
- [ ] It should be contained in a nav region
- [ ] the nav region should have a label.

### DIFF
--- a/app/templates/main_template.html
+++ b/app/templates/main_template.html
@@ -165,13 +165,6 @@
 </script>
 
 {% block body_start %}{% endblock %}
-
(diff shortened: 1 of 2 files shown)


andrewleith (commented):
I think the recommendation here is to put it into an existing `<nav>` (such as the one in `<header>`), rather than creating a new `<nav>`.  What do you think?

PR #2354, opened 2025-08-28
Redundant hidden legend

amazingphilippe:
# Summary | Résumé

fixes https://github.com/cds-snc/notification-planning/issues/2400

We already support h1 in legends. So I'm just removing the h1, adding the back link on its own, and passing the right prop to the select macro. 

On these screens:
- [ ] Add recipients
- [ ] Create new template
- [ ] h1 should be the only element in the `<legend>`


smcmurtry (approved):
Tested in the review app, looks good!

PR #2357, opened 2025-08-29
Remove manual group attr from fieldset

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/2375

We don't need to specify that a fieldset is a group if it contains a legend and some inputs of type radio or checkbox

- [ ] Find a radio or checkbox list
- [ ] Inspect html and see that the role attr is not re-defined.


andrewleith (approved):
LGTM

PR #2358, opened 2025-08-29
Let platform admin update auth type

jzbahrai:
# Summary | Résumé

Let platform admin change the auth type of a user

1. Go to the admin user and find yourself as a user
2. Change your auth type
3. Login with the new auth type
4. Change your auth type to security key
5. Go and change your auth type under the admin view (only email/ sms allowed)


smcmurtry (approved):
Tested manually on the review app. Works great!

PR #2361, opened 2025-08-29
Required select inputs (radios and chackboxes)

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/2373

- [ ] Find any radio or checkbox list
- [ ] Inspect html
- [ ] Find the aria-required attribute on the fieldset.


whabanks (approved):
LGTM! radio/check lists that are required have the `aria-required` attribute and those that are optional do not. Nice work!

PR #2368, opened 2025-09-03
A11y/fix dialog trigger

andrewleith:
# Summary | Résumé

This pull request focuses on improving accessibility and styling for the Terms of Use (TOU) dialog trigger, as well as making a minor adjustment to task list styling. The most significant change is switching the TOU dialog trigger from an anchor tag to a button element, which enhances semantic correctness and accessibility.

# Related cards
- https://github.com/cds-snc/notification-planning/issues/2403

- Go to the register an account page
- Interact with the "Read and agree to the terms of use button"
- Ensure it can be accessed via keyboard and mouse
- Ensure it updates when you agree to the terms
- Ensure you can submit the form when all form validations are passed


app/assets/stylesheets/tailwind/components/tou-dialog.css
```diff
@@ -32,5 +32,10 @@
       max-height: calc(100vh - 6rem);
     }
   }
+
+  button#tou-dialog-trigger:focus-visible {
+    @apply bg-yellow text-black border-b-2 border-blue no-underline outline outline-offset-0 outline-2 outline-yellow;
```
smcmurtry, line 37: Not sure if the focus styles are getting applied correctly or not. This is what I'm seeing in the review app:

Image

And this is the focus style in staging currently:

Image

Could we use a yellow outline or background here instead of blue? That would be more consistent with the other focus styles on the page.

smcmurtry (changes requested):
Looks good! I have a minor styling request.

app/assets/stylesheets/tailwind/components/tou-dialog.css
andrewleith: Yeah I added that but I didn't regenerate the CSS, doing that now.

PR #2371, opened 2025-09-03
Describe collapsible button with the controlled field label

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/2291

Review the content to be "Show folders" as it describes the action of the button. 
We can also leverage the fieldset label to describe the button content. 

- [ ] On a service with folders
- [ ] Add a team member
- [ ] With coiceover, navigate to "Choose folders"
- [ ] It should announce that "choosing folders" refers to "Folders this team member can see"


amazingphilippe:
@YedidaZalik I thought about it some more, and I think we might want to change the button text after all. Here's why:

1. Collapsed: "Show folders" The button action is to show the folder options, which you would then choose. 
2. Expanded: "Done" follows the checkbox options. The action is to confirm that you are done choosing.

PR #2374, opened 2025-09-03
Remove obvious empty spans

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/2413

Removed obvious spans which would never receive content. 

- [ ] On the create account form
- [ ] Inspect html and find no `<span></span>`


smcmurtry (approved):
Nice fix!

PR #2377, opened 2025-09-03
Terms of use modal - a11y fixes

smcmurtry:
# Summary | Résumé

Terms of use model - a11y fixes

- [x] role="dialog" on modal
- [x] aria-modal="true" on modal.
- [x] aria-labelledby="modal-title-id": Points to the heading inside the modal so the SR announces the title.
- [x] Set tabindex="-1" on container
- [x] Move focus to first interactive element inside modal: h2
- [x] Trap focus within modal: tested and this seems to be working
- [x] Return focus to triggering button: working
- [x] Use tabindex="-1" instead and move focus to the first actionable item inside 

https://github.com/cds-snc/notification-planning/issues/2402

Verify that the above checkbox items are completed in the review app


app/assets/javascripts/touDialog.js
```diff
@@ -85,10 +90,53 @@
     });
   }
 
+  /**
+   * Sets up the focus trap by finding all focusable elements within the dialog
+   */
+  function setupFocusTrap() {
+    // Get all focusable elements inside the dialog
+    focusableElements = dialog.querySelectorAll(
+      'button, [href], input, select, textarea, [tabindex]:not([tabindex="-1"])',
```
amazingphilippe, line 99: verynice

amazingphilippe (approved):
Nicely done! Looks good to me

PR #2379, opened 2025-09-04
Fix/rm details polyfill

smcmurtry:
# Summary | Résumé

This PR should fix this a11y issue: https://github.com/cds-snc/notification-planning/issues/2390

I can't actually reproduce the issue on OSX / Chrome / Voiceover, but it looks like it is caused by the `aria-ignore="true"` property getting set and then either not changing when the details element is expanded, or it does change but the browser does not pick this up correctly.

We were using the `details` polyfill to make the component work/be accessible in older browsers. But we no longer support IE and it looks like all modern browsers have good support for the `details` element: https://caniuse.com/details

I've tested with Voiceover and the native accessibility of the `details` element seems totally fine. So I think we can safely remove this code that is causing an a11y problem.

1. Open the review app
2. Start to create a new service and get to step 2 of 3
3. turn on voiceover and navigate the page. Verify that you can open the details element and that its content is read out when open.


whabanks:
VoiceOver wouldn't read the content of the details element after expanded. I tested it on the review app:

https://github.com/user-attachments/assets/5f7e45e9-89f5-40e0-a4ca-8eaea3f890a3

amazingphilippe (changes requested):
If the details element comes from a WordPress block, it seems to be missing some of the markup we target through CSS. 
image

We should try to style any direct children that are not a `<summary>` element with that gray border. And make sure that there is no vertical space in between direct children to ensure the border doesn't break. Usually, we should have only one direct children though.

PR #2380, opened 2025-09-04
Add confirmation msg when moving templates/folders

whabanks:
# Summary | Résumé
This PR adds a confirmation message after moving a template or folder.

https://github.com/cds-snc/notification-planning/issues/2338

Perform these steps with a screen reader enabled:

1. Move a template or folder into another folder
2. Note the confirmation message "Moved `x` template(s) or folder(s) to the `destination folder name` folder" (plural when moving multiple, singular when moving 1 item)
4. Note the message is read properly by screen readers after moving


app/translations/csv/fr.csv
```diff
@@ -2325,4 +2325,7 @@
 "Authentication type","Type d’authentification"
 "Change auth type","Modifier l'authentification"
 "Please choose a valid authentication type.","Veuillez choisir un type d’authentification valide."
-"templates found","gabarits trouvés"
\ No newline at end of file
+"templates found","gabarits trouvés"
+"template or folder","gabarit ou dossier"
+"templates or folders","gabarits ou dossiers"
+"Moved {} {} to the '{}' folder","Déplacement de {} {} vers le dossier '{}'"
```
amazingphilippe, line 2331: 2 suggestions:
- We can reduce "templates or folders" to just "items". Either one is as specific, but items is shorter!
-  In FR, the grammar is a bit difficult to adapt to your string interpolation, I think setting it up with one less format option would be better: (need to adapt the python code too)

```suggestion
"Moved {} items to the '{}' folder","{} items déplacés vers le dossier '{}'"
"Moved {} item to the '{}' folder","{} item déplacés vers le dossier '{}'"
```

whabanks: Makes sense, I started off using `item(s)` but second guessed myself hah. Will update!

PR #2382, opened 2025-09-04
Add aria label to validation summary

jzbahrai:
# Summary | Résumé

Added aria-labelledby="error-heading-id" where the heading has id="error-heading-id" (e.g.,
There was a problem
).

1. Added an ID to h2 heading
2. fixed the ol/ul tags


smcmurtry (changes requested):
One issue that still needs fixing:
> "Ensure focus reliably moves to the alert on form submission errors for consistent announcement"

When I submit the "new account creation" form with no data, the focus goes back up to the top of the page instead of going to the alert message when the page loads.

You might be able to use `data-module="autofocus"` for this.

PR #2385, opened 2025-09-09
Pill component - fix issue with missing `aria-labelledby` id

smcmurtry:
# Summary | Résumé

Fix for: https://github.com/cds-snc/notification-planning/issues/2349
Screenshot of issue: https://drive.google.com/drive/folders/121EPnNRWEbGrHPd9vKXNt6xw_SpAsh3V

Other elements on the page use `aria-labelledby="pill-selected-item"`, but no element with that ID exists on the page. This should point to the title of the currently selected tab (see w3c [example](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/)).

I added the id to the currently selected tab.

1. open the review app
3. navigate to a jobs page 
4. turn on voiceover and verify that the title of the currently selected tab is read out when you focus the "only show problem addresses" checkbox and the table

https://github.com/user-attachments/assets/da491902-700d-4b05-bfdb-4e81de655249

### DIFF
--- a/app/templates/components/pill.html
+++ b/app/templates/components/pill.html
@@ -22,7 +22,7 @@
       {% set count = item[3] %}
       {% set icon = item[4] if item|length > 4 else None %}
       {% if current_value == option or (loop.index == 1 and ns.current_value_not_in_items) %}
-        <a href="{{ link }}" class='pill-selected-item' aria-current='true'>
+        <a href="{{ link }}" class='pill-selected-item' id="pill-selected-item" aria-current='true'>
       {% else %}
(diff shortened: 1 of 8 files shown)


andrewleith (changes requested):
This seems to duplicate the id on the notifications report page (`/services/{xxx}/notifications/email`).  The id is being used in a few templates:
- `usage.html`
- `notifications.html`
- `monthly.html`

So we'll need to balance fixing the main notification report page with the bulk send report page (which I think is what you were focusing on).

andrewleith (approved):
LGTM - no more issues found using a11y tools.

PR #2388, opened 2025-09-10
Dashboard widgets as regions

amazingphilippe:
# Summary | Résumé

Fixes https://github.com/cds-snc/notification-planning/issues/2317
Fixes https://github.com/cds-snc/notification-planning/issues/2315

This PR tweaks some dashboard widgets so that they are announced in a few different ways.

1. Limits and problem widgets are now regions, so that users navigating by regions can land on those. They get a label from the contents of the widget via `aria-labelledby`
2.  I've simplified the markup of the problem email widget, so that we can use it as a description. So it might look a tiny bit different.
3. I've created a bunch of examples in the storybook.

Examples for problem email widget in storybook:
Capture d’écran, le 2025-09-10 à 13 49 28


amazingphilippe:
> discrepancy between the two types of regions

It might come down to UX testing, but my assumption was to focus on what can be found in each region, without being too verbose with the numbers. This means:
- For problem emails, you either have no problems, or X percent of problem.
- For limits, you either have plenty left, or near your limit. 

For the regions, we want to expose their purpose, and let users decide when to drill down into the numbers!

PR #2389, opened 2025-09-10
Ensure select lists are properly labeled

whabanks:
# Summary | Résumé

This PR fixes https://github.com/cds-snc/notification-planning/issues/2289.

- Add aria-labelledby to checkbox and radio button inputs which are part of select input groups
- Removed explicit non-use of aria-labelledby for select lists in the go-live checklist pages

1. On a trial services click `Request to go live` -> `Tell us about how you intend to use GC Notify`
2. Turn on VoiceOver and select one of the checkboxes
3. Note that `label`, `hint`, and associated `fieldset legend` are announced
4. Completed the first step -> on the second step select a radio button in one of the groups
5. Note the same behaviour as in step 3, minus the `hint` as there are none for these radio buttons.

### DIFF
--- a/app/templates/components/select-input.html
+++ b/app/templates/components/select-input.html
@@ -6,10 +6,12 @@
     field, hint, disable, option_hints, hide_legend, collapsible_opts, legend_style, input, is_page_heading, use_aria_labelledby, testid, optional
   ) %}
   {% for option in field %}
-    {{ select_input(option, disable, option_hints, input=input, option_conditionals=option_conditionals, testid=option.data, badge_options=badge_options, bordered=bordered) }}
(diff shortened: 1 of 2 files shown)


smcmurtry (commented):
Testing on prod with Voiceover vs testing in the review app here, there is no difference, the labels for checkboxes and radio buttons are read out correctly in both cases.

Looking at the DOM, I think you would need to add an id to the label, like `id="intended_recipients-0"` for the `aria-labelledby` to do anything. 

Image

I am skeptical about this recommendation, I think we already have this set up correctly. We are doing it the same way as W3C in these examples - they aren't using `aria-labelledby` for checkboxes or radios:
- https://www.w3.org/WAI/tutorials/forms/grouping/
- https://design-system.w3.org/styles/form-errors.html

PR #2405, opened 2025-09-16
Improve accessibility of `more` menus

whabanks:
# Summary | Résumé
This PR updates the `more menu`'s behaviour to align with the behaviour of the `disclosure menu`.
Fixes: https://github.com/cds-snc/notification-planning/issues/2388

- Expanding a more menu will now focus the first item in the list
- When expanded, the menu items are navigatable via the arrow keys
- When expended, focus is soft trapped inside the menu list when navigating via the arrow keys. i.e. after navigating `next` when on the last item, focus is returned to the first item in the menu list
- When expanded, pressing the `esc` key will now close the menu and return focus to the menu button itself
- Added an aria-label to the more menu button

1. Navigate to `/_storybook?component=more-menu`
2. Focus on the menu button
- [ ] Screen readers announce the new `More menu items` label
3. Expand the menu with `space` or `enter`
- [ ] The first menu item in the list is focused and read by the screen reader
4. Navigate the menu using the `left` and `right` arrow keys
- [ ] Navigation works and the menu items are read by the screen reader
- [ ] Hitting the **`right arrow key`** while on the **last** item in the menu brings you back to the **first** menu item in the list
- [ ] Hitting the **`left arrow key`** while on the **first** item in the menu brings you back to the **last** menu item in the list
5. Press the `esc` key
- [ ] The menu is closed
- [ ] Focus is returned to the menu button and the `More menu items` label is read out
6. Open the menu again, then press `tab`
- [ ] The menu is closed, and focus moves to the next interactable element on the screen

### DIFF
--- a/app/assets/javascripts/moreMenu.js
+++ b/app/assets/javascripts/moreMenu.js
@@ -1,7 +1,6 @@
 (function (Modules) {
   "use strict";
 
-  const registerKeyDownEscape = window.utils.registerKeyDownEscape;
   const registerKeyBasedMenuNavigation =
     window.utils.registerKeyBasedMenuNavigation;
 
--- a/app/assets/javascripts/moreMenu.js
+++ b/app/assets/javascripts/moreMenu.js
@@ -1,13 +1,24 @@
 (function (Modules) {
   "use strict";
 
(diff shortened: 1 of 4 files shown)


andrewleith (approved):
Works perfectly, nicely done!

PR #2410, opened 2025-09-18
Improve auto-focus management

whabanks:
# Summary | Résumé
This pull request refactors how autofocus is handled across the frontend codebase, removing the legacy JavaScript autofocus module and shifting autofocus logic into the HTML templates using the `autofocus` attribute. Additionally, it introduces a comprehensive example form in the Storybook view for testing validation and form controls. Several macros are updated to support autofocus when errors are present, improving accessibility and user experience.

## Related cards
- https://github.com/cds-snc/notification-planning/issues/2484

- Ensure forms throughout notify set focus automatically on the first field when there is an error (and no validation summary)
- Ensure forms with validation summaries have the summary itself being focused when there is an error
- Ensure banners receive focus when they are present


app/templates/components/banner.html
```diff
@@ -10,6 +10,7 @@
       role="status"
     {% endif %}
     tabindex="-1"
```
whabanks, line 12: The banner has [this CSS](https://github.com/cds-snc/notification-admin/blob/58f7bae2488eea8611bfa85147f8f96e96de0452/app/assets/stylesheets/tailwind/components/layout.css#L10-L11) which applies the yellow outline on focus. To me it seems out of place given the element isn't interactable. I feel like the green outline we already have is enough to grab attention. We can probably remove this class, thoughts?
image

andrewleith: What do you think, @amazingphilippe ?

amazingphilippe: If we expect the banner to receive focus, via tabindex="0" for example, we should keep this class. Given it has been -1 all along? How are we getting this outline on focus?

app/templates/components/autocomplete.html
```diff
@@ -1,19 +1,19 @@
...
     <label for="{{ field.name }}">
-      <strong>{{ field.label }}</strong>
+      <strong>{{ field.label.text }}</strong>
       {% if hint %}
         {% set hint_id = field.name + '-hint' %}
         <span id="{{ hint_id }}" class="form-hint">{{ hint }}</span>
       {% endif %}
       {% if field.errors %}
-        <span id="{{ field.name }}-error-message" class="error-message" data-module="track-error" data-error-type="{{ field.errors[0] }}" data-error-label="{{ field.name }}">
+        <span id="{{ field.name }}-error-message" class="error-message" data-module="track-error" data-error-type="{{ field.errors[0] }}" data-error-label="{{ field.name }}" tabindex="-1" autofocus>
```
amazingphilippe, line 11: What are we expecting to happen with tabindex="-1" and autofocus attributes? These seem not compatible to me...

app/templates/components/file-upload.html
```diff
@@ -32,7 +32,7 @@
   accept="",
   testid=None
 ) %}
-  <div class="file-upload-group {% if field.errors %}form-group-error{% endif %} relative inline-flex flex-col gap-2 items-start" {% if field.errors %}data-module="autofocus"{% endif %}>
+  <div class="file-upload-group {% if field.errors %}form-group-error{% endif %} relative inline-flex flex-col gap-2 items-start" {% if field.errors %}autofocus{% endif %}>
```
amazingphilippe, line 35: Have we tested or researched that the autofocus attribute works as well as the autofocus module on different browsers?

app/templates/components/radios.html
```diff
@@ -1,20 +1,20 @@
 {% from "components/select-input.html" import select, select_list, select_nested, select_input %}
 
-{% macro radios(field, hint=None, disable=[], option_hints={}, hide_legend=False, is_page_heading=False, use_aria_labelledby=True, testid=None, badge_options=None, bordered=None) %}
-  {{ select(field, hint, disable, option_hints, hide_legend=hide_legend, input="radio", is_page_heading=is_page_heading, use_aria_labelledby=use_aria_labelledby, testid=testid, badge_options=badge_options, bordered=bordered ) }}
+{% macro radios(field, hint=None, disable=[], option_hints={}, hide_legend=False, is_page_heading=False, use_aria_labelledby=True, testid=None, badge_options=None, bordered=None, autofocus=False) %}
```
amazingphilippe, line 3: Can we lean into field.errors here as well? Passing autofocus as a boolean seems a bit verbose

smcmurtry (approved):
I tested the following in the review app:
- [x] tested banners are auto-focussed
- [x] tested error summaries are auto-focussed
- [x] tested when there is no error summary, the first input with an error is auto focussed and the error is read out
- [x] tested file upload error is auto-focussed
- [x] tested autocomplete combobox for selecting department - the error is focussed rather than the select input, but that is at least an improvement on how it was working before

I also looked at the storybook pages and they look good. 

I'm not sure I managed to test every different case with different form inputs or all the autocomplete changes but I think that is bound to happen with a PR of this size. 

For next time here are some ideas for how we could split up work like this:
- PR to add the autofocus attr to the banner
- multiple PRs to add the autofocus attr to each form field, these could then each be tested more thoroughly
- PR to make changes to the autocomplete combobox
- final PR to delete the JS autofocus code

PR #2641, opened 2026-04-02
Tailwind 4 migration

amazingphilippe:
# Summary | Résumé

Upgrading to tailwind 4 and vite

- Vite is ... quick. The compilation is very efficient and takes a fraction of a second. It compiles on save, so you can start the development server and forget about `npm run tailwind` when writing css and html
- There are a few things I had to fix after the upgrade:
- `.container` is not a utility provided by tailwind by default, but we can create our own. 
- `@screen` function has been replaced by `@variant`
- Linting is more efficient, so I had to fix some invalid utilities in some `@apply` blocks
- No more tailwind.config.js - Its all a css theme, and integrates perfectly with native css properties. CS properties are available everywhere, not just using tailwind utilities. 

- Locally, `npm install`
- Maybe rebuild the container
- `npm run css` does the build, but it also runs on `make run-dev`


app/assets/stylesheets/tailwind/components/menu.css
```diff
@@ -36,7 +36,7 @@
     @apply transform rotate-180;
   }
 
-  :has > [data-module="menu"] {
+  :has(> [data-module="menu"]) {
```
amazingphilippe, line 39: It picked up this syntax error that previously went unnoticed

app/assets/stylesheets/tailwind/toolkit/wp.css
```diff
@@ -151,7 +151,7 @@
   }
 
   .page-content ol {
-    @apply list list-number pl-8;
+    @apply list-decimal pl-8;
   }
 
   .page-content blockquote {
```
amazingphilippe, line 157: Open the diff below in style.css to see tailwind 4's config

app/assets/stylesheets/tailwind/style.css
```diff
@@ -125,34 +298,26 @@ input[type="search"]::-webkit-search-decoration {
 
 section[id]:target {
   scroll-margin-top: 30px;
-  --yellow: theme("colors.yellow.DEFAULT");
-  --white: white;
   box-shadow:
-    0 0 0 15px var(--white),
-    0 0 0 20px var(--yellow);
+    0 0 0 15px var(--color-white),
```
amazingphilippe, line 302: An example of using tailwind tokens as regular CSS properties

PR #2654, opened 2026-04-13
Add new guide to sitemap

amazingphilippe:
# Summary | Résumé

Adds `counting_text_messages` to the sitemap

We can add a parameter `show` to our links for when we want to link it to Feature Flags. We can remove it when we remove all the other flags and the link will behave like the others. When `show` is not defined, the link is displayed by default. 

- [x] Navigate to sitemap
- [x] Notice the new guide is listed, and in the right position, alphabetically
- [x] Test the link
- [x] Go back to the sitemap, check the FR sitemap
- [x] Check new link position. Should be alphabetical
- [x] Test the link


whabanks (approved):
✅  Links are present and work in FR/EN
✅  Alphabetical order
🚀 LGTM

PR #2658, opened 2026-04-14
A11y/tiptap button labels

amazingphilippe:
# Summary | Résumé

This pull request refactors the rich text editor's toolbar and tooltips to enhance accessibility. It centralizes all editor UI labels and keyboard shortcuts (with platform-specific variants) into a new localization module, replaces ad-hoc label definitions with context-based translation, and updates the Cypress tests to accommodate these changes. Additionally, it improves tooltip accessibility and updates the test suite to check for both English and French UI labels.

- [x] Label toggles: Verify each toolbar button shows either "Apply {Label}" or "Remove {Label}" 
- [ ] `aria-pressed`: Confirm `aria-pressed` is `false` → `true` → `false` when applying/removing formatting for every button.
- [x] Live region announcements: After applying, live region contains "applied"; after removing, contains "removed".
- [x] Tooltips: Hover each toolbar button and assert the tooltip box appears and the tooltip label is not empty.
- [x] Toolbar keyboard focus: Press `Alt+F10` to focus the toolbar; assert initial focus (H1) and that arrow keys move focus forwards/backwards and restore remembered position.
- [x] Shortcut toggles: For each keyboard shortcut (e.g., `Mod+Alt+1` for Heading, `Mod+b` for Bold), verify the corresponding button toggles aria-pressed.
- [x] Link modal: Press `Mod+K` to open the link modal; verify it is visible and that Escape closes it.


app/assets/javascripts/tiptap/MenuBar.js
```diff
@@ -214,10 +215,43 @@ const AccessibleToolbar = ({ children, label, editor, className = "" }) => {
...
+    <TooltipWrapper label={label} shortcut={shortcut}>
+      <button
+        type="button"
+        data-testid={testId}
+        onClick={onClick}
+        onMouseDown={onMouseDown}
+        disabled={isDisabled}
+        className={"toolbar-button" + (isActive ? " is-active" : "")}
+        title={label}
+        aria-pressed={testId === "rte-horizontal_rule" ? undefined : isActive}
```
andrewleith: Why the specific check for this here?

app/assets/javascripts/tiptap/MenuBar.js
```diff
@@ -214,10 +215,43 @@ const AccessibleToolbar = ({ children, label, editor, className = "" }) => {
...
+const ToolbarButton = ({
+  testId,
+  onClick,
+  onMouseDown,
+  isActive = false,
+  isDisabled = false,
+  labels,
+  children,
+}) => {
+  const isStringLabel = typeof labels === "string";
```
andrewleith: Maybe you should comment what this means - what is the scenario when a label isnt a string?

andrewleith (commented):
A couple of comments below.

app/assets/javascripts/tiptap/MenuBar.js
amazingphilippe: +1 shouldn't be so specific.

app/assets/javascripts/tiptap/MenuBar.js
amazingphilippe: Replaced with 3 simpler to read definitions

andrewleith (approved):
LGTM!

PR #2661, opened 2026-04-15
Fix/tables after node 20

amazingphilippe:
# Summary | Résumé

Upgrading to node 20 seems to have broken our fragile table component. 

I think I narrowed it down to some dodgy javascript, where we were trying to apply unitless values to css properties. We were also using absolute positioning in an element which is normally part of the content flow. 

Its still not great, we can see some glitches, but at least we're not covering parts of the page. 

Before:
Capture d’écran, le 2026-04-15 à 15 52 16

After:
Capture d’écran, le 2026-04-15 à 15 52 46

From the processed html in the browser, this was the only diff:
Capture d’écran, le 2026-04-15 à 15 37 53

- [x] Storybook table
- [x] When breaking variable names in a template (change your variable names)
- [x] When sending bulk (I did not notice anything weird in prod, but I know it was broken for some people)


amazingphilippe:
Also fixes the "These messages exceed your daily limit" Screen. Assuming it might be a good global hot fix for now:

Before:
Capture d’écran, le 2026-04-15 à 16 08 13

After:
Capture d’écran, le 2026-04-15 à 16 08 02

jzbahrai (approved):

PR #2663, opened 2026-04-15
[Security] - Experiment migrating to fido2 2.2.0

whabanks:
# Summary | Résumé
We need to [bump `cryptography` in utils](https://github.com/cds-snc/notification-utils/pull/353) to patch a vulnerability. However, this also requires migrating to `fido2 >= 2.2.0` in API, as `fido2` transitively pulls in `cryptography` and `2.2.0` is the latest version that includes a patched `cryptography` version.

In `fido2 >= 2.0.0` `CBOR` encoding has been dropped in favour of `JSON` encoding. Changing the data shape and endpoint contract between API and Admin. This PR aims to update the contract to be compatible with the [API change made here](https://github.com/cds-snc/notification-api/pull/2836).

**Key changes:**
* Removed all `CBOR` encode/decode usage in `user_api_client` and `user_profile`
* Switch from `CBOR` encode/decode to JSON and build a standard `WebAuthn` response in `fido2.js` when building registration and authentication responses.

1. Run the app locally, and add a security key to your account if you are not already using one
1. Check out this branch and the [accompanying API branch](https://github.com/cds-snc/notification-api/pull/2836)
    - Make sure to run `poetry install` in API and `npm run tailwind` in admin before continuing
3. Log into notify with an existing security key to ensure backwards compatibility
- [x] Login should work correctly
4. Log out, then log back in but use a different security key, one that won't pass validation (physical vs 1pass for example)
- [x] Shouldn't be able to log in
5. Log back in and remove your security key
6. Log out and in again
- [x] Note you received the email code prompt as usual
7. Finish logging in and add a security key once again
8. Log out, log in using the new key
- [x] Login works


jzbahrai (approved):

whabanks:
Cypress tests will fail here until https://github.com/cds-snc/notification-api/pull/2836 is merged

PR #2664, opened 2026-04-16
Use non-regional en and fr attributes

amazingphilippe:
# Summary | Résumé

One Fable participant wasn't able to trigger the FR voice when reading FR content block. 

We use regional language codes, such as en-CA and fr-CA. We're unsure if this might be the problem. We might not need to be so specific with the language codes to increase the likeliness that users will have any fr and en language packs installed, rather than specifically Canadian packs. This might also be completely irrelevant, but we need to at least try


andrewleith (approved):
I am skeptical that this will make a difference.  I have only ever used `en-CA` (and `fr-CA`)!

amazingphilippe:
> I am skeptical

Same: Lets test it out before we merge

andrewleith:
> > I am skeptical
> 
> Same: Lets test it out before we merge

Testing with chrome+voice over and edge+nvda: neither read in the language for `lang="fr"` nor `lang="fr-CA"`.  I think this might have to do with it being a contenteditable?

andrewleith:
Although I can't find any conclusive sources, AI seems to also think lang tags won't work as we expected in a `contenteditable`:
> `lang` is present in your TipTap nodes, but many screen readers do not reliably switch voice/language inside a live editing surface.

> Why this happens:
> 
> Screen readers often treat a [contenteditable](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) region as one interactive text control, and keep one speech language/voice for the whole control.
> Nested changes are better supported in static reading mode than while editing/caret-browsing.

So, we are correctly marking language changes with the `lang` attribute, but we should not assume perfect voice switching in every screen reader/browser mode. [W3C’s Understanding SC 3.1.2 notes](https://www.w3.org/WAI/WCAG21/Understanding/language-of-parts#intent) that in some modes, language changes may not be exposed to assistive technologies, so pronunciation switching can fail due to UA/AT limitations (not authoring failure).

amazingphilippe:
Makes sense. Lets not merge this

PR #2666, opened 2026-04-16
Remove duplicate ID in footer

amazingphilippe:
# Summary | Résumé

We have a duplicate ID in our footer

- [ ] Run a11y scan on any page
- [ ] Don't find the duplicate ID in the footer


andrewleith (approved):
LGTM.  The cypress test dont check the header and the footer for reasons I can't remember (it only checks `<main>`) so that is why this was never found!  We may want to revisit that!

PR #2669, opened 2026-04-20
A11y/rte describe editable blocks

amazingphilippe:
# Summary | Résumé

## Announcing blocks

This PR improves how editable blocks are announced. Until now, only some blocks would be announced, and it was based on their `role` such as `heading` or `list` This cause 2 problems
- You couldn't tell what type of list you were editing
- For nodes without a role, like bold or italics, nothing would be announced

This isn't like reading content, this is about editing content so more verbose-ness is needed to help users confirm formats are applied or not. 

We're adding a live region, similar to the toolbar live region to announce what formats are applied at any given cursor position. This includes nested formats, so that you can get a sense of where things are with more confidence. 

There will be items that are announced twice, where some screen readers already announce headings and bockquotes, and lists, etc. For example, in NVDA on Windows, a heading may be read: 

```
heading level 1 The actual content
Heading
```

Where 
- "heading level 1" is what NVDA volunteers to announce
- "The actual content" is the content of the heading
- "Heading" is our way to identify what type of heading. We use "Heading" and Subheading" So there is some benefit to announcing the block by the name we gave its toolbar button too. 

## Toolbar feedback when focus moves back to editor

This PR also adds the toolbar feedback as a editor textbox description, so that applied styles are also announced when the toolbar focus is lost.

We can test with NVDA on Edge and Firefox, We can test with Voiceover. 
Ultimately we will need to book a test with some real users to see how this solution works out, and how it can be improved.


app/assets/javascripts/tiptap/AnnouncerPlugin.js
```diff
@@ -0,0 +1,91 @@
+import { Plugin, PluginKey } from "@tiptap/pm/state";
+import { Extension } from "@tiptap/core";
+
+/**
+ * A TipTap extension that provides a ProseMirror plugin to track the active node/mark stack
+ * and announce it to screen readers via an ARIA live region.
+ */
+export const AnnouncerPlugin = (t) => {
```
andrewleith: This is a neat idea.  I wonder if it will interfere with the alert the menu bar uses? Should we combine these?

amazingphilippe: I can test this, but I should not interfere since we can never focus the toolbar and the editor at the same time!

amazingphilippe: As far as I can see, this does not interfere with the toolbar live region. However, the toolbar's live region is already competing with the editor itself. Should I try to fix it in this PR?

andrewleith: Oh good point about them firing at different times

> Should I try to fix it in this PR?

Yeah maybe? I think with what you did here, if we can reliably get AT users to get notified when formatting is added and removed this would all be extremely solid.

amazingphilippe: I think I managed to create a better experience. We have two live regions, but the one in the toolbar is mostly useless since we move the focus to the textbox. 

To make it useful, I've set the textbox description to refer to the toolbar's live region. This means that for screen readers who begin to read the entire textbox, they will start by announcing the latest format applied, before reading the textbox content. 

After a bit of a delay, we also clear the toolbar live region to keep the description fresh. 

For voiceover, this is reasonably nice. Next, we can test this on NVDA to see what happens there.

app/assets/javascripts/tiptap/localization.js
```diff
@@ -95,6 +95,23 @@ export const translations = {
...
+      horizontalRule: "Section break",
+      roleSeparator: "Separator",
+      blockquote: "Blockquote",
+      variable: "Variable",
+      englishBlock: "English content",
+      frenchBlock: "French content",
+      conditionalBlock: "Conditional section",
+      conditionalInline: "Conditional text",
+      rtlText: "Right-to-left text",
+    },
```
andrewleith: There are a lot of duplicated strings here which may lead to some maintenance issues. Should we consider trying to consolidate these?

app/assets/javascripts/tiptap/CustomComponents/LanguageNode.js
```diff
@@ -51,6 +51,7 @@ const createLanguageNode = (language, langCode) => {
           lang: langCode,
           "data-lang": language.toLowerCase(),
           "data-type": `${language.toLowerCase()}-block`,
+          role: "region",
```
andrewleith, line 54: Why `region` here and `group` for the conditional nodes?  What is the difference?

app/assets/javascripts/tiptap/AnnouncerPlugin.js
```diff
@@ -0,0 +1,96 @@
...
+            return null;
+          },
+        }),
+      ];
+    },
+  });
+};
+
+function updateAnnouncer(text) {
+  const announcer = document.getElementById("editor-announcer");
```
andrewleith: This makes it so we can only ever have one editor on a page.  So far that may be the case, but the rest of the code was written so that it would be safe to have multiple.  Also, consider not using `document.getElementById` in react code and instead use `refs`.

app/assets/javascripts/tiptap/MenuBar.js
```diff
@@ -491,9 +482,11 @@ const MenuBar = ({
       >
         <div
           className="sr-only"
-          role="alert"
+          role="status"
```
andrewleith, line 486: Is `role="status"` what made this work so much better with this implementation? Nice job!

andrewleith (commented):
Really nice work here!  I am very impressed you got it to behave so cleanly, and even in Voiceover!  Also, the french block reads in a french accent in NVDA (although maybe it did that before too not sure I tested with NVDA)

**Voiceover testing:**
For the toolbar live-region:
- when a variable is inserted (i.e. no text selected and you press the variable button), it says "Variable removed"
- when a section break is inserted, it says "section break removed"

All the other buttons seemed to work correctly.

Moving around with the cursor:
- When I am focused on the separator it no longer says separator, it reads the subsequent line of text
- Voiceover sometimes reads the node type as you enter them, but it is very hit and miss (I think we expected this?)

---

**NVDA testing**:
Toolbar:
- every time you press a toolbar button it says "editor toolbar, toolbar" before it gives relevant info - EXCEPT for conditionals - for some reason it doesn't do it for those 2
- when a section break is inserted, it says "section break removed"
- when a variable is inserted (i.e. no text selected and you press the variable button), it says "Variable removed"

Moving around with the cursor:
- When I am focused on the separator it no longer says separator, it reads the subsequent line of text
- "bold" and "italics" are only announced on the second letter of a word and only if you pause for a second
- Bulleted lists are announced specifically but numbered lists are not they are only referred to as "list"
- When i cursor into a variable, it is not announced
- Exiting EN/FR blocks is not announced

app/assets/javascripts/tiptap/CustomComponents/LanguageNode.js
amazingphilippe: the language block only contains content, but the conditionals are a group of inputs.

app/assets/javascripts/tiptap/MenuBar.js
amazingphilippe: alert makes it assertive, and since we don't rely on it to be assertive anymore, we can make the role status. Not technically part of the solution, but we're cleaning up a solution that did not work as expected

amazingphilippe:
1. The voiceover bugs go a bit deeper than we think. To me its all related to how we compute "isActive". It seems a bit inconsistent for each format. For example, place your cursor before a bolded format, and the toolbar button won't show the style as active. Maybe it should? If we fix that, I think the announcer will pick it up.
2. Same with the announcer for the variable, after inserting a variable, we place the cursor at the end of the variable, at a spot where we consider the variable not active. Same with the separator: after inserting it, we also add an exttra line and place the cursor below the hr. 
3. There's something else wrong with the separator, I think it would be safer to fix it in its own PR. Maybe the hr needs a label? Or some hidden content? SRs seem to want to say something, and they probably pick something random because we don't give them something clear to say. 
4. In NVDA, I confirm it's reading our label, then the semantic role. We could drop the "toolbar" from our own label, and only label it as "Editor" or better: "Formats" Lets also do it in a clean PR
5. Lets explore announcing block exits in a different PR. Lets also confirm the UX.

app/assets/javascripts/tiptap/localization.js
amazingphilippe: Agreed, fixed

app/assets/javascripts/tiptap/AnnouncerPlugin.js
amazingphilippe: Fixed

andrewleith:
Hey @amazingphilippe - I've added a suite of tests to ensure the live region is updated as expected.  If all looks good there I think we can merge!

andrewleith (approved):
Nice work!!

PR #2671, opened 2026-04-20
Add support for the one-click unsubscribe header

smcmurtry:
# Summary | Résumé

Part of https://github.com/cds-snc/notification-planning/issues/3096

This PR does the following:
- adds 2 new variables to config.py: `ONE_CLICK_UNSUB_SERVICE_IDS` to allowlist a comma separated list of  service IDs in prod, and `ONE_CLICK_UNSUB_ALL_SERVICES`, which is a boolean that I'm using to turn this feature on in our staging/dev environments
- adds code to detect if an allowed service has included a personalisation variable called `unsubscribe_url` in their email template, and displays the checkbox in the screenshot below
- passes the new `use_custom_unsubscribe_url` field to the api
- adds tests

Screenshot 2026-04-21 at 4 39 40 PM

1. Check out this branch and run locally
2. Check out the api branch from https://github.com/cds-snc/notification-api/pull/2841 and run locally
3. Create a new email template and add a personalization variable called "unsubscribe_url" to your template
4. Verify that you see the "Use ((unsubscribe_url)) for one-click unsubscribe" checkbox
5. Click the checkbox and save the template
6. Send yourself an email with that template, using "https://google.com" for the `unsubscribe_url`
7. Open the email in gmail
8. In Gmail, click: "3-dot menu" > "Show Original" to view the raw email
9. Verify the following headers are present:
```
List-Unsubscribe: <https://google.com>
List-Unsubscribe-Post: List-Unsubscribe=One-Click
```


jzbahrai:
tested and it worked!

To: jumana.bahrainwala@cds-snc.ca
List-Unsubscribe: <https://google.com>
List-Unsubscribe-Post: List-Unsubscribe=One-Click

app/config.py
```diff
@@ -192,6 +196,7 @@ class Development(Config):
     DEBUG = True
     DEBUG_KEY = "debug"
     MOU_BUCKET_NAME = "notify.tools-mou"
+    ONE_CLICK_UNSUB_ALL_SERVICES = True
```
jzbahrai, line 199: i dont love two configs for the same feature.. but using the other one for both usecases will be business logic based.

jzbahrai (approved):

jzbahrai:
also checked the db the link gets set

andrewleith:
Why doesn't the unsubscribe button show up in gmail?  Is this configured correctly?  It should look like this:
image

app/templates/views/edit-email-template.html
```diff
@@ -123,6 +134,17 @@ <h2 class="heading-medium">{{ _('Template category') }}</h2>
...
+  (function () {
+    var setting = document.getElementById('custom-unsub-url-setting');
+    if (!setting) return;
+    var contentInput = document.getElementById('template_content');
+    if (!contentInput) return;
+    contentInput.addEventListener('change', function () {
+      setting.classList.toggle('hidden', !contentInput.value.includes('((unsubscribe_url))'));
+    });
+  })();
+
```
andrewleith: Should this inline script be moved into a `.js` file?

app/templates/views/edit-email-template.html
```diff
@@ -95,6 +95,17 @@ <h2 class="heading-medium">{{ _('Template category') }}</h2>
       {% if current_user.platform_admin %}
         {{ radios(form.process_type, hint=_('This is only manageable by platform admins'), use_aria_labelledby=false) }}
       {% endif %}
+
+      {% if one_click_unsub_enabled and form.use_custom_unsubscribe_url is defined %}
+        <div id="custom-unsub-url-setting" class="form-group mb-8 {{ 'hidden' if '((unsubscribe_url))' not in (form.template_content.data or '') else '' }}">
```
andrewleith, line 100: I found this particularly confusing.  I had enabled the feature in `config.py` and in `.env` so I expected to see it when I edited my template, but it only appeared once I had input the reserved variable name into the template.

andrewleith (commented):
This works overall but I did find it confusing to have to enter a special reserved variable name in order to turn the feature on and see the checkbox.

smcmurtry:
> Why doesn't the unsubscribe button show up in gmail? Is this configured correctly? It should look like this: image

Whether or not Gmail displays the unsubscribe button is up to their discretion and related to some reputation logic on their end that they do not disclose. Maybe Gmail probes the url before displaying the button? I think we won't be able to see the button until we test with a real url in prod. Here's their official documentation, but it is light on details: https://support.google.com/a/answer/81126?visit_id=639124641691966526-2203493287&rd=1#subscriptions

app/templates/views/edit-email-template.html
smcmurtry: done!

PR #2676, opened 2026-04-24
A11y/fix rte link modal

andrewleith:
# Summary | Résumé

This pull request simplifies and improves the link editing experience in the Tiptap editor. The main focus is on cleaning up the logic for opening the link modal and updating link marks, removing unnecessary state and event listeners, and making the codebase easier to maintain.

**Link editing logic improvements:**

* The logic for updating a link when the cursor is inside an existing link is improved: now, if the selection is collapsed and inside a link, the link's `href` is updated across its full range instead of inserting new content. If the cursor is not inside a link, a markdown link is inserted as before.

- [ ] Clicking on a link in the editor should no longer pop the modal automatically
- [ ] Using the keyboard in the editor, navigating on to a link should no longer pop the modal automatically
- [ ] After placing the cursor within a link and then opening the link modal, changing the url should change it for the whole link (and not insert a new link at the cursor)

## Related cards
* https://github.com/cds-snc/notification-planning/issues/3111


amazingphilippe (approved):
LGTM: Tested in Firefox and works as expected.

PR #2678, opened 2026-04-27
Fix the validation used on the "Service settings > Change service name" page

smcmurtry:
# Summary | Résumé

Fix for https://github.com/cds-snc/notification-planning/issues/3154

We are using the correct validation for the service name on Step 2 of the "Create a new Service" flow, but the same validation was not applied when a user edits the name of an existing service from the Service Settings page. This PR fixes that.

1. Log into the review app
2. Visit the Settings page for one of your existing services
3. Try to change the name of your service to the following (less than 255 chars):
```
Message de Jimmy Royééééééééé – Gestionnaire principal responsable des urgences et des évacuations du BIN // Message from Jimmy Royééééééééé – NPB Lead senior manager for emergencies and évacuations
```
4. observe that you see the following error message:
Screenshot 2026-04-27 at 4 26 57 PM

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -570,6 +570,7 @@ class RenameServiceForm(StripWhitespaceForm):
         validators=[
             DataRequired(message=_l("This cannot be empty")),
             validate_service_name,
+            validate_combined_email_header_length,
         ],
     )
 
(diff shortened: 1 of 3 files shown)


andrewleith (approved):
LGTM

PR #2681, opened 2026-04-30
Update the free sms parts limit per fiscal year to be 100k

smcmurtry:
# Summary | Résumé

Update the free sms parts limit per fiscal year to be 100k. This is what the free limit is set to when a new service is created (depending on the service's organisation type).

Part of https://github.com/cds-snc/notification-planning/issues/3158

1. log into the review app (with your platform admin account)
2. create a new service
3. go to the service settings page
4. verify that "Free text messages per fiscal year" is 100,000 
Screenshot 2026-04-30 at 1 34 37 PM

### DIFF
--- a/app/config.py
+++ b/app/config.py
@@ -66,14 +66,14 @@ class Config(object):
     DEBUG = False
     DEBUG_KEY = os.environ.get("DEBUG_KEY", "")
     DEFAULT_FREE_SMS_FRAGMENT_LIMITS = {
-        "central": 25_000,
-        "local": 25_000,
-        "nhs_central": 250_000,
-        "nhs_local": 25_000,
-        "nhs_gp": 25_000,
-        "emergency_service": 25_000,
-        "school_or_college": 25_000,
-        "other": 25_000,
+        "central": 100_000,
+        "local": 100_000,
+        "nhs_central": 100_000,
+        "nhs_local": 100_000,
+        "nhs_gp": 100_000,
+        "emergency_service": 100_000,
+        "school_or_college": 100_000,
+        "other": 100_000,
     }
     DEFAULT_LIVE_SERVICE_LIMIT = env.int("DEFAULT_LIVE_SERVICE_LIMIT", 10_000)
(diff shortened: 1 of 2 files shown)


app/config.py
```diff
@@ -66,14 +66,14 @@ class Config(object):
...
-        "school_or_college": 25_000,
-        "other": 25_000,
+        "central": 100_000,
+        "local": 100_000,
+        "nhs_central": 100_000,
+        "nhs_local": 100_000,
+        "nhs_gp": 100_000,
+        "emergency_service": 100_000,
+        "school_or_college": 100_000,
+        "other": 100_000,
```
andrewleith, line 76: I'm not sure that we can use this for determining a service's free limit. I think we use this to set the service's entire annual limit.

Although looking at API, when a service is created, we may actually take the value [from here](https://github.com/cds-snc/notification-api/blob/main/app/models.py#L70).

Either way, we would need a column representing what portion of their annual send is free, and I don't think this is it.

We have another table in the DB that we may or may not use called `annual_billing` which appears at first glance to record the free portions of each service by fiscal year.

andrewleith: I may be mixing this PR up with your comment about the billing dashboard earlier - perhaps you didn't intend to try to use this for the billing dashboard!

smcmurtry: > Although looking at API, when a service is created, we may actually take the value [from here](https://github.com/cds-snc/notification-api/blob/main/app/models.py#L70).

That one does not appear to be used

> Either way, we would need a column representing what portion of their annual send is free, and I don't think this is it.

That is the `free_sms_fragment_limit` on the `annual_billing` table in the db. The values I'm modifying here are used to populate that column for new services.

When you create a new service on staging, and choose any GC organisation, the "Free text messages per fiscal year" on the settings page is 25k.

After this code change, you will see 100k. Try it out!

jzbahrai (approved):
tested and checked! we need another PR where if a service has their limit set to 25000 we need to update to 100k (i already did it once), but for the new services that were set up since

PR #2686, opened 2026-05-05
Separate react code into reusable bundle

andrewleith:
# Summary | Résumé
This pull request reorganizes how JavaScript assets are loaded and bundled, with a focus on optimizing the inclusion of React and related libraries. The main focus is extracting React into its own chunk that can be shared between any react components we use (i.e. scheduler and rich text editor right now) without loading duplicate code like we are currently doing.

- [x] Editor works in storybook
- [x] Editor works on template editor page
- [x] Scheduler works in storybook
- [x] Scheduler works on notification send page


webpack.config.js
```diff
@@ -28,6 +28,19 @@ module.exports = {
     filename: "javascripts/[name].min.js",
     path: path.resolve(__dirname, "app/assets")
   },
+  optimization: {
+    splitChunks: {
+      cacheGroups: {
```
whabanks, line 33: I like this, letting the browser cache the react js is great. I wonder if this would be a useful technique to leverage in the future. We could identify JS that changes infrequently, split it and let it remain cached for longer to speed up page load times. Likely just micro-optimizations but just a thought.

whabanks (approved):
LGTM!

PR #2689, opened 2026-05-06
Add manage_templates permission option to API key creation

jzbahrai:
## Summary
Adds the ability to assign a manage_templates permission to API keys during creation. This permission will allow API keys to create, update, or delete templates via the API (enforcement coming in a follow-up API PR).

## Changes
Feature flag: FF_ADD_TEMPLATE_PERM — enabled in dev, staging, and test; disabled in production
Form: Added manage_templates checkbox field to CreateKeyForm
API client: create_api_key() now accepts and passes a permissions list to the API
Template: Added "Add permission for above API key" heading + checkbox on the create key page (gated by FF)
Translations: Added French translations for all new strings
Tests: Form validation, view integration, and API client tests in test_create_api_key_form.py

### Test 1 
1. Set FF_ADD_TEMPLATE_PERM=True in .env
2. Navigate to API Integration → Create an API key
3. Verify the "Add permission for above API key" heading and checkbox appear below the key type radios
4. Create a key with the checkbox checked and confirm permissions: ["manage_templates"] is sent to the 
5. Check the DB, under permissions in the api key table, you will see {manage_templates}

You should see the following:

Screenshot 2026-05-06 at 12 06 06 PM

Screenshot 2026-05-06 at 12 08 53 PM

##$ test 2
1. Set FF to false 
2. You should still be able to create a key
3. No manage_template in summary
4. No manage_template in permissions


app/main/views/api_keys.py
```diff
@@ -109,6 +114,7 @@ def create_api_key(service_id):
         form=form,
         disabled_options=disabled_options,
         option_hints=option_hints,
+        show_manage_templates=current_user.has_permissions("manage_api_keys"),
```
whabanks: Let's add the FF check here `and current_app.config["FF_ADD_TEMPLATE_PERM"]` + current_app import

whabanks (changes requested):
Looks good, just need a FF check to hide the options in the UI.

whabanks (approved):

PR #2695, opened 2026-05-19
Disable Slack notifications in workflow

jzbahrai:
Comment out Slack notification steps in Cypress staging workflow.


andrewleith (approved):
LGTM

PR #2712, opened 2026-06-24
Add metrics around the duplicate recipients warning feature

smcmurtry:
# Summary | Résumé

This PR adds two `INFO` log lines to the bulk-send flow:

| When | Log message |
|---|---|
| User uploads a CSV, gets the yellow "duplicates" warning on the review screen | `bulk_send.duplicate_recipients_warning_shown` |
| User clicks **Send** anyway (without fixing the file) | `bulk_send.duplicate_recipients_sent_anyway` |

Each log line carries the `upload_id` and `count_of_duplicate_recipients`.

The purpose of these logs are so we can answer this question:
> "Of the X bulk sends per month that contained duplicate recipients, Y% of senders noticed the warning and fixed their CSV before sending — preventing Z duplicate messages from reaching Canadians."

Imagine a month's worth of these logs in a table:

| event | count_of_duplicate_recipients |
|---|---|
| `warning_shown` | 12 |
| `warning_shown` | 3 |
| `sent_anyway` | 3 |
| `warning_shown` | 50 |
| `warning_shown` | 7 |
| `sent_anyway` | 7 |

Then:

- **X = count of `warning_shown`** → 4 bulk sends had duplicates.
- **`sent_anyway` count** → 2 senders ignored the warning. So 2 out of 4 senders *fixed* the file. **Y% = 50%**.
- **Z = sum of `count_of_duplicate_recipients` on `warning_shown`, minus the same sum on `sent_anyway`** → (12+3+50+7) − (3+7) = **62 duplicate messages prevented**.

I also added the updated en/fr content into this pr.

## Follow up

Once these logs are added, we will be able to get the metrics with the following cloudwatch query:
```
fields @timestamp,
       message,
       upload_id,
       count_of_duplicate_recipients,
       (message = "bulk_send.duplicate_recipients_warning_shown") as is_warning,
       (message = "bulk_send.duplicate_recipients_sent_anyway")  as is_sent_anyway
| filter is_warning or is_sent_anyway
| stats sum(is_warning)                                   as bulk_sends_with_duplicates,
        sum(is_sent_anyway)                               as proceeded_anyway,
        sum(is_warning      * count_of_duplicate_recipients) as duplicates_detected,
        sum(is_sent_anyway  * count_of_duplicate_recipients) as duplicates_sent_anyway
```

1. Run the branch locally
2. upload a csv with duplicates
3. download the list of duplicates
4. remove some and re-upload
5. send the job
6. verify the new logs are printed out in your terminal


app/main/views/send.py
```diff
@@ -988,6 +1008,25 @@ def check_notification_preview(service_id, template_id, filetype):
...
+    count_of_duplicate_recipients = int(upload_metadata.get("count_of_duplicate_recipients", 0) or 0)
+    if count_of_duplicate_recipients:
+        current_app.logger.info(
+            "bulk_send.duplicate_recipients_sent_anyway",
+            extra={
+                "service_id": str(service_id),
+                "template_id": upload_metadata.get("template_id"),
+                "upload_id": upload_id,
+                "notification_count": int(upload_metadata.get("notification_count", 0) or 0),
+                "count_of_duplicate_recipients": count_of_duplicate_recipients,
```
jzbahrai, line 1025: do we need to pass all this? all we care about is that they sent anyways, how many duplicates/ notification_count - im unsure if they are necessary. I would just keep service_id, template_id and upload_id

smcmurtry: we need count_of_duplicate_recipients so that we can fill in `Z` the following statement:
> Of the X bulk sends per month that contained duplicate recipients, Y% of senders noticed the warning and fixed their CSV before sending — preventing Z duplicate messages from reaching Canadians

The PR description explains it more with an example - basically we are going to subtract the sum of counts at send time from the sum of counts at warning time

smcmurtry: I could remove `notification_count` but it might be useful for debugging if we think the counts are off.

jzbahrai (approved):

PR #2713, opened 2026-06-24
Otel span correlation

ben851:
# Summary | Résumé

The client side otel telemetry was reporting but wasn't being correlated with the backend requests because the traceparent id was not being sent. This is now being sent correctly and I've verified that the traces are being correlated now. 

image

ALSO including a dockerfile.local because the regular dockerfile will not build on my M3.

Merge to staging, verify staging spans are grouped.

### DIFF
--- a/ci/Dockerfile.local
+++ b/ci/Dockerfile.local
@@ -1,0 +1,51 @@
+# Node.js build stage — Node 22 on Alpine (musl) so Rolldown's musl native
+# binding resolves correctly. npm install re-resolves platform optional deps.
+FROM node:22.22.3-alpine AS node-build
+WORKDIR /app
+COPY package.json package-lock.json ./
+RUN npm install
+COPY . .
+RUN npm run build
+
+FROM python:3.12.7-slim-bookworm
+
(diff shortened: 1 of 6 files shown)


ci/Dockerfile.local
```diff
@@ -0,0 +1,51 @@
...
+RUN make babel
+
+RUN make generate-version-file
+
+ENV PORT=6012
+
+ARG GIT_SHA
+ENV GIT_SHA=${GIT_SHA}
+
+CMD ["sh", "-c", "gunicorn -c gunicorn_config.py application"]
```
jimleroyer, line 51: What's this dockerfile for? To launch the app locally without the vscode devcontainer?

jimleroyer (approved):

ci/Dockerfile.local
ben851: When I'm testing this, I build the docker container and push it to dev to run. I don't want to have to do a PR every time I make a change to the app, so I build and push locally. The default dockerfile hangs on installing zope-interface when using my m4. I don't know why, could not find a reason or a solution other than changing from alpine to debian-slim.

PR #2714, opened 2026-06-25
Otel proxy protection

ben851:
# Summary | Résumé

### Protect the OTEL proxy endpoint on the admin:

1. Unauthenticated requests (i.e. before login) has a signed token with a 5 minute TTL. After 5 minutes we won't get the traces but the app will still work. We can play with this number if we want. We need to strike a balance between not allowing malicious users to hammer the system with unauthenticated requests vs the value of having metrics on people bouncing around the different unauthenticated pages (features, by and for the gc etc).

2. Once the user is authenticated via their GC Sign in, the trace authentication is handled by the session. As long as your session is valid, we will see traces. 

### This PR also adds page paths to the traces so that we can see what pages the client was hitting.

In staging, bounce around unauthenticated pages, check in signoz for the traces. Then sign in, and bounce around authenticated pages (dashboards etc) and check for those traces in signoz.


app/main/views/otlp_proxy.py
```diff
@@ -21,6 +24,21 @@ def otlp_proxy(signal_type):
     if not current_app.config.get("ENABLE_CLIENT_SIDE_OTEL", False):
         abort(404)
 
+    if current_user.is_authenticated:
+        token = request.headers.get("X-CSRFToken", "")
+        try:
+            validate_csrf(token)
```
jimleroyer, line 32: oh that's cool

app/__init__.py
```diff
@@ -320,6 +320,15 @@ def inject_global_template_variables():
...
         otel_traceparent = _get_current_traceparent() if otel_enabled else ""
+        otel_auth_token = ""
+        otel_auth_mode = ""
+        if otel_enabled:
+            if current_user.is_authenticated:
+                otel_auth_token = generate_csrf()
+                otel_auth_mode = "csrf"
+            else:
+                otel_auth_token = URLSafeTimedSerializer(application.config["SECRET_KEY"]).dumps(nonce or "", salt="otel-proxy")
+                otel_auth_mode = "signed"
```
jimleroyer: This would be nice in a function of its own, for configuration otel related stuff and keep the function lighter here.

app/config.py
```diff
@@ -94,6 +94,7 @@ class Config(object):
     FF_ADD_TEMPLATE_PERM = env.bool("FF_ADD_TEMPLATE_PERM", False)
     FF_FILE_ATTACHMENTS = env.bool("FF_FILE_ATTACHMENTS", False)
     ENABLE_CLIENT_SIDE_OTEL = env.bool("ENABLE_CLIENT_SIDE_OTEL", False)
+    OTEL_CLIENT_SIDE_TOKEN_TTL = env.int("OTEL_CLIENT_SIDE_TOKEN_TTL", 5 * 60)  # seconds, default 5 minutes
```
jimleroyer, line 99: These two above shouldn't be part of the feature flags section, we could make a section for OTEL.

jimleroyer (approved):

app/config.py
ben851: Changed as per request

app/__init__.py
ben851: Moved to its own thing as per request

PR #2717, opened 2026-07-06
fix mounting issue

jzbahrai:
## Bug:

https://cds-snc.freshdesk.com/a/tickets/29015
A client reported the tiptap editor wasnt loading

## Root cause
In TipTap 3 , editor.view is a getter that returns a Proxy stub — not undefined — until the underlying ProseMirror view is actually mounted. Accessing most properties on that stub (like dom, coordsAtPos, posAtDOM) throws:

[tiptap error]: The editor view is not available. Cannot access view['dom']. The editor may not be mounted yet.

Crucially, optional chaining like editor?.view?.dom does not protect against this, because editor?.view succeeds (returns the truthy Proxy) and it's the .dom access on the stub that throws.

The specific trigger: in SimpleEditor.js:403 the RTE (EditorContent) is only rendered in RTE mode. When a user's preferred editor is markdown (initialMode="markdown"), EditorContent is never mounted, so the view never exists. An effect then ran if (!editor || !editor.view || !editor.view.dom) return; — the editor.view.dom check itself threw instead of returning early. This is timing/config dependent, which is why it surfaced intermittently (e.g. Firefox on Windows).

## Fix
In SimpleEditor.js:

1. Added a module-level helper isEditorViewMounted(editor) that safely detects whether the real ProseMirror view is mounted (wrapping the access in try/catch so the throwing getter can't leak).
2. Replaced the unsafe guards (if (editor?.view && …), !editor.view.dom, editor?.view?.dom?.getBoundingClientRect?.()) with that helper.
3. Added isMarkdownView to the select-all effect's dependencies so the keyboard handler correctly re-attaches when the user switches from markdown back to RTE mode (when the view remounts) — otherwise it would never re-run since editor stays the same instance.

### DIFF
--- a/app/assets/javascripts/tiptap/SimpleEditor.js
+++ b/app/assets/javascripts/tiptap/SimpleEditor.js
@@ -32,6 +32,25 @@ import { EditorProvider } from "./EditorContext";
 import { AnnouncerPlugin } from "./AnnouncerPlugin";
 import { shortcuts, translations } from "./localization";
 
+// In TipTap 3, `editor.view` returns a Proxy stub (which is truthy) until the
+// underlying ProseMirror view is actually mounted. Accessing most properties on
+// that stub — e.g. `dom`, `coordsAtPos`, `posAtDOM` — THROWS:
+//   "[tiptap error]: The editor view is not available. Cannot access
+//    view['dom']. The editor may not be mounted yet."
+// Optional chaining (`editor?.view?.dom`) does NOT protect against this because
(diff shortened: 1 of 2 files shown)


whabanks (approved):
Tested production using Firefox and consistently had trouble getting the editor load as the client described:

Image

Running locally with Firefox, it's consistently loading:

https://github.com/user-attachments/assets/12a878a3-6887-48ca-8e5d-7b304bcc4e93

PR #2718, opened 2026-07-06
Fix/add service stick css

amazingphilippe:
# Summary | Résumé

Use CSS sticky positioning instead of JavaScript on the account page. 

- Check the main screen where you can add a service
- When you have multiple services (you can fake it by duplicating some li nodes)
- The button sticks at the bottom, and un-sticks only when you reach the end of the list

https://github.com/user-attachments/assets/93b234ef-cff6-4d84-a5ec-ce0a4959ec1d

### DIFF
--- a/.devcontainer/devcontainer-lock.json
+++ b/.devcontainer/devcontainer-lock.json
@@ -1,0 +1,9 @@
+{
+  "features": {
+    "ghcr.io/devcontainers/features/node:1": {
+      "version": "1.7.1",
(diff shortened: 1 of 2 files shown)


smcmurtry (changes requested):
Awesome that you found a pure css way to do this! One small thing I noticed: there is no bottom padding on the button when it goes to sticky mode:

Image

smcmurtry:
Sorry, one more thing I didn't catch earlier. The horizontal position and width is off for narrower screens:
Screenshot 2026-07-06 at 4 20 46 PM

smcmurtry (approved):
Looks good!

PR #2725, opened 2026-07-08
Fix bug with negative "in transit" counts for jobs

smcmurtry:
# Summary | Résumé

This PR changes the in-transit calculation in admin to use the real number for that status, whereas before it was using a value arrived at by subtracting delivered and failed from the total rows:
```
notification_count − delivered − failed = 33169 − 33006 − 166 = −3
```

Which is wrong when the delivery total includes duplicates.

This is a long-standing bug that we have just received a support ticket for. The underlying problem is that duplicate notifications are sometimes sent for a given job. I have confirmed this by running the following query:

```
select count(id), n.job_row_number
from notifications n 
where n.job_id = '74677119-94f3-4d46-b937-a586c0647e50'
group by n.job_row_number  
order by count(id)
```

and observed that there are 60 rows (out of a 33,169 row job) that have been inserted twice.

Looking at the actual counts with this query:
```
select count(id), n.notification_status
from notifications n 
where n.job_id = '74677119-94f3-4d46-b937-a586c0647e50'
group by n.notification_status 
```
We get:

```
32 temporary-failure
57 sending
134 permanent-failure
33006 delivered
```

I connected my locally running admin to the prod api to verify that the fix works:

Before:
Screenshot 2026-07-08 at 3 50 51 PM

After:
Screenshot 2026-07-08 at 3 51 36 PM

1. check out the branch
2. modify your .env to connect to the prod api
3. visit the following link: http://localhost:6012/services/8cc00648-0b2e-474e-a5a5-9da3c72b2560/jobs/74677119-94f3-4d46-b937-a586c0647e50?status=sending
4. verify that you see the correct (non-negative) total for in transit notifications.

### DIFF
--- a/app/main/views/jobs.py
+++ b/app/main/views/jobs.py
@@ -404,7 +404,7 @@ def get_status_filters(service, message_type, statistics):
         }
     else:
         stats = statistics[message_type]
-    stats["sending"] = stats["requested"] - stats["delivered"] - stats["failed"]
+    stats["sending"] = max(0, stats["requested"] - stats["delivered"] - stats["failed"])
 
     filters = [
         # key, label, option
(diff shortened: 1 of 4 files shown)


app/main/views/jobs.py
```diff
@@ -431,11 +431,20 @@ def get_status_filters(service, message_type, statistics):
...
+    # retries), which made the derived value go negative.
+    if job["job_status"] == "scheduled":
+        sending = 0
+    elif "notifications_sending" in job:
+        sending = job["notifications_sending"]
+    else:
+        sending = max(
+            0,
+            job.get("notification_count", 0) - job.get("notifications_delivered", 0) - job.get("notifications_failed", 0),
+        )
```
smcmurtry, line 447: We fall back to the subtraction method if the job hasn't started sending yet and the total is not present

app/main/views/jobs.py
```diff
@@ -404,7 +404,7 @@ def get_status_filters(service, message_type, statistics):
         }
     else:
         stats = statistics[message_type]
-    stats["sending"] = stats["requested"] - stats["delivered"] - stats["failed"]
+    stats["sending"] = max(0, stats["requested"] - stats["delivered"] - stats["failed"])
```
jzbahrai, line 407: i love math

jzbahrai (approved):

PR #2735, opened 2026-07-17
feat: replace FF_SALESFORCE_CONTACT env var

patheard:
# Summary | Résumé
Replace the feature flag with an inferred `IS_GC_ORGANISATIONS` boolean based on the existence of the GC organisation bucket and filename.

This is being done as part of the Notify Salesforce disconnect.

# Related
- https://github.com/cds-snc/notification-api/pull/2955
- https://github.com/cds-snc/ADR/pull/92

### DIFF
--- a/app/__init__.py
+++ b/app/__init__.py
@@ -254,7 +254,10 @@ def create_app(application):
     application.jinja_env.globals["NotifyEnv"] = NotifyEnv
 
     # Initialize the GC Organisation list
-    if application.config["FF_SALESFORCE_CONTACT"]:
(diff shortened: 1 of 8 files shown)


andrewleith (approved):
LGTM - since the env vars `GC_ORGANISATIONS_BUCKET_NAME` and `GC_ORGANISATIONS_FILENAME` will not be changing here, the app will continue to load and show the org list from the s3 bucket and/or the fallback file. ✅ 

Thanks for doing this Pat!

PR #2736, opened 2026-07-20
Auto-remove malware-containing files

whabanks:
# Summary | Résumé
This PR updates the file attachment component to exclude file attachments which failed malware scanning. Upon initial upload the file will be displayed in the list of attachments with the existing error message about malware. Upon leaving the page for that template and returning to it the file that failed malware scanning will not be listed in the attached files section.

1. Upload a file that will fail the malware scan:
```
X5O!P%@AP[4\PZX54(P^)7CC)7}$EICAR-STANDARD-ANTIVIRUS-TEST-FILE!$H+H*
```
- [x] Note that the file appears in the attachment list with an error
3. Navigate away from the template --> Navigate back to the template
- [x] Note that the "malicious" file is no longer listed as an attachment for that template.

### DIFF
--- a/app/main/views/templates.py
+++ b/app/main/views/templates.py
@@ -111,6 +111,20 @@ def get_email_preview_template(template, template_id, service_id):
     return email_preview_template
 
 
+def _get_visible_template_attachments(attachments):
+    visible_attachments = []
+
+    for attachment in attachments:
+        status = attachment.get("status") or "uploaded"
+
+        if status in ("deleted", "virus_scan_failed"):
+            continue
+
+        visible_attachments.append(attachment)
+
+    return visible_attachments
+
+
 def set_preview_data(data, service_id, template_id=None):
     key = f"template-preview:{service_id}:{template_id}"
     redis_client.set(key=key, value=json.dumps(data), ex=int(timedelta(days=1).total_seconds()))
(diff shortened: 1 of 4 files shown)


jzbahrai (commented):
-

app/main/views/templates.py
```diff
@@ -209,7 +223,7 @@ def view_template(service_id, template_id):
         and template["template_type"] == "email"
         and current_service.has_permission("upload_document")
     ):
-        template_attachments = current_service.get_template_attachments(template_id)
+        template_attachments = _get_visible_template_attachments(current_service.get_template_attachments(template_id))
```
jzbahrai: should we override get_template_attachments with visible=False/ true and default it accordingly vs a new func?

whabanks: We could certainly do that. The reason I didn't is filtering semantics differ depending on the context and a `visible` boolean flag would create some cognitive load here. You'd have to understand that visible = `uploaded` / `pending_virus_scan` and invisible = `virus_scan_failed` / `deleted`.

1. Getting attachments during a send we only want `uploaded` files
2. When we're at the template level we want everything except `uploaded` and `pending_virus_scan`

What do you think about a `filter_attachements` method, where we either pass in the desired statuses, or statuses to exclude instead, so it's explicit in each path?

jzbahrai: I think that is reasonable! lets do that

whabanks: Implemented!

jzbahrai (approved):

PR #2737, opened 2026-07-20
Update verbiage for send files by email setting

whabanks:
# Summary | Résumé
This PR removes the `(API Only)`text from the `Send files by email` setting as we can now send via the UI and API

# Related
* https://docs.google.com/document/d/1m_EzXNJbqjJH-WLzxRKMtO4k5CAHwk51uCXi7xJQR44/edit?pli=1&tab=t.0#task=doULW0OCbF-OneiL

1. Navigate to service settings
- [ ] Note that the `Send files by email` setting no longer indicates (API only)

### DIFF
--- a/app/main/views/service_settings.py
+++ b/app/main/views/service_settings.py
@@ -436,10 +436,13 @@ def set_sensitive_service(service_id):
 )
 @user_has_permissions("manage_service")
 def service_switch_upload_document(service_id):
+    if not current_app.config.get("FF_FILE_ATTACHMENTS"):
+        abort(404)
+
     title = _("Send files by email")
     form = ServiceOnOffSettingForm(name=title, enabled=current_service.has_permission("upload_document"))
     help = _(
-        "This feature is only available when sending through the API.<br>Learn more in the <a href='{}'>API documentation</a>."
+        "Allow files to be attached to email notifications.<br>Learn more in the <a href='{}'>API documentation</a>."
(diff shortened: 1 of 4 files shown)


app/main/views/service_settings.py
```diff
@@ -436,10 +436,13 @@ def set_sensitive_service(service_id):
...
 @user_has_permissions("manage_service")
 def service_switch_upload_document(service_id):
+    if not current_app.config.get("FF_FILE_ATTACHMENTS"):
+        abort(404)
+
     title = _("Send files by email")
     form = ServiceOnOffSettingForm(name=title, enabled=current_service.has_permission("upload_document"))
     help = _(
-        "This feature is only available when sending through the API.<br>Learn more in the <a href='{}'>API documentation</a>."
+        "Allow files to be attached to email notifications.<br>Learn more in the <a href='{}'>API documentation</a>."
```
jzbahrai: do we need the second part to this anymore or it should say you can upload through the interface or by API (and then link)?

 and this should be behind a feature flag in the line? vs aborting on line 440 means throwing a 404 even if they should be able to send through the API

whabanks: Good catches here, the latter suggestion makes sense to me. I adjusted the content and moved the FF check that shouldn't have blocked usage of the endpoint in the first place. Let me know if the content makes sense.

jzbahrai (approved):

PR #2740, opened 2026-07-20
Add the `manage_reports` permission for new api keys

smcmurtry:
# Summary | Résumé

Add the `manage_reports` permission for new api keys. This is hidden behind the `FF_REPORT_API`.

1. check out the branch locally 
2. add `FF_REPORT_API=true` to your .env for admin and api
3. run admin and api locally
4. create a new api key, verify that you see the new checkbox for the "manage reports" permission
5. look in your local db in the `api_keys` table, verify that the new api key has "manage_reports` in the permissions column.

### DIFF
--- a/app/main/forms.py
+++ b/app/main/forms.py
@@ -1014,9 +1014,13 @@ class ChooseTimeForm(StripWhitespaceForm):
 
 
 class CreateKeyForm(StripWhitespaceForm):
-    def __init__(self, existing_keys, *args, **kwargs):
+    def __init__(self, existing_keys, ff_report_api=False, *args, **kwargs):
         self.existing_key_names = [key["name"].lower() for key in existing_keys if not key["expiry_date"]]
         super().__init__(*args, **kwargs)
+        if ff_report_api:
+            self.manage_templates.choices = list(self.manage_templates.choices) + [
+                ("manage_reports", _l("Allow this key to create and download delivery reports"))
+            ]
 
     key_type = RadioField(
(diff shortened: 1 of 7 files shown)


app/main/views/api_keys.py
```diff
@@ -115,6 +118,7 @@ def create_api_key(service_id):
         disabled_options=disabled_options,
         option_hints=option_hints,
         show_manage_templates=current_app.config["FF_ADD_TEMPLATE_PERM"] and current_user.has_permissions("manage_api_keys"),
```
jzbahrai, line 120: just an ask if you can edit this to remove the FF (you dont have to)

app/main/forms.py
jzbahrai, line 1036: should l1020 be added here? and lets change the variable from manage_templates to manage_perms?

app/main/views/api_keys.py
smcmurtry: sure! happy to do that but I will do it in a separate pr since it is in a few different places

app/main/forms.py
smcmurtry: Okay I have changed the variable name for the options from "manage_templates" to "permissions". The code to add the new option in is line 1020 to 1023: https://github.com/cds-snc/notification-admin/pull/2740/changes#diff-a1c8d24b22d4478fe71f75fd43b71b18dd82aae97bc63de84473a6da1902909bR1020-R1023

jzbahrai (approved):

PR #2743, opened 2026-07-22
Show list of attachments on template edit page

whabanks:
# Summary | Résumé
This PR adds a non-editable list of template file attachments to the edit template page. The design mirrors what is displayed on the template page but as simple static markup with the "cancel" button replaced by the file size

1. Navigate to a template with files attached --> click edit
- [ ] Note the new section just below the editor that displays file attachments
3. Turn on VoiceOver and navigate through the elements
- [ ] `<ul>` focused --> "Files attached to this template"
- [ ] `<li>` focused --> "Attachment, `<filename>`, `<filesize>`"

### DIFF
--- a/app/templates/views/edit-email-template.html
+++ b/app/templates/views/edit-email-template.html
@@ -10,6 +10,7 @@
 {% from "components/template-content.html" import template_content with context %}
 {% from "components/links.html" import content_link with context %}
 {% from "components/tiptap-editor.html" import tiptap_editor with context %}
+{% from "components/notice.html" import notice %}
 
 {% block service_page_title %}
   {{ heading }}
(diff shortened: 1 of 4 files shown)


app/templates/views/edit-email-template.html
```diff
@@ -85,6 +86,50 @@
         initialMode="rte" if current_user.default_editor_is_rte else "markdown") 
       }}
 
+      {% if config['FF_FILE_ATTACHMENTS'] and attachments %}
+        {% set attachment_count = attachments|length %}
+
+        <section class="mb-16">
```
andrewleith, line 90: I created a partial for this for the sending page, could you use that here? Its in `/app/templates/partials/template-attachments-list.html`.

andrewleith (changes requested):
I think you could re-use a partial for the static list, details below.

app/templates/views/edit-email-template.html
whabanks: Fixed!

andrewleith (approved):
LGTM!

PR #2747, opened 2026-07-27
feat: implement file attachment service allowlist functionality and remove `FF_FILE_ATTACHMENTS`

andrewleith:
# Summary | Résumé

This pull request refactors how file attachments are enabled for services, moving from a global feature flag (`FF_FILE_ATTACHMENTS`) to a per-service allowlist (`FILE_ATTACH_SERVICES`). The codebase now checks if a service is allowed to use file attachments using a new helper function, improving flexibility and security. The update also modifies tests and templates to use the new logic.

This change allows us to roll this feature out to production more slowly, test with clients, and ensure our service doesn't get marked as spammy by downstream providers as we roll this out more broadly.

- Add a service to the safelist (`FILE_ATTACH_SERVICES="your-service-id,`) via `config.py`
- [ ] File attachments features are enabled and working


whabanks (approved):
LGTM

PR #2757, opened 2026-08-12
fix(tests): add realPress to navigate to end of block conditional content

andrewleith:
# Summary | Résumé

This PR updates a test that recently started failing in CI.  I tried to figure out the cause thinking it was an update to one of RTE's dependencies or maybe cypress itself but I could not find the cause.

This change makes the test more explicit and should now pass.

- [ ] CI passes


P0NDER0SA (approved):
sure!

PR #2760, opened 2026-08-14
Add more help to the info pane in th RTE

amazingphilippe:
# Summary | Résumé

This PR adds different help content to the RTE component. We can now add different focused sections of help, triggered by a small navigation. This pane will be used to address the RTE's most problematic shortcommings with semi-detailed docs. 

- There are some tests to ensure the nav items match the sections available
- Each pane is a distinct component file, for readability

- [ ] On the RTE component
- [ ] Trigger Help button
- [ ] See new nav items
- [ ] Each nav item activates a different pane with content.
- [ ] Test in French too.


jzbahrai (approved):

app/assets/javascripts/tiptap/InfoPane/CustomContent.js
andrewleith, line 1: The `tiptap/` folder is getting a bit full here. May be worth adding these new files to a subfolder to keep organized.

app/assets/javascripts/tiptap/MenuBar.js
```diff
@@ -807,29 +848,26 @@ const MenuBar = ({
...
+                    aria-current={infoPaneTarget === id ? "page" : undefined}
+                    onClick={(event) => handleInfoPaneLinkClick(event, id)}
+                  >
+                    {tabLabel}
+                  </a>
+                </li>
+              ))}
             </ul>
-          </p>
+          </nav>
```
andrewleith, line 866: I find the interaction here a little cumbersome. It may be that when I toggle the help pane the first tab actually takes up nearly the whole screen, pushing the editor content itself almost out of view.  Not sure exactly but it left me feeling a little lost for a few moments.

andrewleith: Would an animation help?

andrewleith (commented):
Functionally this works as designed.  A few comments around file organization and UX for discussion.

app/assets/javascripts/tiptap/InfoPane/CustomContent.js
amazingphilippe: 💯 I almost did that. I think these sections are bound to change in the future as we improve RTE. I'll do the reorganization.

app/assets/javascripts/tiptap/MenuBar.js
amazingphilippe: Since the info pane is removed entirely when hidden, there is nothing to animate. There is support for @starting-styles across all browsers, but not for keyword. So animations going from 0 to auto will not work in firefox and safari. I think its still a great enhancement and we can add the height transition, then wait for FF and Safari to support it.

amazingphilippe: Animating between two intrinsic size values is also not possible. So switching tabs also can't be animated if the height is different.

PR #2761, opened 2026-08-18
New feat banner

amazingphilippe:
# Summary | Résumé

Adding a prototype for a new-feature banner that we can anchor to any part of the UI, or insert as a flowing container where needed.

How it works:

- Add the new feature banner anywhere. We don't want it to hide content (for a11y), so it has to position in the document flow. 
- If we plan this well, we can add an optional CSS anchor on a specific element. A caret will point to that anchor for some visual interest.
- Set a lifespan so that the banner stops showing up. 
- There is a script to find expired banners so that we can easily clean up periodically

- [x] Check any email template where file attachments is enabled. 
- [x] Refresh the page, the banner should persist
- [x] Hide the banner
- [x] Refresh the page, the banner should not show up


app/templates/views/templates/template.html
```diff
@@ -102,6 +103,13 @@
             url_for('main.template_attachment_status', service_id=current_service.id, template_id=template.id),
             url_for('main.download_template_attachment', service_id=current_service.id, template_id=template.id)
           ) }}
+          {{ new_feature(
+              title=_("Upload attachments without the API"),
+              cta=_("About file attachments"),
+              cta_link="/new-features",
```
smcmurtry, line 109: It would be better if we could link directly to the relevant section on this page with a `/new-features#ui-file-attachments` link. But it looks like that page doesn't have ids on each section. Also the formatting of the new-features page looks a bit off. Not really related to the new component, but it would be good to fix that before this goes live.

app/templates/views/templates/template.html
```diff
@@ -102,6 +103,13 @@
             url_for('main.template_attachment_status', service_id=current_service.id, template_id=template.id),
             url_for('main.download_template_attachment', service_id=current_service.id, template_id=template.id)
           ) }}
+          {{ new_feature(
```
smcmurtry, line 106: Could you move this above the "Attached files" component? It seems more natural to me that you would read the "new feature" banner before you see the new feature.

scripts/check_expired_banners.py
smcmurtry, line 1: I don't think we need this new script - if we can remember to run the script we can remember to do a "find in project" for `new_feature(` banners and check the lifespans.

smcmurtry (changes requested):
The new component looks good! Could we leave leave it off the template page until we get section links working on the "new features" page?

PR #2764, opened 2026-08-19
Fix RTE ally errors

amazingphilippe:
# Summary | Résumé

Fixes a11y errors introduced by the RTE info pane:

- svg icons were missing role=img
- some tabs were referencing an id that did not exist. Replaced with an aria-label. 

- [ ] On rte component
- [ ] Open info pane
- [ ] Run a11y scanner - no warnings for the RTE itself


andrewleith (approved):
Looks like you need to run `make format` - but the cypress tests are now passing!

PR #2774, opened 2026-08-28
chore: remove broken roadmap redirect from navigation

andrewleith:
# Summary | Résumé
This pull request removes the internal redirect and navigation references for the `/features/roadmap` route, effectively deprecating or removing the roadmap feature from the application's navigation and routing.

Navigation and routing updates:

* Removed the `/features/roadmap` internal redirect route from `index.py`, so users can no longer access it via that endpoint.
* Removed `"redirect_roadmap"` from both `HeaderNavigation` and `OrgNavigation` in `navigation.py`, so the roadmap is no longer present in the application's navigation menus. [[1]](diffhunk://#diff-39387df3a9f89b313976957e7b5457be20deab1017b2d895541b142b957f1972L460) [[2]](diffhunk://#diff-39387df3a9f89b313976957e7b5457be20deab1017b2d895541b142b957f1972L824)

- [ ] Going to `/features/roadmap` does not result in a 500 error


jimleroyer (approved):
LGTM

PR #2779, opened 2026-09-10
Block domains 

jzbahrai:
# Summary | Résumé

Policy requires us to block parl.gc.ca from using notify.This PR is made to facilitate that change


smcmurtry (approved):
Tested locally - this will block new accounts from being created but it doesn't block existing users with that email domain from signing in. But the blocking new accounts part is working ✅

PR #2781, opened 2026-09-14
validator for newsletter

ben851:
# Summary | Résumé

We had a fuzzing attack alarm on the newsletter route using this URL
```
/newsletter/%3Ffoo%3Dbar/change-language
```

It's obviously malformed but the admin still forwarded the request to the API. This shouldn't have happened. This introduces a validator at the newsletter route to verify that the URL is at least not malformed before sending to the API.

Merge to staging, try the above URL - hopefully you get a 404 instead of a 500


jimleroyer (approved):
Great

PR #2783, opened 2026-09-15
Migrate image consumers away from latest

P0NDER0SA:
## Summary

Migrate image consumers away from mutable \ tags while keeping deployment and scan tag selection aligned.

## Details

- Removes the build cache dependency on \\`notify-admin:latest\\`
- Generates the build SBOM from the build SHA
- Makes the scheduled vulnerability scan consume \\`ADMIN_DOCKER_TAG\\` from manifests staging config

## Validation

- Scheduled scans read the exact image tag from \
- Scan workflows fail when the required manifest tag is missing or mutable
- Build-time SBOMs continue to use the same SHA tag dispatched to manifests
- Ran \
- Workflow linting was not run locally because \ is unavailable


whabanks (approved):
LGTM

